PR #1, opened 2019-02-27
SREP-923: Copy over SRE alerting rules

lisa:
* stuck-volumes.yaml SREP-520
* stuck-builds.yaml SREP-683
* router-health.yaml SREP-695
* ebs-iops-burstbalance.yaml SREP-686
* dns-latency.yaml N/A
* check-full-cust-pv.yaml SREP-687


lisa:
@jewzaam quick thought on the directory structure? I know we can change it later, but we should probably get the pattern set early.

jewzaam:
Why not follow the pattern used in operators and other projects and put resources under a "deploy" directory?  I can see having separation, not sure if by kind or by purpose makes more sense.  To the `managed-operator` that will be deploying this stuff it doesn't matter.  I want to have a single dir we point it at that will be used for applying to the cluster and it can be recursive.  Order within this repo shouldn't matter.

Kind of a design-on-the-fly thought though:  why have this static stuff separate from `managed-operator`?  Could we just have it be part of that?  I guess we can have a different structure here that is more human friendly.

deploy/
--- crds/
--- deployments/
--- roles/
etc...

lisa:
@jewzaam One reason I didn't put them in a `deploy` directory is because these raw YAML files aren't suitable for `oc apply -f` since they're not complete `PrometheusRule` manifests. I had thought the deployment tool (whatever that looks like) would wrap these rules in the appropriate CR.

Perhaps we need a way to differentiate "curated" CRs and "raw" content? The former could go into the `deploy` directory and maybe the latter some other.

jewzaam:
Why not make the files all complete CRs?  This should just be the definition of any computed metrics and new alerts.  Having some code that has to run to build the complete YAML adds complexity.  Do we need that complexity?  Do we want it?  What is the value vs cost?

lisa:
@jewzaam No reason they can't be. I just thought the management operator would prefer the rules to be in "raw" format (eg, not `PromethesRule` CRs), but it's just as easy for the operator to pluck the rules out of the CR. I'll update the PR and shuffle things around some.

lisa:
@jewzaam 4f6e248 has some updates. I'm easy to flatten those directories more.

deploy/crs/prometheus/prometheus_rules/sre/check-full-cust-pv.yaml
```diff
@@ -0,0 +1,35 @@
...
+  labels:
+    prometheus: sre-filesysfull
+    role: alert-rules
+  name: sre-filesysfull
+spec:
+  groups:
+  - name: sre-filesysfull
+    rules:
+      # This matches with the default KubePersistentVolumeUsageCritical; this rule inverts the regex match.
+    - alert: KubeCustomerPersistentVolumeUsageCritical
```
jewzaam: The group has a prefix of `sre-` but wonder if we need to also prefix alerts with something like `Sre`.  This goes back to the question of how we'll ensure PE rules and alerts don't conflict / overlap with SRE rules and alerts.

I have deployed this file twice, one with the `sre` prefix and one with `sre2`.  I haven't been able to review alerts from a UI point of view (need fix for auth bug to land) but in-cluster it appears to be happy...

lisa: You "should" (scare quotes) be able to use the Alertmanager API to query the status of those things? I'm not sure, and that might be harder than the other auth problem.

On the rule name, I do wonder what the chances of two groups creating a rule with the same name are. The more I think about it the more I like the idea of heading it off at the pass and having our team name in there somewhere.

After lunch today I'll put up a change.

lisa:
@jewzaam no, it makes sense. I kept it separate to distinguish other SRE teams. But if it's just us then we can drop the distinction.

PR #3, opened 2019-03-14
SREP-693: Add alert rule for unschedulable nodes

lisa:
Warn after an hour.

### DIFF
--- a/deploy/prometheus/100-ebs-iops-burstbalance.PrometheusRule.yaml
+++ b/deploy/prometheus/100-ebs-iops-burstbalance.PrometheusRule.yaml
@@ -5,6 +5,7 @@ metadata:
     prometheus: sre-ebs-iops-burstbalance
     role: alert-rules
   name: sre-ebs-iops-burstbalance
+  namespace: openshift-monitoring
 spec:
   groups:
   - name: sre-ebs-iops-burstbalance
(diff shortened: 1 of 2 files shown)


deploy/prometheus/100-node-unschedulable.PrometheusRule.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-node-unschedulable
+    role: alert-rules
+  name: sre-node-unschedulable
```
rogbas: should we explicitly lock the namespace?

lisa: @rogbas I didn't add it because none of the other rules in this repository specify. I think that's a defect though and that we ought to place these in `openshift-monitoring`. I'll add it for this rule and then separately add for the rest.

deploy/prometheus/100-ebs-iops-burstbalance.PrometheusRule.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
     prometheus: sre-ebs-iops-burstbalance
     role: alert-rules
   name: sre-ebs-iops-burstbalance
-  namespace: openshift-monitoring
```
lisa: This was made in the wrong file. While I will be making this change, I want to make it in a separate changeset and not for #3.

rogbas: ok

PR #5, opened 2019-03-19
Added osd-sre-admin ClusterRole

jewzaam:
SRE shouldn't be a cluster-admin for day-to-day cluster administration.  To do this, a new clusterrole `osd-sre-admin` is created in this PR to revoke some specific permissions.  Unfortunately we cannot deny access in a clusterrole, it's allow only.  So the script to generate the clusterrole was created to walk the API.  See the readme in `generate/sre-authorization` for more info on that (what, why, when, who).

There originally was a group created in this PR but when it's recreated all users are deleted.  We'll simply manage it somewhere else.  The binding will be applied once the group is created and doesn't require the group.


generate/sre-authorization/blacklist.json
```diff
@@ -0,0 +1,32 @@
...
+        "verbs": ["create","delete","patch","update","deletecollection"]
+    },
+    {
+        "apiGroup": "config.openshift.io",
+        "resources": ["oauths"],
+        "verbs": ["create","delete","patch","update","deletecollection"]
+    },
+    {
+        "apiGroup": "monitoring.coreos.com",
+        "resources": ["prometheusrules"],
```
rogbas: Shouldn't we add to this black list:
* `prometheus.monitoring.coreos.com`
* `alertmanagers.monitoring.coreos.com`
* `clusteroperators.operatorstatus.openshift.io`

generate/sre-authorization/osd-sre-admin.sh
```diff
@@ -0,0 +1,156 @@
...
+            echo "  - $RESOURCE"
+            echo "  verbs:"
+            echo "  - '*'"
+            continue
+        fi
+
+        VERBS=""
+        VERB_COUNT=0
+
+        for VERB in `oc get --raw /api/${VERSION} | jq -r ".resources[] | select(.name == \"$RESOURCE\") | .verbs[]" | sort`;
```
rogbas: if I do a `oc get --raw /api/v1` or `oc get --raw /apis`, custom objects like prometheusrules never appear.

Should we consider another way to tackle those objects? Like `oc get crds`

rogbas: nevermind, I just saw the step below

generate/sre-authorization/blacklist.json
```diff
@@ -0,0 +1,32 @@
...
+        "verbs": ["create","delete","patch","update","deletecollection"]
+    },
+    {
+        "apiGroup": "machineconfiguration.openshift.io",
+        "resources": ["machineconfigs","machineconfigpools","controllerconfigs"],
+        "verbs": ["create","delete","patch","update","deletecollection"]
+    },
+    {
+        "apiGroup": "machine.openshift.io",
+        "resources": ["machines","machinesets"],
```
rogbas: should we add `machineclass` as well?

generate/sre-authorization/blacklist.json
```diff
@@ -0,0 +1,32 @@
...
+        "verbs": ["create","delete","patch","update","deletecollection"]
+    },
+    {
+        "apiGroup": "config.openshift.io",
+        "resources": ["oauths"],
```
rogbas: Is it missing `authentication.config.openshift.io` and `authentication.operator.openshift.io`?

deploy/sre-authorization/03-osd-sre-admin.ClusterRoleBinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: osd-sre-admin
```
rogbas: if we are using this setup for Online and Pro as well, would it make sense to replace OSD with managed?

generate/sre-authorization/blacklist.json
jewzaam: We could, didn't see any resources in there so didn't put it in.

generate/sre-authorization/blacklist.json
jewzaam: Was focused on what we are driving into the cluster via hive.

generate/sre-authorization/blacklist.json
jewzaam: We don't manage prometheus or clusteroperators CR via hive so didn't include.  Alertmanagers makes sense to blacklist.  Don't see a problem blacklisting prometheuses too.  Will leave clusteroperators as I don't know if that means we can't upgrade now.

generate/sre-authorization/osd-sre-admin.sh
```diff
@@ -0,0 +1,156 @@
...
+
+echo "apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-sre-admin
+rules:"
+
+# /api
+GROUP=""
+for VERSION in `oc get --raw /api | jq -r .versions[] | sort`;
```
lisa: We are going to be running this rebuild script periodically and I think we want to avoid the dynamic rebuild and, instead, provide a static input file to the script. The inputs can be the safelist and blocklist of resources the group are allowed to access, and not allowed to access, respectively.

What leads me to the conclusion is the pre-run setup: log into a cluster (where are the creds? how to keep them safe?), connect to it (what if there's a network partition?) and fetch the resources. This is a kind of toil and could lead to future issues. In addition to my setup concern, I have a readability concern as well. 

We don't know what the output of the script will be by only reading it because the inputs are essentially dynamic (via the `oc` command). If we provide the resources in plain text then we can better predict what happens when it runs.

jewzaam: But that's part of the problem, the inputs are dynamic.  If we have 10 new CR's in an OSD cluster that we need to manage, we need to add them to the osd-sre-admin clusterrole.  

I do not see this script being run often.  The clusterrole could be managed manually if it's better.  But I think we should keep it around until a bit past GA so we can easily generate something new.

On knowing what will change, that's true.  But we have a baseline to look at for future use of the script, simply `make; git diff` and see what changed.  The output of the script is static.  CI should not run the script.  So CI will only ever see a static file.  Humans need to review changes to ensure the delta is good.

jewzaam:
Talked to @lisa just now and we've landed on a different approach being better.  Instead of blacklisting permissions we'll grant cluster-reader to osd-sre-admin and only a select subset of delete, create, update, patch to specific resources.  I'll update the PR as I have time today.

PR #7, opened 2019-03-21
Moved read verbs to osd-sre-admin clusterrole

jewzaam:
The `cluster-reader` role doesn't give everything, found I couldn't view operator stuff and cluster settings.  So moving to allow get, list, and watch explicitly in `aos-sre-admin` clusterrole.


deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
```diff
@@ -64,14 +72,6 @@ rules:
...
   verbs:
   - '*'
-- apiGroups:
-  - config.openshift.io
-  resources:
-  - clusterversions
-  verbs:
-  - get
-  - list
-  - update
```
lisa: Why remove `update` ?

jewzaam: Not intentional at the time, but I have an answer now.  Operator version management should come from `hive` or `managed-operator`.  If we need to manipulate CSVs it will be `create` and maybe `delete`, not `update.

PR #10, opened 2019-03-26
Add login branding templates

liamwazherealso:
These templates should be available as artifacts so that clusters can import these templates as secrets.


jewzaam:
done in https://github.com/openshift/managed-cluster-config/pull/24

PR #13, opened 2019-05-03
Added selectorsyncset generation

rafael-azevedo:
This PR added a script and template to generate selectorsyncset. In addition it added a Makefile and project.mk to run the script and pass in the required parameters.

### DIFF
--- a/scripts/generate_syncset.py
+++ b/scripts/generate_syncset.py
@@ -44,8 +44,8 @@ if __name__ == '__main__':
     # Get the template data
     template_data = get_yaml(arguments.template_path)
     # Configure template
-    template_data['metadata']['labels']['url'] = arguments.git_url
-    template_data['metadata']['labels']['gitHash'] = arguments.git_hash
+    template_data['metadata']['labels']['managed.openshift.io/git-source'] = arguments.git_url
+    template_data['metadata']['labels']['managed.openshift.io/git-hash'] = arguments.git_hash
     template_data['resources'] = yamls
     # write selectorsyncset file
     with open(arguments.destination,'w') as outfile:
--- a/scripts/generate_syncset.py
+++ b/scripts/generate_syncset.py
@@ -1,0 +1,52 @@
+#!/usr/bin/python
+import os
+import sys
+import yaml 
+import argparse
+
+def get_yaml_all(filename):
+    with open(filename,'r') as input_file:
+        return list(yaml.load_all(input_file))
+
+def get_yaml(filename):
+    with open(filename,'r') as input_file:
+        return yaml.safe_load(input_file)
+
+def get_all_yaml_files(path):
+    file_paths = []
+    for r,d,f in os.walk(path):
+        for file in f:
+            if file.endswith('.yml') or file.endswith('.yaml'):
+                file_paths.append(os.path.join(r,file))
+    return file_paths
+
+def get_all_yaml_obj(file_paths):
+    yaml_objs = []
+    for file in file_paths:
+        objects = get_yaml_all(file)
+        for obj in objects:
+            yaml_objs.append(obj)
+    return yaml_objs
+
+
+if __name__ == '__main__':
+    #Argument parser
+    parser = argparse.ArgumentParser(description="selectorsyncset generation tool", usage='%(prog)s [options]')
+    parser.add_argument("--template-path", "-t", required=True, help="Path to template file [required]")
+    parser.add_argument("--yaml-directory", "-y", required=True, help="Path to folder containing yaml files [required]")
+    parser.add_argument("--destination", "-d", required=True, help="Destination for selectorsynceset file [required]")
+    parser.add_argument("--git-url" , "-u", required=True, help="URL for the git repo [required]")
+    parser.add_argument("--git-hash", "-c", required=True, help="Commit hash of commit used [required]")
+    arguments = parser.parse_args()
+
+    # Get all yaml files as array of yaml objecys
+    yamls = get_all_yaml_obj(get_all_yaml_files(arguments.yaml_directory))
+    # Get the template data
+    template_data = get_yaml(arguments.template_path)
+    # Configure template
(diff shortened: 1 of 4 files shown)


rafael-azevedo:
We will need to create a directory as required for hive to pull the yaml file and then change the variable SELECTOR_SYNC_SET_DESTINATION in projects.mk to reflect this directory change.

scripts/generate_syncset.py
```diff
@@ -0,0 +1,55 @@
...
+    for file in file_paths:
+        objects = get_yaml_all(file)
+        for obj in objects:
+            yaml_objs.append(obj)
+    return yaml_objs
+
+
+if __name__ == '__main__':
+    #Argument parser
+    parser = argparse.ArgumentParser(description="selectorsyncset generation tool")
```
jharrington22: Adding `, usage='%(prog)s [options]'` to `argparse.ArgumentParser()` would print the name of the file in the help usage like `usage: generate_syncset.py [options]`

scripts/templates/selectorsyncset.yaml
```diff
@@ -0,0 +1,16 @@
+
+apiVersion: hive.openshift.io/v1alpha1
+kind: SelectorSyncSet
+metadata:
+  generation: 1
+  name: osd-manged-cluster-config
+  labels:
+    osd-config: True
```
jharrington22: @jewzaam @rafael-azevedo should we make this osdConfig, looks kind of odd but matches convention.

scripts/generate_syncset.py
```diff
@@ -0,0 +1,55 @@
...
+
+def get_yaml(filename):
+    with open(filename,'r') as input_file:
+        return yaml.safe_load(input_file)
+
+def get_all_yaml_files(path):
+    file_paths = []
+    for r,_,f in os.walk(path):
+        for file in f:
+            if '.yml' in file:
```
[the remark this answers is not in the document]
jewzaam: And we want files that end with those extensions, not have them in the middle.

scripts/templates/selectorsyncset.yaml
jewzaam:
```suggestion
    managed.openshift.io/osd: "true"
```

jewzaam: Also were we not going to put the URL to the git repo?  Hash is going to be a little bit of a pain to put in, but would be nice.  Not required.

```
    managed.openshift.io/git-source: "https://github.com/openshift/managed-cluster-config/"
```

rafael-azevedo: is `managed.openshift.io/osd: "true"` to generic ? Maybe `managed.openshift.io/osd-config: "true"`

rafael-azevedo: This is resolved by the method in the comments below.

rafael-azevedo: Created the labels pr our conversation

jharrington22:
@jewzaam I still feel the following is better as it conforms to current Kubernetes key names.

```
  labels:
    managed.openshift.io/osd: True 
    managed.openshift.io/gitSource: 
    managed.openshift.io/gitHash:
```

Thoughts?

rafael-azevedo:
> @jewzaam I still feel the following is better as it conforms to current Kubernetes key names.
> 
> ```
>   labels:
>     managed.openshift.io/osd: True 
>     managed.openshift.io/gitSource: 
>     managed.openshift.io/gitHash:
> ```
> 
> Thoughts?

I agree, but I think either will work.

jewzaam:
> > @jewzaam I still feel the following is better as it conforms to current Kubernetes key names.
> > ```
> >   labels:
> >     managed.openshift.io/osd: True 
> >     managed.openshift.io/gitSource: 
> >     managed.openshift.io/gitHash:
> > ```
> > 
> > 
> > Thoughts?
> 
> I agree, but I think either will work.

That'll be fine for the key.  Taking a look at what is allowed for label values we'll not be able to put the git repo url into gitSource.  See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set: `Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.`

jharrington22:
@jewzaam Interesting I didn't know there was a character restriction or limit.

rafael-azevedo:
> > > @jewzaam I still feel the following is better as it conforms to current Kubernetes key names.
> > > ```
> > >   labels:
> > >     managed.openshift.io/osd: True 
> > >     managed.openshift.io/gitSource: 
> > >     managed.openshift.io/gitHash:
> > > ```
> > > 
> > > 
> > > Thoughts?
> > 
> > 
> > I agree, but I think either will work.
> 
> That'll be fine for the key. Taking a look at what is allowed for label values we'll not be able to put the git repo url into gitSource. See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set: `Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.`

We can shorten the url in the python code.

scripts/templates/selectorsyncset.yaml
```diff
@@ -4,13 +4,8 @@ kind: SelectorSyncSet
 metadata:
   generation: 1
   name: osd-manged-cluster-config
-  labels:
-    managed.openshift.io/osd: True
```
jewzaam: Hive is using all strings. Not sure about reason. We are not (at this time) planning to do anything with these other than help in operations and reporting spits, probably fine.

jewzaam:
I see the URL is removed. Decided shortening won't work or we don't really need the URL?

rafael-azevedo:
I decided that both shortening didn't work and we probably didn't need it.

PR #22, opened 2019-05-24
Added image_tag to template

rafael-azevedo:
This pr addes in required parameters for app-sre pipeline


hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -8,7 +8,7 @@ objects:
   metadata:
     generation: 1
     labels:
-      managed.openshift.io/gitHash: cbecd76
+      managed.openshift.io/gitHash: 4dee044
```
rogbas: Considering it's replaced by the script, shouldn't we make it obvious here? Like having a placeholder

PR #24, opened 2019-05-30
Add branding for console login.

jewzaam:
When applying changes in k8s an annotation is created w/ the previous state.  This has a limit of about 250k chars.  The login branding lands as secrets and the files are pretty big.  In order to have this be managed successfully we have to break those templates into separate resources, or oc apply will fail.  This means hive SelectorSyncSet also fails with a very large resource.

Therefore I have refactored a bit so that each sub-dir of deploy/ becomes a separate SelectorSyncSet in the Template.

There is one SelectorSyncSet added in the template that is not represented in deploy/.  It is to replace the older single SSS, `osd-managed-cluster-config`.  This means when we deploy this template no manual changes are required to disable the old one.


jewzaam:
@lisa @rafael-azevedo changes are squashed and pushed, updated desc of the PR to reflect modifications we made.

jewzaam:
@lisa @rafael-azevedo late addition, I updated to use IMAGE_TAG parameter, which is the git hash.  The generated template file will no longer change every commit unless there's an actual change, the hash will be injected by saasherder.

scripts/generate_syncset.py
```diff
@@ -29,31 +30,62 @@ def get_all_yaml_obj(file_paths):
...
-    parser.add_argument("--git-hash", "-c", required=True, help="Commit hash of commit used [required]")
+    parser.add_argument("--repo-name", "-r", required=True, help="Name of the repository [required]")
     arguments = parser.parse_args()
 
-    # Get all yaml files as array of yaml objecys
-    yamls = get_all_yaml_obj(get_all_yaml_files(arguments.yaml_directory))
     # Get the template data
     template_data = get_yaml(arguments.template_path)
-    # Configure template
+    # Configure template.  Use the last object as a template for adding SSS for each child dir of yaml_directory.
```
lisa: > `Use the last object as a template for adding SSS for each child dir of yaml_directory.`

Why? Can you expand on why this decision is made / why it's important?

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -41,7 +41,7 @@ objects:
   metadata:
     generation: 1
     labels:
-      managed.openshift.io/gitHash: 4fd61b1
+      managed.openshift.io/gitHash: ${IMAGE_TAG}
```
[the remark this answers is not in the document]
jewzaam: Fixed, `os.walk` will walk through each subdir.  Tried to find a better method or way to loop through, but this seems pretty straight forward.

scripts/generate_syncset.py
jewzaam: Instead of having a separate template for what the SSS looks like I am simply taking the last element of the objects array.  Probably better to have a separate file for this. I can make that change..

jewzaam: Updated to have separate template files for the `Template` and `SelectorSyncSet`.

scripts/generate_syncset.py
```diff
@@ -29,31 +32,60 @@ def get_all_yaml_obj(file_paths):
             yaml_objs.append(obj)
     return yaml_objs
 
+def process_yamls(name, directory, obj):
+    o = copy.deepcopy(obj)
+    # Get all yaml files as array of yaml objecys
```
lisa:
```suggestion
    # Get all yaml files as array of yaml objects
```

typo

scripts/generate_syncset.py
```diff
@@ -29,31 +32,60 @@ def get_all_yaml_obj(file_paths):
...
+    yamls = get_all_yaml_obj(get_all_yaml_files(directory))
+
+    for y in yamls:
+        if 'patch' in y:
+            if not 'patches' in o['spec']:
+                o['spec']['patches'] = []
+            o['spec']['patches'].append(y)
+        else:
```
lisa: Would we ever have a case when we have both patches and resources?

scripts/generate_syncset.py
```diff
@@ -29,31 +32,60 @@ def get_all_yaml_obj(file_paths):
...
-    # write selectorsyncset file
+    template_data = get_yaml(os.path.join(arguments.template_dir, "template.yaml"))
+    selectorsyncset_data = get_yaml(os.path.join(arguments.template_dir, "selectorsyncset.yaml"))
+
+    # Create required labels on SelectorSyncSet
+    if not 'labels' in selectorsyncset_data['metadata']:
+        selectorsyncset_data['metadata']['labels'] = {}
+
+    selectorsyncset_data['metadata']['labels']['managed.openshift.io/osd'] = "true"
+    selectorsyncset_data['metadata']['labels']['managed.openshift.io/gitRepoName'] = arguments.repo_name
```
lisa: Can/should(?) this and the previous line be moved into the template?

jewzaam: Done.

scripts/generate_syncset.py
jewzaam: In the case of oauth templates property, if the objects were not over the 250k-ish limit for oc apply I would have put the secret(s) and patch in a single SSS.

scripts/generate_syncset.py
rafael-azevedo: These could be moved into the template. The only con would be we would have to check if they exist . Also they only place they get changed is in this code so I am not sure what benefits we would get by moving this there. I had this thought when i wrote this and couldn't decide between which was more proper.

PR #27, opened 2019-06-04
Fix patch issue

lisa:
There were too many braces previously in the JSON, this removes the extras.


dak1n1:
For reference, this PR fixes SREP-1529.

PR #28, opened 2019-06-06
Fix PV quota to 100Gi

cblecker:
https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#local-ephemeral-storage

> Limits and requests for `ephemeral-storage` are measured in bytes.

Changes 100B to 100GiB


dak1n1:
I confirmed in 3.11 prod that 100Gi is what we're setting today.

cblecker:
Thanks @dak1n1!

@lisa: I'll leave this for you to review/merge :)

PR #29, opened 2019-06-06
Use scoped labels for resource quota exemption

cblecker:
This changes the labels used for resource quota exemption to use the scoped "managed.openshift.io" namespace. This way, it's easy to tell that these settings are specific to the OSD offering.

I've also swapped the negative label value for a positive one. As we're using `DoesNotExist` as the match operator, this makes the two possible states 'true' if the namespace is exempt, and the label should be omitted completely for normal namespaces.


cblecker:
/assign @lisa 

This is a totally optional improvement on this operator, @lisa. I'm not sure if there are specific reasons for the way it was, but I was digging through this code when I ran into an issue on my own cluster and these improvements made sense to me.

PR #30, opened 2019-06-06
Prune what sre-admins can do in OSD clusters.

jewzaam:
The goal of this is to drive all changes to config management.  It is expected that some updates may be required over time.

### DIFF
--- a/deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
+++ b/deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
@@ -62,6 +62,13 @@ rules:
   - projects
   verbs:
   - '*'
+# SRE can manage namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - '*'
 # SRE can review pod security policies
 - apiGroups:
   - security.openshift.io
--- a/deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
+++ b/deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
@@ -47,7 +47,8 @@ rules:
   verbs:
   - '*'
 # SRE can review pod security policies
-- apiGroup: security.openshift.io
+- apiGroups:
+  - security.openshift.io
   resources:
   - podsecuritypolicyreviews
   - podsecuritypolicyselfsubjectreviews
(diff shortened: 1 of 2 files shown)


aweiteka:
This looks reasonable but it's difficult to know what capabilities we might need given our lack of experience supporting v4. There are so many objects. What are some things an SRE cannot do? I'm not looking for a comprehensive list but it would be good to think this through and discuss here.

robotmaxtron:
> What are some things an SRE cannot do? I'm not looking for a comprehensive list but it would be good to think this through and discuss here.

- Create cronjobs (This seems strange to me)
- Create CRDs
- deletecollection nodes (Will this interfere with upgrades?)
- Create/Manage rolebindings & roles
- Create pods? Not sure if I'm reading that right.

@jewzaam Does this seem correct? Both in identifying some of the new restrictions and intended restrictions?

jewzaam:
Right.  SRE will be able to do very little.  I verified these changes allow us to cordon a node and drain it.  Haven't gone so far as to kill a node off, per existing SOPs that is done by deleting the instance in AWS not the node in the cluster.  Are there things that we should be able to do that we can't given the rbac in this PR?

jewzaam:
@robotmaxtron @aweiteka any chance of a lgtm or do we need some more changes?

deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
```diff
@@ -3,12 +3,15 @@ kind: ClusterRole
...
   name: osd-sre-admin
 rules:
+# SRE can get, list, and watch all resources
 - apiGroups:
   - ""
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
```
lisa: This (and the following 2 lines) appears to duplicate lines 20-22?

jewzaam: Value for `apiGroup` is "" on line #8 and "*" on line 16.  They are not duplicates, though they could be merged.

lisa: What I mean is does `*` cover `""`?

jewzaam: Turns out `"*"` does cover `""`.  Removed `""`.

jeremyeder:
I think SRE might be missing the capability to run ```oc debug node```.

```
$ oc debug node/ip-10-0-134-242.ap-southeast-2.compute.internal
Error from server (Forbidden): pods is forbidden: User "jeder@redhat.com" cannot create resource "pods" in API group "" in the namespace "default"
```

https://jira.coreos.com/browse/SREP-1699

PR #31, opened 2019-06-06
Regenerate and fix resource-quota and template

jewzaam:
The last PRs didn't include generation of everything, this means they didn't actually land in any envs.


Makefile
```diff
@@ -30,7 +30,7 @@ generate-oauth-templates:
 	# Each SSS must not be too big as well.  each sub-dir of deploy/ becomes a SSS.  therefore each of the html
 	# becomes a separate dir.  This is a k8s limitation for annotation value size.
 	for TYPE in login providers errors; do \
-		oc create secret generic osd-oauth-templates-$$TYPE -n openshift-config --from-file=$$TYPE.html=source/html/$$TYPE.html --dry-run -o yaml > deploy/osd-oauth-templates-$$TYPE/osd-oauth-templates-$$TYPE.secret.yaml; \
+		oc --config=.kubeconfig create secret generic osd-oauth-templates-$$TYPE -n openshift-config --from-file=$$TYPE.html=source/html/$$TYPE.html --dry-run -o yaml > deploy/osd-oauth-templates-$$TYPE/osd-oauth-templates-$$TYPE.secret.yaml; \
```
lisa: What happens when `--config=.kubeconfig` is absent? I'm a little surprised that this make target would fail.

jewzaam: It will fail saying it can't find a kubeconfig.  :angry: 
So, crafted this one so it's valid format with invalid data and it's happy to do a dry run. :man_shrugging:

PR #35, opened 2019-06-12
Fix kubeletconfig error

dak1n1:
A related PR (https://github.com/openshift/managed-cluster-config/pull/33) introduced these errors into stage:
`message: 'Error: could not find any MachineConfigPool set for KubeletConfig custom-kubelet'`

This PR fixes the above error, and updates the kubeletconfig to use a map instead of a list, since both the API spec and the default `kubeletconfig` show the `systemReserved` setting as being a map.

https://github.com/kubernetes/kubernetes/blob/9016740a6ffe91bb29824f80c34087b993903bd6/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L618-L628

SREP-1339


deploy/kubelet-config/03-kubelet-config.yaml
```diff
@@ -3,11 +3,9 @@ kind: KubeletConfig
 metadata:
   name: custom-kubelet
 spec:
-  machineConfigSelector:
+  machineConfigPoolSelector:
     matchLabels:
       managed.openshift.io/custom-kubelet: true
   kubeletConfig:
-    EnforceNodeAllocatable:
       system-reserved:
```
cblecker: This is what you're trying to modify, yes?:
https://github.com/kubernetes/kubernetes/blob/9016740a6ffe91bb29824f80c34087b993903bd6/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L618-L628

dak1n1: I think so, yes.

dak1n1: ok, updated to use a map instead.

dak1n1:
This change should be ready to go. There was a documentation error initially, but it's been fixed here: https://github.com/openshift/openshift-docs/pull/15435/files

cblecker:
@dak1n1 Did you re-run make after updating your change?

dak1n1:
@cblecker I ran make at some point and it seems to be up-to-date. I checked the generated template file in `git log` and saw this in the template:

```
+    - apiVersion: machineconfiguration.openshift.io/v1
+      kind: KubeletConfig
+      metadata:
+        name: custom-kubelet
+      spec:
+        kubeletConfig:
+          systemReserved:
+            cpu: 1000m
+            memory: 1Gi
```

jeremyeder:
@dak1n1 can you rebase this so we can get the approvals again?	@mwoodson or @aweiteka can you please take a look at this and add lgtm?

cblecker:
@dak1n1 is on PTO. As I noticed this issue is still happening, I helped to rebase her PR. Should be good to review.

/assign @jewzaam

jeremyeder:
Thanks @cblecker but @jewzaam is also on PTO through July 8th.

PR #36, opened 2019-06-17
[WIP] Removing sort of yaml and relying on git status for PR check

jewzaam:


jewzaam:
Order changes in the generated template.  Going to just close this and try to remember the PR if asked why we're sorting the yaml anymore.

PR #37, opened 2019-06-24
Adding SRE's identity provider configuration

jewzaam:
Adds SelectorSyncIdentityProvider that takes vars for many of the configuration items.  No defaults are provided to force configuration and to not expose how we have set this up internally.  The ConfigMap added is the public ca.crt.


hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -723,6 +754,63 @@ objects:
         - '*'
         verbs:
         - '*'
+    - apiVersion: v1
+      data:
+        ca.crt: '-----BEGIN CERTIFICATE-----
```
jharrington22: Is it just me or are there extra newlines here?

jewzaam: Source file uses literal style with `|`.. looking at best way to preserve this when adding to the template.

jewzaam: Tried dropping the SSS directly as-is (with extra \n) and it syncs correctly to the cluster.  So it's weird to look at, but it's valid and works.

scripts/templates/template.yaml
```diff
@@ -20,3 +36,35 @@ objects:
...
+          email:
+          - ${IDENTITY_ATTR_EMAIL}
+          id:
+          - ${IDENTITY_ATTR_ID}
+          name:
+          - ${IDENTITY_ATTR_NAME}
+          preferredUsername:
+          - ${IDENTITY_ATTR_PREFERRED_USERNAME}
+        bindDN: ${IDENTITY_BIND_DN}
+        bindPassword:
```
lisa: @jewzaam  Is this secure?

jewzaam: it's a reference to a secret

jewzaam: so it doesn't expose any password, just the name of the secret.  dno't see that being a big deal.  if you can get to the secret you can get to the oauth cr

PR #39, opened 2019-06-27
Add alerts for SRE operators being offline for 15 minutes

jewzaam:
https://bugzilla.redhat.com/show_bug.cgi?id=1724742

OLM on OCP 4.1.3 caused outage of dedicated-admin operator in stage.  This will at least alert us when this happens for our operators.

### DIFF
--- a/deploy/sre-prometheus/100-configure-alertmanager-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-configure-alertmanager-operator.PrometheusRule.yaml
@@ -1,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-configure-alertmanager-operator-offline-alerts
+    role: alert-rules
+  name: sre-configure-alertmanager-operator-offline-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-configure-alertmanager-operator-offline-alerts
+    rules:
+    - alert: ConfigureAlertmanagerOperatorOfflineSRE
+      expr: absent(up{service="configure-alertmanager-operator"})
+      for: 15m
+      labels:
+        severity: critical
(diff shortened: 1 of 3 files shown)


deploy/sre-prometheus/100-dedicated-admin-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-dedicated-admin-operator-alerts
+    role: alert-rules
+  name: sre-dedicated-admin-operator-alerts
```
rogbas: The prometheusRule deployed with d-a-o to monitor blacklisted namespaces is called `sre-dedicated-admin-alerts`
(https://github.com/openshift/dedicated-admin-operator/blob/master/build/templates/olm-artifacts-template.yaml.tmpl)

Wont this `sre-dedicated-admin-operator-alerts` create some confusion?

rogbas: What about renaming this one to `sre-dedicated-admin-operator-offline-alerts` ?

deploy/sre-prometheus/100-configure-alertmanager-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-configure-alertmanager-operator-alerts
+    role: alert-rules
+  name: sre-configure-alertmanager-operator-alerts
```
rogbas: For the same reason as with d-a-o, what about renaming that prometheusRule to `sre-configure-alertmanager-operator-offline-alerts`

deploy/sre-prometheus/100-dedicated-admin-operator.PrometheusRule.yaml
rogbas: You updated the Group, shouldn't we do the same with `metadata.Name` ?

PR #41, opened 2019-06-28
Implement SLA tracking and increase Prometheus retention

abyrne55:
Current usage on osd-v4stg is showing that 15d of data only generates about 320MB of data under /prometheus. We plan on increasing retention to 30d to support monthly SLA reporting, but even with increased retention and increased load, we shouldn't need more than 5-7GB. For safety, I set volume size to 10GB.


jeremyeder:
This PR would affect all OSDs 4.x environments.  If you're asking to expose retention as a tunable on OSD, that is perhaps a reasonable ask.  But changing for all is no good.

https://docs.okd.io/latest/scaling_performance/scaling_cluster_monitoring.html#cluster-monitoring-capacity-planning

aweiteka:
We're not seeing anywhere near those numbers. This PR is the only way to provide active alerts and exportable reporting via PagerDuty on 30d SLO/SLA. How do you propose we achieve those numbers?

jeremyeder:
@aweiteka I've read your comment several times, and it's still unclear to me how the proposed changes solve any alerting issues?

jeremyeder:
The 50Gi number was determined by @nmalik looking at production OSDs in 3.11 and starter clusters, so I think you should leave that alone.

The retention probably should become configurable so you can lay it down only on the clusters that you want.

abyrne55:
Got it, I'll close this then and look into making retention configurable

abyrne55:
Actually, I'd like to re-open this with the PVC size change reversed (so leaving volume size at 50GB, and only changing retention length to 30d). After speaking with @jharrington22, we decided we want 30d to be the default retention across all managed clusters, since all of the SLAs for those clusters work on a 28-day timetable.

Since this is only a change to a ConfigMap that's applied by a SelectorSyncSet to all clusters, it is technically configurable per-cluster, but at the moment any changes will be replaced. We'll have to look into fixing that behavior in a separate PR.

jeremyeder:
SLA/SLOs are calculated using data back-hauled from each cluster via telemeter-client.

The Telemeter and datahub infrastructures are where the retentions are set (correctly).  In-cluster retention is unrelated to SLA/SLO.

abyrne55:
Note: just added SLA tracking rules and generator script to this PR, since they should be deployed together

jewzaam:
A note on the data usage on existing clusters, we are not running any workloads of note.  Prometheus PV Size rational is documented here (sorry for internal link): https://docs.google.com/document/d/1y5AbbhNPyoPN8QXJkcqQJ6PWMjr3cKqOGDVfTdwn9j0/edit#heading=h.ldax5zt7icvg

abyrne55:
Closing this since we're switching to a different strategy for SLA/SLO tracking

PR #44, opened 2019-07-11
Install logging and ES operator and pre-configure curator

jewzaam:
In v4 we'll support the customer creating a ClusterLogging CR to opt-in to the logging stack running on their cluster.  It will run on the workers.  Ability to do this requires update to dedicated-admin permissions which will be coming soon.


fahlmant:
What happens if the namespace creation fails?

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -706,6 +706,151 @@ objects:
...
+      metadata:
+        labels:
+          csc-owner-name: installed-redhat-openshift-logging
+          csc-owner-namespace: openshift-marketplace
+        name: cluster-logging
+      spec:
+        channel: preview
+        installPlanApproval: Automatic
+        name: cluster-logging
+        source: installed-redhat-openshift-logging
```
rogbas: I don't see a catalogsource in this repo with that name, is it supposed to be pre-installed in the clusters?

jewzaam: Good catch, looking into this.  May be installed as part of installation via the console UI.  I will update this PR when I have an update.

jewzaam: Fixed.  OperatorHub in console drops a CatalogSourceConfig, which I have included now for the 2 operators.  Also had missed the OperatorGroup and had a broken Subscription for ES.

jeremyeder:
And the example CR in the [product docs](https://docs.openshift.com/container-platform/4.1/logging/efk-logging-deploying.html) is what we will point users to?

Any tweaks to the storage aspects necessary?

```
apiVersion: "logging.openshift.io/v1"
kind: "ClusterLogging"
metadata:
  name: "instance" 
  namespace: "openshift-logging"
spec:
  managementState: "Managed"  
  logStore:
    type: "elasticsearch"  
    elasticsearch:
      nodeCount: 3 
      storage:
        storageClassName: gp2
        size: 200G
      redundancyPolicy: "SingleRedundancy"
  visualization:
    type: "kibana"  
    kibana:
      replicas: 1
  curation:
    type: "curator"  
    curator:
      schedule: "30 3 * * *"
  collection:
    logs:
      type: "fluentd"  
      fluentd: {}
```

dak1n1:
Just wanted to remind you to run `make` if you haven't already. I do see one generated template file, so maybe that means you already ran it.

PR #45, opened 2019-07-11
Dockerfile update, No more need for epel

ALimaRedHat:
previous commit depended on epel  which was not available when testing at `openshift/release`, this multistage Dockerfile update hopefully remedies the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,21 @@
 FROM openshift/origin-cli:v3.11
 
+# Multistage with python
+FROM python:2.7
+
+# Bring oc binary to python image
+COPY --from=0 /bin/oc /bin/
+
 # Environment
 ARG IN_DOCKER_CONTAINER="true"
-ENV VENV=/venv
 ENV REPO_PATH=/managed-cluster-config
 
 # Copy repo into docker image:
 COPY . ${REPO_PATH}
 WORKDIR ${REPO_PATH}
 
-# Docker dependencies:
-RUN yum update -y && yum install -y \
-    git \
-    make \
-    epel-release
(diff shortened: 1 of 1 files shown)


ALimaRedHat:
@jewzaam could you please QC this for me?

Dockerfile
```diff
@@ -9,13 +15,6 @@ ENV REPO_PATH=/managed-cluster-config
...
-# Docker dependencies:
-RUN yum update -y && yum install -y \
-    git \
-    make \
-    epel-release
-RUN yum -y install python-pip && yum clean all
-
 # Upgrade pip and install necessasry packages
 RUN pip install \
     --upgrade pip \
```
rogbas: Now that we are using the python 2.7 docker image I don't see why using the venv.

Couldn't we remove it and go straight to `RUN make`?

Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM openshift/origin-cli:v3.11
 
+# Multistage with python
+From python:2.7
```
rogbas: Can we have this From is upper case too?

Dockerfile
```diff
@@ -9,13 +15,6 @@ ENV REPO_PATH=/managed-cluster-config
...
 
-# Docker dependencies:
-RUN yum update -y && yum install -y \
-    git \
-    make \
-    epel-release
-RUN yum -y install python-pip && yum clean all
-
 # Upgrade pip and install necessasry packages
 RUN pip install \
```
rogbas: Since pip is already installed in the img, I assume we don't need to install/upgrade it in the line below

PR #46, opened 2019-07-15
Branding fix

rafael-azevedo:
Updates branding to split the data to the two new CRs and updates the selectorsyncset template


lisa:
Why is this in two CRs with the same name?

rafael-azevedo:
> Why is this in two CRs with the same name?

No idea console operator now uses two CRs for this which is why we were having some problems to begin with. They point to different api versions 

consoles.config.openshift.io
consoles.operator.openshift.io

lisa:
> > Why is this in two CRs with the same name?
> 
> No idea console operator now uses two CRs for this which is why we were having some problems to begin with. They point to different api versions
> 
> consoles.config.openshift.io
> consoles.operator.openshift.io

Ooh, weird, and a bit confusing. This is valid YAML, but I don't know if it does what we need, so I'll defer to someone else for a semantic check.

cblecker:
If I was to take a guess, for non-OSD uses of the console, we don't necessarily need to expose all the operator settings (like setting OSD branding) for other OCP deployments. However, a post-logout redirect sounds like something they'd want to document and expose to everyone.

lisa:
I'd also like to hold this for a question:

Why did a change elsehwere impact us, and are we at risk of that in the future? I would have expected us to pin to a specific version that used the config that's in place already (before this PR changes it). But my understanding is that there was a change and this is the fix.

rafael-azevedo:
@jewzaam 
Before this PR we had the logoutRedirect in the wrong console CR. This meant that the lougoutRedirect url was not being applied since it needed to be in the other CR. Now on my test cluster I get redirected to the `"https://www.redhat.com/i/should/logout"` which is the behavior we wanted originally, but as it turns out that url does not exist

jewzaam:
@rafael-azevedo is this something that will be worked on or can we close this PR?

jewzaam:
Stale and not  directing to an existing page.  Closing.

PR #49, opened 2019-07-15
Manage telemeter server URL

jewzaam:
https://jira.coreos.com/browse/SREP-1594

So we can pass in a different telemeter instance for non-production OSD clusters.


jewzaam:
/hold

Requires changes to pass the new parameter.  Will reference this PR from that other PR.

PR #51, opened 2019-07-15
Add pods/exec permission for sre admin.

wanghaoran1988:


jewzaam:
Need to run `make` and commit the updated template.

wanghaoran1988:
@jewzaam Updated.

PR #56, opened 2019-07-17
Script SRE can use to manage upgrades of clusters.

jewzaam:
See README for how to use.
Note you have to run the script multiple times, so it's a less than ideal.  But it's a start..

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -109,7 +109,7 @@ spec:
     patchType: merge
 EOF
 
-    oc $CD_NAMESPACE create -f $SS_FILENAME || oc $CD_NAMESPACE replace -f $SS_FILENAME
+    oc -n $CD_NAMESPACE create -f $SS_FILENAME || oc -n $CD_NAMESPACE replace -f $SS_FILENAME
     rm -f $SS_FILENAME
     echo "Created SyncSet for '$CD_NAME' upgrade: $OCP_CURRENT_VERSION -> $OCP_VERSION_TO"
     done
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -20,7 +20,7 @@ NOW_EPOCH=$(date +"%s")
 CHANNEL_NAME=$(echo "${OCP_VERSION_TO}" | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
 GRAPH=$(curl -s -H "Accept: application/json" https://api.openshift.com/api/upgrades_info/v1/graph?channel="${CHANNEL_NAME}")
 GRAPH_INDEX_FROM=$(echo "${GRAPH}" | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_FROM\" ] | index(true)")
-GRAPH_INDEX_TO=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_TO\" ] | index(true)")
+GRAPH_INDEX_TO=$(echo "${GRAPH}" | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_TO\" ] | index(true)")
 
 GRAPH_EDGE=$(echo $GRAPH | jq -r ".edges[] | select(.[0] == $GRAPH_INDEX_FROM) | .[1] == $GRAPH_INDEX_TO" | grep true)
 
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -1,4 +1,4 @@
- #!/bin/bash
+#!/bin/bash
 
 OCP_VERSION_FROM=$1
 OCP_VERSION_TO=$2
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -94,7 +94,7 @@ do
         # always report status
         echo "done ($OCP_CURRENT_VERSION)"
 
-        if [ ! -z $OCP_VERSION_FROM ];
+        if [ -n $OCP_VERSION_FROM ];
         then
             # upgrade
 
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -19,7 +19,7 @@ NOW_EPOCH=$(date +"%s")
 # Verify we can actually upgrade (is this in the graph)
 CHANNEL_NAME=$(echo "${OCP_VERSION_TO}" | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
 GRAPH=$(curl -s -H "Accept: application/json" https://api.openshift.com/api/upgrades_info/v1/graph?channel="${CHANNEL_NAME}")
-GRAPH_INDEX_FROM=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_FROM\" ] | index(true)")
+GRAPH_INDEX_FROM=$(echo "${GRAPH}" | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_FROM\" ] | index(true)")
 GRAPH_INDEX_TO=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_TO\" ] | index(true)")
 
 GRAPH_EDGE=$(echo $GRAPH | jq -r ".edges[] | select(.[0] == $GRAPH_INDEX_FROM) | .[1] == $GRAPH_INDEX_TO" | grep true)
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -17,7 +17,7 @@ UPGRADE_DONE=""
 NOW_EPOCH=$(date +"%s")
 
 # Verify we can actually upgrade (is this in the graph)
-CHANNEL_NAME=$(echo $OCP_VERSION_TO | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
+CHANNEL_NAME=$(echo "${OCP_VERSION_TO}" | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
 GRAPH=$(curl -s -H "Accept: application/json" https://api.openshift.com/api/upgrades_info/v1/graph?channel=$CHANNEL_NAME)
 GRAPH_INDEX_FROM=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_FROM\" ] | index(true)")
 GRAPH_INDEX_TO=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_TO\" ] | index(true)")
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -19,13 +19,20 @@ NOW_EPOCH=$(date +"%s")
 # Verify we can actually upgrade (is this in the graph)
 CHANNEL_NAME=$(echo $OCP_VERSION_TO | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
 GRAPH_VERSION=$(curl -s -H "Accept: application/json" https://api.openshift.com/api/upgrades_info/v1/graph?channel=$CHANNEL_NAME | jq -r ".nodes[] | select(.version == \"$OCP_VERSION_TO\") | .version")
-
 if [ "$OCP_VERSION_TO" != "$GRAPH_VERSION" ];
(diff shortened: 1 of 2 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,136 @@
+ #!/bin/bash
+
+OCP_VERSION_FROM=$1
+OCP_VERSION_TO=$2
+
+if [ "$OCP_VERSION_TO" == "" ];
```
```suggestion
if [ -z $OCP_VERSION_TO ];
```
This is a bit more idiomatic Bash.

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,136 @@
...
+
+# Verify we can actually upgrade (is this in the graph)
+CHANNEL_NAME=$(echo $OCP_VERSION_TO | sed 's/\([^.]*\.[^.]*\)\..*/stable-\1/g')
+GRAPH=$(curl -s -H "Accept: application/json" https://api.openshift.com/api/upgrades_info/v1/graph?channel=$CHANNEL_NAME)
+GRAPH_INDEX_FROM=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_FROM\" ] | index(true)")
+GRAPH_INDEX_TO=$(echo $GRAPH | jq -r "[ .nodes[] | .version == \"$OCP_VERSION_TO\" ] | index(true)")
+
+GRAPH_EDGE=$(echo $GRAPH | jq -r ".edges[] | select(.[0] == $GRAPH_INDEX_FROM) | .[1] == $GRAPH_INDEX_TO" | grep true)
+
+if [ "$GRAPH_INDEX_FROM" == "" ];
```
```suggestion
if [ -z $GRAPH_INDEX_FROM ];
```

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,136 @@
...
+then
+    echo "Cannot upgrade from $OCP_VERSION_FROM, it is not available in upgrade graph."
+    exit 2
+fi
+if [ "$GRAPH_INDEX_TO" == "" ];
```
```suggestion
if [ -z $GRAPH_INDEX_TO ];
```

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,136 @@
...
+        elif [ "$SS_CURRENT_VERSION" == "" ] && [ "$OCP_CURRENT_VERSION" != "$OCP_VERSION_TO" ];
+        then
+            # report can't upgrade if version is not upgradable and upgrade is not in progress
+            echo "done ($OCP_CURRENT_VERSION)"
+            UPGRADE_NOT_POSSIBLE="${UPGRADE_NOT_POSSIBLE}$CD_NAME ($OCP_CURRENT_VERSION)\n"
+            continue
+        else
+            # cluster is already upgraded?
+            echo "done ($OCP_CURRENT_VERSION)"
+            if [ "$SS_CURRENT_VERSION" != "" ];
```
```suggestion
            if [ -z $SS_CURRENT_VERSION ];
```

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,136 @@
...
+            then
+                # delete the syncset, it's not needed anymore
+                oc -n $CD_NAMESPACE delete SyncSet $SS_NAME                
+                UPGRADE_DONE="${UPGRADE_DONE}$CD_NAME ($OCP_CURRENT_VERSION)\n"
+            fi
+            continue
+        fi
+
+        SS_FILENAME=$SS_NAME.syncset.yaml
+        rm -f $SS_FILENAME
```
cblecker: `rm -f` on not-fully-qualified filenames can be dangerous.. if this variable is set to something incorrect, you could damage your system.

jewzaam: Changed to put files in a temp dir then delete that dir at the end.  Could leave the temp dir alone if we don't want to delete at all, nothing sensitive about the files.

jewzaam:
@cblecker @lisa @mwoodson could I get another look post-changes?  thanks

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,138 @@
...
+
+# Verify we have ClusterDeployment CRs to work with
+if [ `oc get crd clusterdeployments.hive.openshift.io --no-headers 2>/dev/null | wc -l` == "0" ];
+then
+    echo "ERROR: Current cluster does not have Hive installed.  Verify where you are logged in."
+    exit 1
+fi
+
+TMP_DIR=`mktemp -d`
+
```
```suggestion
if [[ $? -ne 0 ]];
then
    echo "ERROR: Couldn't create a temporary directory. Aborting.
    exit 1
fi
```

Add a protection for a failure to create a temporary directory. An alternative is to `set -e` and/or `set -o nounset` earlier in the program.

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+fi
+
+UPGRADE_STARTED=""
+UPGRADE_PROGRESSING=""
+UPGRADE_NOT_POSSIBLE=""
+UPGRADE_DONE=""
+
+NOW_EPOCH=$(date +"%s")
+
+if [ ! -z $OCP_VERSION_FROM ];
```
```suggestion
if [ -n $OCP_VERSION_FROM ];
```
The opposite of `! -z` is `-n`, however, I don't think this code will ever be false because earlier you ensure that `$OCP_VERSION_FROM` is set. Thus, this will never be empty.

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+
+TMP_DIR=`mktemp -d`
+OCP_PATCH="{\"spec\":{\"desiredUpdate\": {\"force\": false, \"image\": \"\", \"version\": \"$OCP_VERSION_TO\"}}}"
+if [[ $? -ne 0 ]];
+then
+    echo "ERROR: Couldn't create a temporary directory. Aborting.
+    exit 1
+fi
+# kick off the upgrades (or status check)
+for CD_NAMESPACE in `oc get clusterdeployment --all-namespaces | awk '{print $1}' | sort | uniq`;
```
lisa: Might want to look into using `-o jsonpath` instead of piping through awk/sort/uniq? (also, `sort -u` == `sort | uniq`)

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+    for CD_NAME in `oc -n $CD_NAMESPACE get clusterdeployment -o json | jq -r '.items[] | select(.metadata.labels["api.openshift.com/managed"] == "true") | select(.status.installed == true) | select(.status.clusterVersionStatus.history[0].state == "Completed") | .metadata.name'`;
+    do  
+        echo -n "Checking $CD_NAME..."
+        # TODO figure out how to get the latest version, for now just getting the first one (ordered descending)
+        OCP_CURRENT_VERSION=`oc -n $CD_NAMESPACE get clusterdeployment $CD_NAME -o json | jq -r '.status.clusterVersionStatus.history[] | select(.state == "Completed") | .version' | head -n1`
+
+        SS_NAME=$CD_NAME-upgrade
+        SS_VERSION_FROM=`oc -n $CD_NAMESPACE get syncset $SS_NAME --show-labels --no-headers 2> /dev/null | sed 's|.*managed.openshift.io/from=\([^,]*\).*|\1|g'`
+        SS_VERSION_TO=`oc -n $CD_NAMESPACE get syncset $SS_NAME --show-labels --no-headers 2> /dev/null | sed 's|.*managed.openshift.io/to=\([^,]*\).*|\1|g'`
+        SS_CREATED_ON=$(date --date="$(oc -n $CD_NAMESPACE get syncset $SS_NAME --template '{{ .metadata.creationTimestamp }}' 2>/dev/null)" +"%s" 2>/dev/null)
```
lisa: Note that `date --date` is not portable. I don't know if it will come up, but in testing, if this is run on MacOS it could be an issue.

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+        SS_AGE_MIN=$(($(($NOW_EPOCH-$SS_CREATED_ON))/60))
+
+        # always report status
+        echo "done ($OCP_CURRENT_VERSION)"
```
lisa: Question - what does "done" mean or imply here? That the check has been finished?

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+
+        if [ ! -z $OCP_VERSION_FROM ];
```
```suggestion
        if [ -n $OCP_VERSION_FROM ];
```

Similar to before, `! -z` == `-n`, and this will always be true.

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
...
+                    UPGRADE_DONE="${UPGRADE_DONE}  $CD_NAME ($OCP_CURRENT_VERSION)\n"
+                fi
+            fi
+        fi
+    done
+done
+
+rm -rf $TMP_DIR
+
+if [ ! -z $OCP_VERSION_FROM ];
```
```suggestion
if [ -n $OCP_VERSION_FROM ];
```
As before, `! -z` == `-n`, and this will always be true.

scripts/cluster-upgrade.sh
jewzaam: There's a check now that will unset this var if the value starts with "status".

scripts/cluster-upgrade.sh
jewzaam: Yea, optimizations would be nice.  I use `uniq -c` sometimes so it's habit to do the `sort | uniq` instead of `sort -u`.  The formatting bits that oc gives are nice but don't come naturally to me.. tend to toss them at other things like awk, jq, yq..

scripts/cluster-upgrade.sh
jewzaam: You have to run this on the hive cluster to use it, so it's good enough in my view.  You could hit hive from MacOS if you tunnel through bastion I guess, but it's not the expected usage.

scripts/cluster-upgrade.sh
jewzaam: yup, a word I picked when hacking out the initial script and hasn't been changed since..

scripts/cluster-upgrade.sh
lisa: I totally get you here. I still struggle with `-o jsonpath`! I'm not aiming for minimal forking :) Just offering some suggestions that are "native" and ones from which we could all learn. This change is up to you to dive into, or not; I'll mark resolved.

scripts/cluster-upgrade.sh
lisa: Sounds good. In that case can I ask for a short comment at the top of the script saying that it's Linux only by design, to run on Hive cluster? I don't think we have to go through capabilities testing like we're autoconf. (I can make the suggested change for review)

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
+
```
```suggestion
# This script contains non-portable components and is intended to run on a Hive cluster.

```

scripts/cluster-upgrade.sh
```diff
@@ -0,0 +1,162 @@
...
+    # Verify we have ClusterDeployment CRs to work with
+    if [ `oc get crd clusterdeployments.hive.openshift.io --no-headers 2>/dev/null | wc -l` == "0" ];
+    then
+        echo "ERROR: Current cluster does not have Hive installed.  Verify where you are logged in."
+        exit 1
+    fi
+fi
+
+TMP_DIR=`mktemp -d`
+OCP_PATCH="{\"spec\":{\"desiredUpdate\": {\"force\": false, \"image\": \"\", \"version\": \"$OCP_VERSION_TO\"}}}"
```
lisa: You might want to move this `OCP_PATCH` up before the `TMP_DIR` assignment to remove any worry about the `$?` referring to this assignment and not `mktemp`.

jewzaam:
@lisa note if using -n $var need to use `[[ -n $var ]]`

lisa:
> @lisa note if using -n $var need to use `[[ -n $var ]]`

Is `[ -z $var ]` ok though?

jewzaam:
> > @lisa note if using -n $var need to use `[[ -n $var ]]`
> 
> Is `[ -z $var ]` ok though?

No, that might as well be `[ true ]` though I didn't dig into the why.. just needed a solution to make it work.   `[ ! -z $var ]` works too.

jewzaam:
> > > @lisa note if using -n $var need to use `[[ -n $var ]]`
> > 
> > 
> > Is `[ -z $var ]` ok though?
> 
> No, that might as well be `[ true ]` though I didn't dig into the why.. just needed a solution to make it work. `[ ! -z $var ]` works too.

Missed this is asking -z, not -n... I haven't checked.  Given there's a working solution I'm in favor of stopping.  Not really keen on learning all the nuances of these options in bash tbh.

PR #57, opened 2019-07-22
Attempt to document resources -> SelectorSyncSet -> Template flow

cben:
Hi, I was curious how the yamls actually make it into the cluster (I think it's useful to understand for other teams involved around OSD).
Here is my attempt at reverse engineering — please review carefully :-)

Questions:

- I wanted to add step "3. then we do `oc process $TEMPLATE | oc apply` into the Hive cluster" or something like that but didn't find any details...   How, and *when* is the resulting template actually deployed?  
  `hack/app_sre_build_deploy.sh` does nothing.
  Is every merged PR automatically deployed?  Is it deployed on MR to `saas-osd-operators` repo bumping the hash?  Is it manual?  This is worth explaining in the README.

- Are all existing clusters affected when you change/add yamls here?  I assume yes (unless one uses patch `applyMode: ApplyOnce`), but it's worth explicitly saying in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,26 @@ This repo contains static configuration specific to a "managed" OpenShift Dedica
 
 ## How to use this repo
 
-To add a new SelectorSyncSet, add your yaml manifest to the `deploy` dir, then run the `make` command.
+Yaml files are taken from [./deploy/](./deploy/) dir and deployed through couple layers of indirection:
+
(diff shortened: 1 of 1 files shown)


cben:
Another question: IIUC plans from https://jira.coreos.com/browse/SREP-1594, the SelectorSyncSets for the different environments were supposed to be differentiated by matching on `api.openshift.com/environment` label.
I don't see this label used.  What separates the different environments?
Are they simply separate Hive clusters?

cben:
/remove-lifecycle rotten

Still hoping to see this documented, please review :)  Don't know if still accurate (or even if it was accurate in the first place).

PR #60, opened 2019-07-29
RBAC so SRE can administer logging

jewzaam:
Added rbac so SRE can manage logging.
Generated template and swapped over to oyaml for sorted output.

https://jira.coreos.com/browse/SREP-1838


jewzaam:
NOTE the template probably changed a lot but won't moving forward.  Switched over to `oyaml` module which sorts things and is a drop in for `pyyaml`.

PR #63, opened 2019-08-01
SREP-1846

robotmaxtron:
Updating the namespace, as per SREP-1846


deploy/sre-prometheus/100-machine-api.PrometheusRule.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     prometheus: sre-machine-api-node-mismatch
     role: alert-rules
   name: sre-machine-api-node-mismatch
-  namespace: openshift-monitoring
+  namespace: openshift-machine-api
```
jewzaam: /hold
All rules land in openshift-monitoring.  Not sure this will work if we deploy to a different namespace.
The `namespace` label will be set by the exporter, see https://github.com/openshift/managed-prometheus-exporter-machine-api/blob/master/monitor/main.py

PR #64, opened 2019-08-02
Always cleanup syncset, don't log error for oc delete

jewzaam:
Just annoying output at this time when running the script, this cleans up state in hive better and doesn't report useless errors in script output.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -143,13 +143,9 @@ EOF
                 # report can't upgrade if version is not upgradable and upgrade is not in progress
                 UPGRADE_NOT_POSSIBLE="${UPGRADE_NOT_POSSIBLE}  $CD_NAME ($OCP_CURRENT_VERSION)\n"
             else
-                # cluster is already upgraded?
-                if [ -z $SS_VERSION_FROM ];
-                then
-                    # delete the syncset, it's not needed anymore
-                    oc -n $CD_NAMESPACE delete SyncSet $SS_NAME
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -143,13 +143,9 @@ EOF
...
             else
-                # cluster is already upgraded?
-                if [ -z $SS_VERSION_FROM ];
-                then
-                    # delete the syncset, it's not needed anymore
-                    oc -n $CD_NAMESPACE delete SyncSet $SS_NAME
-                    UPGRADE_DONE="${UPGRADE_DONE}  $CD_NAME ($OCP_CURRENT_VERSION)\n"
-                fi
+                # delete the syncset, it's not needed anymore
+                oc -n $CD_NAMESPACE delete SyncSet $SS_NAME 2>/dev/null
```
jharrington22: Quotes :slightly_smiling_face:

jharrington22: What are the errors? Can we suppress them in any other way that doesn't mask other potential unknown errors?

jewzaam:
```suggestion
                oc -n "$CD_NAMESPACE" delete SyncSet "$SS_NAME" 2>/dev/null
```

jewzaam: Error is the syncset doesn't exist.  We could do better error handling but my expectation is we'll throw this away some day so we can automate the upgrade process.  We can deal w/ fancy then.

PR #66, opened 2019-08-02
Create an alert for the cluster operator being down

fahlmant:
Creates an alert for `cluster_operator_up == 0 for more than 15 minutes` from SREP-1848


jewzaam:
@fahlmant need to run `make` to update template then PR check will pass

PR #67, opened 2019-08-06
SRE can temporarily elevate permissions to cluster-admin

jewzaam:
https://jira.coreos.com/browse/SREP-1805

Added:
- Group: osd-sre-cluster-admins
- ClusterRoleBinding for the group to existing ClusterRole cluster-admin

As noted in README, SRE can add themselves to the `osd-sre-cluster-admins` group to temporarily get elevated access.


deploy/sre-authorization/00-cluster-admins.ClusterRoleBinding.yaml
```diff
@@ -0,0 +1,13 @@
+
```
cblecker: extra white space?

README.md
```diff
@@ -27,7 +27,9 @@ Persistent storage is configured using the configmap `cluster-monitoring-config`
 
 ## SRE Authorization
 
-Instead of SRE having `cluster-admin` a new ClusterRole is created with some permissions removed.  The ClusterRole can be regenerated in the `generate/sre-authorization` directory.
+Instead of SRE having `cluster-admin` a new ClusterRole `osd-sre-admin` is created with some permissions removed.  The ClusterRole can be regenerated in the `generate/sre-authorization` directory.  The role is granted to SRE via `osd-sre-admins` group.
```
lisa:
```suggestion
Instead of SRE having the `cluster-admin` role, a new ClusterRole, `osd-sre-admin`, is created with some permissions removed.  The ClusterRole can be regenerated in the `generate/sre-authorization` directory.  The role is granted to SRE via `osd-sre-admins` group.
```

Editing

README.md
```diff
@@ -27,7 +27,9 @@ Persistent storage is configured using the configmap `cluster-monitoring-config`
...
+
+To elevate privileges, SRE can add themselves to the group `osd-sre-cluster-admins` which is bound to the ClusterRole `cluster-admin`.  When this group is created and managed by Hive, all users are wiped because the SelectorSyncSet will always have `users: null`.  Therefore, SRE will get elevated privileges for a limited time.
```
lisa:
```suggestion
To elevate privileges, SRE can add themselves to the group `osd-sre-cluster-admins`, which is bound to the ClusterRole `cluster-admin`.  When this group is created and managed by Hive, all users are wiped because the SelectorSyncSet will always have `users: null`.  Therefore, SRE will get elevated privileges for a limited time.
```

jewzaam:
> Relatively minor editing for clarity.
> 
> Overall, what is the chance that Hive will blow away my privilege escalation in the middle of maintenance once the cache expires?

Well, all depends on when the cache expires on hive.  Linked story has this comment: `ne possible downside is Hive could reconcile very soon after the update, rbac could be reverted while SRE is doing things.  But my theory is if SRE needs to do a lot of things as cluster-admin there's something big that is wrong and either it shouldn't be done as a one-off, it should be fixed upstream, or if really necessary the kubeadmin creds could be accessed.  More likely SRE just re-adds to the group and keeps going?`

lisa:
/lgtm

I feel like this is a good stopgap even with the potential to reconcile permissions away mid-operation.

PR #68, opened 2019-08-06
Apply curated operators

jewzaam:
https://jira.coreos.com/browse/SREP-1370

Make shipped OperatorSources unmanaged then patch those OperatorSources.


jewzaam:
blocks merging https://github.com/openshift/dedicated-admin-operator/pull/69

PR #69, opened 2019-08-07
Add patch to wipe users from hive

jewzaam:
The group itself isn't reapplyed yet (https://jira.coreos.com/browse/CO-528) so this is a stopgap with the intent to always apply the patch which is expected.


jewzaam:
/hold
Testing in stage right now on a single cluster.

jewzaam:
/hold cancel
Confirmed the patch works in a one-cluster test.  Probe time on the syncset corresponds to users being wiped from group `osd-sre-cluster-admins` group.

PR #73, opened 2019-08-08
Add getopts and usage

jharrington22:
@jewzaam @mwoodson my attempt to make script a little safer. With these changes you can run the script without arguments and it will print the usage (still) but adding a specific option for status and upgrade versions. Thoughts?

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -1,19 +1,76 @@
-#!/bin/bash
-# This script contains non-portable components and is intended to run on a Hive cluster.
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
 
-OCP_VERSION_FROM=$1
-OCP_VERSION_TO=$2
+STATUS=false
+SCRIPT_NAME=$(basename "$0")
 
-if [ -z $OCP_VERSION_FROM ];
-then
-    echo "This script can be used to get the status of all managed clusters or to upgrade clusters."
-    echo "usage: $0 status"
-    echo "usage: $0 <from> <to>"
-    echo "  example: $0 4.1.0 4.1.2"
+usage() {
+    cat <<EOF
+    This script contains non-portable components and is intended to run on a Hive cluster.
+
+    usage: $SCRIPT_NAME [ OPTION ]
+    Options
+    -f         [ VERSION ] OSD version to upgrade from
+    -t         [ VERSION ] OSD version to upgrade to
+    -s         Status of cluster versions
+
+    Example upgrade:
+      $0 -f 4.1.0 -t 4.1.2
+
+    Example status:
+      $0 -s
+EOF
+}
+
+if ( ! getopts ":t:f:sh" opt); then
+    cat <<EOF
+
+    $SCRIPT_NAME requries an argument!
+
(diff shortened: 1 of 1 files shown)


jharrington22:
I've tested the status option but not the upgrade.

scripts/cluster-upgrade.sh
```diff
@@ -1,19 +1,75 @@
 #!/bin/bash
-# This script contains non-portable components and is intended to run on a Hive cluster.
 
-OCP_VERSION_FROM=$1
-OCP_VERSION_TO=$2
+STATUS=false
```
```suggestion
STATUS=false
set -e
```

little safer

jharrington22: Good point :)

jharrington22: @lisa fixed

scripts/cluster-upgrade.sh
```diff
@@ -1,19 +1,75 @@
-#!/bin/bash
-# This script contains non-portable components and is intended to run on a Hive cluster.
+#!/bin/bash -e
```
cblecker:
```suggestion
#!/usr/bin/env bash

set -o errexit
set -o nounset
set -o pipefail
```

scripts/cluster-upgrade.sh
```diff
@@ -1,19 +1,75 @@
...
 
-OCP_VERSION_FROM=$1
-OCP_VERSION_TO=$2
+STATUS=false
+SCRIPT_NAME=$(basename "$0")
+
+usage() {
+    cat <<EOF
```
cblecker: Reason for cat > echo?

jharrington22: Its cleaner, otherwise I'd have to echo every line or string together `\n`'s. Do you think?

jharrington22:
I'll squash before merge, so please don't lgtm @lisa @cblecker

scripts/cluster-upgrade.sh
cblecker: Oh yeah, you're using a here doc. Yeah, that's fine then 👍

scripts/cluster-upgrade.sh
```diff
@@ -1,19 +1,78 @@
...
+        t)
+            OSD_UPGRADE_TO="$OPTARG" >&2
+            ;;
+        s)
+            STATUS=true >&2
+            ;;
+        h)
+            echo "Invalid option: -$OPTARG" >&2
+            usage
+            exit 1
```
lisa: I don't think showing help should be an error?

jharrington22: Good pickup!

scripts/cluster-upgrade.sh
```diff
@@ -1,19 +1,76 @@
...
+        :)
+            echo "$0 Requires an argument" >&2
+            usage
+            exit 1
+            ;;
+        esac
+    done
+
+
+OCP_VERSION_FROM=OSD_UPGRADE_FROM
```
cblecker: Why do we set this as one name from getops, and then set it to the exact same thing down here? Seems redundant.

jewzaam:
@jharrington22 still something you'd like to merge?

PR #23, opened 2019-05-30
SREP-1368: Support quotas & exemptions

lisa:
`make all` will render two kinds of `SelectorSyncSet` manifests, 1) the default quotas to be applied and 2) exemptions to them.

Includes hepler awk script in `scripts/jsonify.awk` to handle the json patch fragments in a lightweight way.

Note: This should be merged after https://github.com/openshift/managed-cluster-config/pull/24 as #24 is making significant structural changes, and changes which enable this (notably the object patching). Prior to merging the numerous commits herein should be squashed.

Signed-off-by: Lisa Seelye <lseelye@redhat.com>

PR #74, opened 2019-08-09
added logging quota overrides from the default, and added a specific quota targeting the openshift-logging namespace so that the customer may provision ES with PV's.

mwoodson:


deploy/osd-logging/03-storage-quota.yaml
```diff
@@ -0,0 +1,8 @@
+- apiVersion: v1
+  kind: ResourceQuota
+  metadata:
+    name: logging-storage-quota
+    namespace: openshift-logging
+  spec:
+    hard:
+      requests.storage: "600G"
```
lisa:
```suggestion
      requests.storage: "600Gi"
```

I believe the unit is `Gi`, not `G`?

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -1062,8 +1062,18 @@ objects:
...
           openshift.io/cluster-logging: 'true'
           openshift.io/cluster-monitoring: 'true'
+    - apiVersion: v1
+      kind: ResourceQuota
+      metadata:
+        name: logging-storage-quota
+        namespace: openshift-logging
+      spec:
+        hard:
+          requests.storage: "600G"
```
lisa:
```suggestion
          requests.storage: "600Gi"
```

Same as before about units.

deploy/osd-logging/01-namespace.yaml
```diff
@@ -3,7 +3,9 @@ kind: Namespace
 metadata:
   name: openshift-logging
   annotations:
-    openshift.io/node-selector: "" 
+    openshift.io/node-selector: ""
   labels:
+    managed.openshift.io/service-lb-quota-exempt: "true"
```
lisa: Does the namespace need to be exempt for load balancers too?

lisa:
/cancel hold

Removing my hold; though I still think we should re-use the framework for managing resource quotas that was added in #23, even though it will require extending it beyond its original intent.

jewzaam:
/hold cancel
@lisa had wrong syntax

deploy/osd-logging/01-namespace.yaml
jewzaam:
```suggestion
```

mwoodson: This is following the pattern of other namespaces having this exemption as well.  I followed that pattern here.

lisa:
> /hold cancel
> @lisa had wrong syntax

Oops. Silly bot, why can't it do what I meant, and not what I told it to do? 😇

PR #78, opened 2019-08-20
Change upgrade script to add capacity to OSD cluster as part of upgrade

jewzaam:
Script no longer uses syncset to patch the cluster.  It adds nodes, upgrades, then tears down nodes.  The script can be cancelled and restarted w/o any issue but needs to be running for everything to complete.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -6,6 +6,7 @@ trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
 
 OCP_VERSION_FROM=$1
 OCP_VERSION_TO=$2
+CLUSTER_NAMES=${@:3}
 
 TMP_DIR=$(mktemp -d)
 if [[ $? -ne 0 ]];
@@ -14,12 +15,15 @@ then
     exit 1
 fi
 
-if [ -z $OCP_VERSION_FROM ];
+if [ -z $CLUSTER_NAMES ];
 then
-    echo "This script can be used to get the status of all managed clusters or to upgrade clusters."
-    echo "usage: $0 status"
-    echo "usage: $0 <from> <to>"
-    echo "  example: $0 4.1.0 4.1.2"
+    echo "This script is used to upgrade clusters from a version to another."
+    echo "Optionally can target a subset of clusters by name."
+    echo "When upgrading all, must specify 'all' for cluster name."
+    echo ""
+    echo "usage: $0 <from> <to> [cluster name1] .. [cluster nameN]"
+    echo "  example: $0 4.1.0 4.1.2 all"
+    echo "  example: $0 4.1.0 4.1.2 nmalik1 nmalik2"
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -20,12 +34,140 @@ then
...
+    
+    echo "$(date "+%Y-%m-%d_%H.%M.%S") - $CD_NAME - $STAGE - $MESSAGE"
+}
+
+setup() {
+    CD_NAMESPACE=$1
+    CD_NAME=$2
+
+    # get kubeconfig so we can check status of cluster's nodes (extra capacity)
+    oc -n $CD_NAMESPACE get secrets "${CD_NAME}-admin-kubeconfig" -o jsonpath='{.data.kubeconfig}' | base64 -d > $TMP_DIR/kubeconfig-$CD_NAME
```
clcollins:
```suggestion
    oc -n $CD_NAMESPACE extract secret/"${CD_NAME}-admin-kubeconfig" --keys=kubeconfig --to=-
```

clcollins: Doesn't really change anything but avoids the need to use base64.  Just a suggestion.

scripts/cluster-upgrade.sh
```diff
@@ -73,91 +215,34 @@ then
...
-if [[ $? -ne 0 ]];
-then
-    echo "ERROR: Couldn't create a temporary directory. Aborting."
-    exit 1
-fi
-
-# kick off the upgrades (or status check)
 for CD_NAMESPACE in `oc get clusterdeployment --all-namespaces | awk '{print $1}' | sort | uniq`;
 do  
     for CD_NAME in `oc -n $CD_NAMESPACE get clusterdeployment -o json | jq -r '.items[] | select(.metadata.labels["api.openshift.com/managed"] == "true") | select(.status.installed == true) | select(.status.clusterVersionStatus.history[0].state == "Completed") | .metadata.name'`;
```
clcollins: This would be so much more efficient if it looped through all the cluster deployments in --all-namespaces and selected on the ones desired, then extracted .metadata.namespace along with .metadata.name, but I cannot for the life of me figure out how to get jq to nicely return the namespace and name in a way that can be used with bash.  ¯\_(ツ)_/¯

scripts/cluster-upgrade.sh
```diff
@@ -20,12 +34,140 @@ then
...
+    
+    log $CD_NAME "upgrade" "Checking $CD_NAME..."
+
+    # - do we need to upgrade?
+    OCP_CURRENT_VERSION=$(oc -n $CD_NAMESPACE get clusterdeployment $CD_NAME -o json | jq -r '.status.clusterVersionStatus.history[] | select(.state == "Completed") | .version' | head -n1)
+
+    if [ "$OCP_CURRENT_VERSION" == "$TO" ];
+    then
+        log $CD_NAME "upgrade" "skipping, already on version $TO"
+        teardown $CD_NAMESPACE $CD_NAME
```
clcollins: Is this teardown necessary?  It looks like it's only scaling down any extra replicas that were added as part of this process but that doesn't happen until further down in the upgrade function when setup is called, I think?

scripts/cluster-upgrade.sh
jewzaam: You could get them both but you'd have to parse the result.

```
oc -n $CD_NAMESPACE get clusterdeployment -o json | jq -r '.items[] | select(.metadata.labels["api.openshift.com/managed"] == "true") | select(.status.installed == true) | select(.status.clusterVersionStatus.history[0].state == "Completed") | .metadata.namespace "," .metadata.name'
```

Maybe there's another way but it's not worth spending time on in my opinion.

scripts/cluster-upgrade.sh
jewzaam: It's there to make sure that teardown is done.  If the label on the clusterdeployment isn't there it will not make any changes.

scripts/cluster-upgrade.sh
clcollins: :+1:

PR #79, opened 2019-08-21
Create alert if telemeter-client isn't forwarding data to telemeter server

jewzaam:
https://jira.coreos.com/browse/SREP-1889

SOP is already published.


lisa:
@jewzaam Does the SOP have the same name as the alert (as is the previous pattern)?

jewzaam:
> @jewzaam Does the SOP have the same name as the alert (as is the previous pattern)?

yes

PR #89, opened 2019-08-29
Provide SRE permissions to create needed velero objects

cblecker:


jewzaam:
Why does SRE in normal day-to-day support of a cluster need this permission added?

cblecker:
Backup creation: SRE may want to kick off a specific backup before taking an action.
Restore creation: SRE may need to restore specific data from a backup.
*requests objects: These are meta objects that velero uses to start specific operations, such as downloading a backup for inspection.

All of these I would expect to be normal operations.

deploy/sre-authorization/01-osd-sre-admin.ClusterRole.yaml
```diff
@@ -135,3 +135,19 @@ rules:
   - certificatesigningrequests/approval
   verbs:
   - update
+# SRE can download backups, review logs, and create/delete backups/restores
+- apiGroups:
+  - velero.io
+  resources:
+  - backups
+  - restores
```
jewzaam:
```suggestion
```

We should not allow restore by default.  It's not a typical use case for SRE, we have only ever restored from backup once in v3.  As the configuration stands right now my understanding is restore would not work.  There are openshift-isms that need to be addressed.  We'll be looking to what the app migration team is doing with restore.  As a fallback, SRE can elevate privileges to do a restore if required.

cblecker: I'm fine with removing restore at this time.

cblecker:
@jewzaam Removed restores.

PR #92, opened 2019-09-11
🎨: Refactor vercomp and fix minor :bug:

lisa:
This commit is largely optional, and the byproduct of a 10 minute rabbit
hole while reviewing the upgrade script in detail as part of SDE-170.

If applied, this will force the bin/bash dependency (already present in
the shebang) due to the `<(...)` input redirection.

Technically, this takes advantage of `sort(1)`'s `-v|--version-sort`
feature since it does all the heavy lifting and handles all kinds of
version strings. The interface has not changed at all, just the inner
workings.

Given two versions, `requested` (`1.2.3`) and `default` (`1.2.2`), we
should expect a return value of 1 and when that happens, we know that
when the two versions are sorted among themselves, `default` will come
first and `requested` will come last:

```
1.2.2 <-- default
1.2.3 <-- requested
```

Thus, if we look at the last line of the sort and find that it is the
same as the requested version, we know it's too far ahead of the default
version and should return a 1 to block the operation. If the two
versions are equal, we have a special case where that is allowed, and
thus return 0.

The :bug: fix is this:

When `requested` = `1.2.3-r2` and `default` = `1.2.3-r1`, we can see
that r2 is more recent than r1 and it should be disallowed based on the
previous logic, however, the previous version will return a 0 to allow
the operation. By utilising `sort(1)`, this behaviour, and other oddball
version strings can be caught.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -43,15 +43,15 @@ log() {
 }
 
 # Compares two version strings together to determine which is more recent.
-# First parameter is the candidate version
-# Second parameter is the check version.
+# First parameter is the requested version
+# Second parameter is the max allowed version.
 # If the candidate > check, or either version are omitted, return 1. Otherwise return 0
 # Functionally this is used to determine if the requested (candidate) version exceeds the maximum allowed version (check).
 # This relies on Bash (due to input redirection), and sort to do the comparing
 vercomp() {
-    local requested=$1 default=$2 rescheck=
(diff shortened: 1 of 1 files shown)


cblecker:
- Use of "default" and "requested" makes this confusing.
- This method isn't actually doing a "compare", so things like providing two inputs that are the same passes this check

Error checking in bash is extremely difficult. I'm tempted to just re-write this whole thing in go, tbh.

PR #94, opened 2019-09-14
Update oc image for pruning jobs

cblecker:
We were using an old v4.0.0 image tag for these pruning jobs. We should use the updated/patched v4.1 tag.


jewzaam:
/lgtm
need to make sure this is verified in stage on a new cluster before promoting to production

PR #95, opened 2019-09-16
Revert "Update oc image for pruning jobs"

jewzaam:
Reverts openshift/managed-cluster-config#94

It's broken in stage.

And I cannot pull the image locally either.

```bash
$ docker pull quay.io/openshift/origin-cli:v4.1
Error response from daemon: manifest for quay.io/openshift/origin-cli:v4.1 not found: manifest unknown: manifest unknown
```


lisa:
Seems like it should have been `quay.io/openshift/origin-cli:4.1` (an errant `v` before the version)?

PR #98, opened 2019-09-16
Use openshift/cli imagestream for pruner

cblecker:
Had one more look at this, and turns out there is an internal imagestream for the cli:
```
$ oc get imagestream -n openshift cli --output=custom-columns="NAME:.metadata.name,IMAGE REPOSITORY:.status.dockerImageRepository,TAGS:.status.tags[].tag"
NAME   IMAGE REPOSITORY                                                 TAGS
cli    image-registry.openshift-image-registry.svc:5000/openshift/cli   latest
```

Using this is best case, as this ensures the oc version always matches that of the cluster.

/assign @jewzaam @lisa
/hold

### DIFF
--- a/deploy/sre-pruning/110-pruning-builds.CronJob.yaml
+++ b/deploy/sre-pruning/110-pruning-builds.CronJob.yaml
@@ -17,7 +17,7 @@ spec:
           restartPolicy: Never
           containers:
           - name: builds-pruner
-            image: quay.io/openshift/origin-cli:4.1
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
             imagePullPolicy: Always
             command:
             - oc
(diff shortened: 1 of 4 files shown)


jewzaam:
@cblecker nice find.  Verified it works on my test cluster.  Only thought is what happens if the cli changes?  It's unlikely but possible.  Might be more possible using this method than pinning to a specific minor version.  Maybe we should look at an e2e test to verify prune functionality?

cblecker:
@jewzaam Couple ways to approach that problem:
- e2e test to verify that this job specifically does what it says it does (a good idea, even if we also do the other, perhaps easier ideas)
- if the syntax for oc changes, I expect that we'd get JobFailed alerts as the container would exit non-zero
- if the oc command somehow still exits 0, we might create a prometheus alert to alert on the fact that there are builds, etc that aren't getting pruned back. That kind of monitoring would cover any possible reason that the pruning isn't working.

All of that said, I think this is still the right move to make as if we stuck with a specific oc version that we knew the cli syntax wouldn't change, we could also have a change in the api behaviours behind the command if the oc version drifts. Given that, I think it's better to stick closely with the cluster version.

jewzaam:
Going to test out if we get alerts for breaking jobs, but that shouldn't hold this PR.
/lgtm

PR #99, opened 2019-09-17
Upgrade script supports variable 'compute' names in clusterdeployment

jewzaam:
It's not the best but it was a quick update.  Right now it's going to assume the compute name is prefixed with "worker".


cblecker:
/lgtm

This is fine for a quick update.

PR #100, opened 2019-09-17
If prune job doesn't start, replace it

cblecker:
When testing the prune cronjob image, I found that if for some reason the prune Job/Pod doesn't actually start (for example, stuck in `ContainerCreating` or `ErrImagePull`), the job will get stuck indefinitely, even if the CronJob spec is later updated to correct it.

Changing the `concurrencyPolicy` to replace, will ensure that if the Job/Pod doesn't start/finish, when the next Job starts in 1 hour, the pod will be replaced.

This also changes the CronJob to run for 2 hours instead of 1, so that if there are problems running the job, alerts will trigger before the Job/Pod is replaced.

/assign @jewzaam @lisa


jewzaam:
/hold
Want to make sure we get an alert when using this if it isn't successful.  Testing this morning, will update in a bit.  Note we get critical alert `KubePodNotReady` for failed jobs with `concurrencyPolicy: Forbid`

jewzaam:
`KubePodNotReady` never goes active because the pod is replaced.  Job is also replaced.  Recommend we create an alert on the cronjob being active for a long time based on `kube_cronjob_status_active{namespace="openshift-sre-pruning"}` before we merge this (or as a part of this PR).

cblecker:
There is a `KubeJobCompletion` alert:
```
alert: KubeJobCompletion
expr: kube_job_spec_completions{job="kube-state-metrics",namespace=~"(openshift-.*|kube-.*|default|logging)"}
  - kube_job_status_succeeded{job="kube-state-metrics",namespace=~"(openshift-.*|kube-.*|default|logging)"}
  > 0
for: 1h
labels:
  severity: warning
annotations:
  message: Job {{ $labels.namespace }}/{{ $labels.job_name }} is taking more than
    one hour to complete.
```

But it won't trigger because we are running the pruner every hour. Perhaps we copy it and set it to a less value? Or perhaps we just prune less often?

jharrington22:
@cblecker @jewzaam Don't we get an alert if a pod/job doesn't start due to ErrImagePull or ContainerCreating?

jewzaam:
@cblecker it's a warning, maybe that's ok.  I prefer critical but if it will get to PD and alert someone it's fine.  But with `Replace` the job is recreated which resets the timer for the alert to fire (for: 1h).  So we might see this alert briefly if timing is just right.

@jharrington22 we will with the current setup, but moving to `Replace` we will not because the job and pod are recreated, resetting the start of the alert.  There is no alert against a long active cronjob.

cblecker:
@jharrington22 We should via the `KubePodNotReady` critical alert, but again, it's an "over an hour" time frame alert.

cblecker:
Perhaps we just increase the cron to only run every 2 hours, and then the default alerts will all work as intended, but we still get the benefits of being able to self-heal the Jobs if the CronJob itself is fixed?

cblecker:
The other way to do it would be to look at how long has it been since the last successful run of the job.

jewzaam:
> Perhaps we just increase the cron to only run every 2 hours, and then the default alerts will all work as intended, but we still get the benefits of being able to self-heal the Jobs if the CronJob itself is fixed?

I see that as a stop-gap solution.  We can't assume any cronjob we create will follow the same pattern.

cblecker:
@jewzaam Can we move to 2 hours for now, and then open a JIRA card to research creating a prometheus metric for "time since last cron job success"? I tried to look into what it would take to create this metric, and it's non-trivial, but looks possible.

jewzaam:
@cblecker proposed update in https://github.com/cblecker/managed-cluster-config/pull/1

PR #106, opened 2019-09-24
Update managed-velero-operator prometheus role

cblecker:
Adds a R/RB for prometheus monitoring of the managed-velero-operator

This was added to the operator in https://github.com/openshift/managed-velero-operator/pull/18


cblecker:
I have tested this on my local staging cluster

PR #107, opened 2019-09-26
SREP-1932: Quote the quota limits

lisa:
With 4.2, it is required to have integers quoted. I have quoted the
storage quota (which is implicitly a string) to be explicit about its
typing.

### DIFF
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -1192,7 +1192,7 @@ objects:
               operator: DoesNotExist
         quota:
           hard:
-            services.loadbalancers: 2
+            services.loadbalancers: '2'
 - apiVersion: hive.openshift.io/v1alpha1
   kind: SelectorSyncSet
   metadata:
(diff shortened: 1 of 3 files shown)


hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -1192,7 +1192,7 @@ objects:
               operator: DoesNotExist
         quota:
           hard:
-            services.loadbalancers: 2
+            services.loadbalancers: '2'
```
jewzaam:
```suggestion
            services.loadbalancers: "2"
```

For consistency only, double quotes instead of single quotes.

lisa: Very strange. The template has double quotes, so I don't understand why the auto-generated file has single quotes.

lisa: @jewzaam I'm not going to apply your suggestion because in the rendered yaml file each other string has single quotes around it. Elsewhere, in non-generated files, we use double quotes, so it seems we have a difference of opinion compared to Python's YAML libraries.

jewzaam: Ah, fair enough.  Didn't notice it was the generated template. :+1:

PR #109, opened 2019-09-30
Various updates to cluster-upgrade.sh

jewzaam:
There's nothing from this script that should be actionable.  Even critical alerts only require action if paged.
    Upgrade script output only labels real errors as 'ERROR'
    Critical alerts are 'WARNING'
    Everything else is 'INFO'

Small things.
    Upgrade script now uses api (ocm) label for the name of the cluster.
    Log output removes trailing comma (was annoying me)

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -35,11 +35,13 @@ then
 fi
 
 log() {
-    CD_NAME=$1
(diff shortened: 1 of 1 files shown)


cblecker:
/lgtm
/hold

@jewzaam Feel free to `/hold cancel` when you're ready for this to merge.

PR #110, opened 2019-10-01
Update value of 'custom-kubelet' label to empty string ("")

jewzaam:
Why:
https://jira.coreos.com/browse/SREP-2164
https://bugzilla.redhat.com/show_bug.cgi?id=1755568

Related:
https://bugzilla.redhat.com/show_bug.cgi?id=1757195


deploy/kubelet-config/03-kubelet-config.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   machineConfigPoolSelector:
     matchLabels:
-      managed.openshift.io/custom-kubelet: true
```
lisa: Is this the real meat of the change, with other parts being auto-generated?

jewzaam: Yes, it's the meat of the change. A problem is json representation of boolean causes some problems. The fix is to move away from boolean values, hence update to empty string.

PR #111, opened 2019-10-01
No longer cut CLUSTER_NAME in upgrade script

jewzaam:
This means you have to pass full cluster name as an argument.  But script was missing some clusters.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -297,11 +297,13 @@ do
             OCM_NAME=$(oc -n $CD_NAMESPACE get clusterdeployment $CD_NAME -o json | jq -r '.metadata.labels["api.openshift.com/name"]')
             for CLUSTER_NAME in $CLUSTER_NAMES
             do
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -297,11 +297,13 @@ do
             OCM_NAME=$(oc -n $CD_NAMESPACE get clusterdeployment $CD_NAME -o json | jq -r '.metadata.labels["api.openshift.com/name"]')
             for CLUSTER_NAME in $CLUSTER_NAMES
             do
-              if [ "$OCM_NAME" == "$(echo $CLUSTER_NAME | cut -c -15)" ];
```
jharrington22: Why remove the cut?

jewzaam: value of `OCM_NAME` is the full name, as supplied by the customer.  It's captured in a label `api.openshift.com/name`.  Ideally we move away from this name mess and go to the ID, it's getting to be a mess.  But without reworking a bunch of things (including upgrade process we're developing) I am sticking with the name.

PR #120, opened 2019-10-18
Add project-request template

dak1n1:
This adds the ability for dedicated-admins to create/edit a project-request template and update the master-api configuration for project creation. It effectively lets dedicated-admins enable/disable project-request templates as needed.
https://jira.coreos.com/browse/SREP-1258

https://docs.openshift.com/container-platform/4.1/applications/projects/configuring-project-creation.html#modifying-template-for-new-projects_configuring-project-creation


deploy/osd-project-request-template/00-template.project-request.yaml
```diff
@@ -0,0 +1,90 @@
+apiVersion: template.openshift.io/v1
```
cblecker: Hmm.. Having this object in a SSS with sync mode will clobber any changes the customer makes to it.

Might need to think how we can deploy objects with a different `resourceApplyMode` in the context of MCC

deploy/osd-project-request-template/02-clusterrole.dedicated-admins-project-request.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
cblecker: Should this be a `Role` in the openshift-config namespace instead, as it refers to a single object?

deploy/osd-project-request-template/02-clusterrole.dedicated-admins-project-request.yaml
```diff
@@ -0,0 +1,17 @@
...
+rules:
+- apiGroups:
+  - template.openshift.io
+  resources:
+  - templates
+  resourceNames:
+  - project-request
+  verbs:
+  - get
+  - list
```
cblecker: Is it possible to list on a single object?

deploy/osd-project-request-template/00-template.project-request.yaml
dak1n1: That's a good point. Do you think it would help to put this template in with an operator's deployment manifests instead? Like in rbac-permissions-operator?

In v3, we had customers set an [annotation](https://docs.openshift.com/dedicated/3/admin_guide/managing_projects.html#managing-limits-and-quota-on-project-resources) on the template in order to prevent overwrites. That way they could choose to keep up-to-date with our changes, or go their own way. CC @jewzaam  any thoughts?

deploy/osd-project-request-template/02-clusterrole.dedicated-admins-project-request.yaml
dak1n1: I've only ever used a `list` in operator code... it doesn't seem to be possible from the CLI. I'll remove it since it's probably not necessary.

dak1n1:
Test result for latest commit:

```
[dakini@nibbana ~]$ oc adm policy who-can update project.config.openshift.io
resourceaccessreviewresponse.authorization.openshift.io/<unknown>
Namespace: default
Verb:      update
Resource:  projects.config.openshift.io

[...]

Groups: dedicated-admins
        system:serviceaccounts:dedicated-admin
```

So now if a dedicated-admin wants to use a custom project-request template, they can follow the docs here:

https://docs.openshift.com/container-platform/4.2/applications/projects/configuring-project-creation.html#modifying-template-for-new-projects_configuring-project-creation

dak1n1:
@c-e-brumm I rebased and squashed, which removed the LGTM label. Could you +1 again?

PR #123, opened 2019-10-24
add cloud platform labeling

mjudeikis:
Enables GCP support in SelectorSyncSets:

* Adds individual yaml file labeling for different cloud providers
* Adds logic to propagate labels to SelectorSyncSet template


cblecker:
[test]

mjudeikis:
[test]

scripts/generate_syncset.py
```diff
@@ -51,7 +63,6 @@ def process_yamls(name, directory, obj):
                 o['spec']['resources'] = []
             o['spec']['resources'].append(y)
 
-    o['metadata']['name'] = name
```
cblecker: Why did this get removed?

scripts/generate_syncset.py
```diff
@@ -85,6 +96,7 @@ def process_yamls(name, directory, obj):
             else:
                 # SSS name is based on dirpath which has the root path prefixed.. remove that prefix
                 sss_name = sss_name[(len(arguments.yaml_directory) + 1):]
+            #print(selectorsyncset_data)
```
cblecker: Left behind debug line?

mjudeikis:
> The one issue I see with this design wise that I didn't think of before, is that the way the script works is there is one SSS per folder, not per object.
> 
> For example, the sre-prometheus rules are now all aws-only, because there are a couple that have the platform label, so it sets it for the entire SSS.
> 
> We need to break out aws-specific things into their own folder, and then figure out a way to denote on a folder by folder basis. Maybe a file in it that defines the clusterDeploymentSelector? Maybe a regex on the folder name?

This is why I not the fan of "foldered" approach and was doing first cut on API based folders[1]. People should know where their components are. All this purpose based foldering is creating unessasary level of complexity and potential duplication in different places. 

1. https://github.com/mjudeikis/managed-cluster-config/tree/refactor.to.go/pkg/sync/data

scripts/generate_syncset.py
mjudeikis: bad auto-rebase, sorry

cblecker:
[test]

cblecker:
Closing in favour of #136

PR #127, opened 2019-10-31
SREP-2178 Add dedicated-admins permissions to customer monitoring NS

rogbas:
- Incorporate RBAC permissions for dedicated-admins in `openshift-customer-monitoring` NS
     - Only a pre-selected list of secrets can be managed by dedicated-admins in that NS
- Add 3 new secrets to the list owned by dedicated-admins in the same NS
     - prometheus-additional-scrape-config
     - prometheus-additional-alertmanager-config
     - alertmanager-config

cc @aditya-konarde


deploy/osd-customer-monitoring/05-role.yaml
```diff
@@ -39,4 +42,491 @@ rules:
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  - apps.openshift.io
+  resources:
+  - deploymentconfigrollbacks
```
jharrington22: Is this right?

deploy/osd-customer-monitoring/05-role.yaml
```diff
@@ -39,4 +42,491 @@ rules:
...
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  - project.openshift.io
+  resources:
+  - projects
+  verbs:
+  - get
+- apiGroups:
```
jharrington22: Out of order?

deploy/osd-customer-monitoring/05-role.yaml
```diff
@@ -39,4 +42,491 @@ rules:
   - servicemonitors
   - podmonitors
   verbs:
-  - "*"
\ No newline at end of file
+  - "*"
+- apiGroups:
+  - ""
```
jharrington22: Create secrets in any namespace?

deploy/osd-customer-monitoring/05-role.yaml
```diff
@@ -39,4 +42,491 @@ rules:
...
+  - authorization.openshift.io
+  resources:
+  - resourceaccessreviews
+  - subjectaccessreviews
+  verbs:
+  - create
+- apiGroups:
+  - ""
```
jharrington22: Is this required? There are a few apiGroups like this.

deploy/osd-customer-monitoring/05-role.yaml
rogbas: It's part of the dedicated-admin set of permissions that we support today.

deploy/osd-customer-monitoring/05-role.yaml
rogbas: I will remove lines 59 and 60

deploy/osd-customer-monitoring/05-role.yaml
rogbas: I'm following the same set of permissions we have for dedicated-admins today, in the original Role(s) there is this "" apiGroup as well. I'm assuming it's there for a reason, probably retro compatibility?

jharrington22: I would have thought that `""` gives access to all apiGroups? or is that specifically `"*"`?

cblecker: it's a blank API group. It's not really needed in either case.

The only resources that need it are:
```
$ oc api-resources --api-group=""
NAME                     SHORTNAMES   APIGROUP   NAMESPACED   KIND
bindings                                         true         Binding
componentstatuses        cs                      false        ComponentStatus
configmaps               cm                      true         ConfigMap
endpoints                ep                      true         Endpoints
events                   ev                      true         Event
limitranges              limits                  true         LimitRange
namespaces               ns                      false        Namespace
nodes                    no                      false        Node
persistentvolumeclaims   pvc                     true         PersistentVolumeClaim
persistentvolumes        pv                      false        PersistentVolume
pods                     po                      true         Pod
podtemplates                                     true         PodTemplate
replicationcontrollers   rc                      true         ReplicationController
resourcequotas           quota                   true         ResourceQuota
secrets                                          true         Secret
serviceaccounts          sa                      true         ServiceAccount
services                 svc                     true         Service
```

deploy/osd-customer-monitoring/05-role.yaml
rogbas: removed

deploy/osd-customer-monitoring/05-role.yaml
rogbas: +1, "" is not "*". It's just a set of resources that can have apiGroup = "", examples: configmaps, endpoints, services, secrets.

Thanks for the list @cblecker

rogbas: To clean it up, I removed all the "" apiGroups that already had another apiGroup assigned in the rule

jharrington22: Interesting thanks @cblecker for the explanation

PR #133, opened 2019-11-06
SREP-2359 Changes to support the new v4.2 features to manage OLM overrides

rogbas:
- Updated generate-syncset script, to support generating SSSs with `resourceApplyMode` different from the default `sync` mode
- Added new upsert SSS for the OperatorHub CRD
- Added new SSS (sync mode) for OperatorSources and OperatorHub CR


cblecker:
/lgtm
/approve
/hold

Seems good enough for now. I need to look at all this logic anyways soon for GCP.

PR #136, opened 2019-11-11
add cloud platform labelling

cblecker:


cblecker:
/cc @mjudeikis @rogbas @jewzaam 

This might be a simpler solution for cloud platform labelling than #123, and consistent with @rogbas' use of the upsert directory. Let me know what you think.

mjudeikis:
Im ok any way as long as it delivers the same result.

PR #142, opened 2019-11-25
Script cluster upgrade verification

jharrington22:
This script will verify that upgrades are complete by checking
1. The control plane reports that the upgrade is complete
2. All nodes are in the "Ready" state
3. All nodes are at the same version

### DIFF
--- a/scripts/verify_cluster_upgrades.sh
+++ b/scripts/verify_cluster_upgrades.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+IFS='
+'
+
+for cluster in "$@"
+do
+  echo -e "\n======================================="
+  echo "verifying $cluster"
+  echo "======================================="
+  ocm cluster login "$cluster"%
+  if ! [ $? = 0 ]
+  then
+    echo "error logging into cluster: $cluster skipping.."
+    continue
+  else
+    echo -e "\n$cluster login successful, checking upgrade status\n"
+  fi
+  CONTROL_PLANE_STATUS=$(oc get clusterversion version -o json | jq -r '.status.conditions[] | select(.status == "True") | select(.type != "RetrievedUpdates") | .message')
+
+  echo "======================================="
+  echo "$cluster upgrade status"
+  echo -e "=======================================\n"
+  
+  if echo "$CONTROL_PLANE_STATUS" | grep -q -E "Done applying 4\.[0-9]+\.[0-9]+"
+  then
+    echo "$cluster control plane status: $CONTROL_PLANE_STATUS"
+  echo
+    echo "$cluster control plane not ready yet!"
+    echo "$CONTROL_PLANE_STATUS"
+  fi
+
+  NODE_STATUS=$(oc get nodes --no-headers)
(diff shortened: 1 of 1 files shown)


scripts/verify_cluster_upgrades.sh
```diff
@@ -0,0 +1,57 @@
...
+  then
+    echo "$cluster control plane status: $CONTROL_PLANE_STATUS"
+  echo
+    echo "$cluster control plane not ready yet!"
+    echo "$CONTROL_PLANE_STATUS"
+  fi
+
+  NODE_STATUS=$(oc get nodes --no-headers)
+  NODE_READY_COUNT=$(echo "$NODE_STATUS" | awk '{print $2}' | uniq | wc -l)
+  if [ "$NODE_READY_COUNT" = 1 ]
```
NautiluX: to work on a mac, this needs a numeric comparison (`-eq` instead of `=`)

scripts/verify_cluster_upgrades.sh
```diff
@@ -0,0 +1,57 @@
...
+    then
+      echo "$cluster nodes state: Ready"
+    fi
+  else
+    echo "$cluster nodes are not all ready yet"
+    echo "$NODE_STATE"  
+  fi
+
+  NODE_VERSION_COUNT=$(echo "$NODE_STATUS" | awk '{print $5}' | uniq | wc -l)
+  if [ "$NODE_VERSION_COUNT" = 1 ]
```
NautiluX: to work on a mac, this needs a numeric comparison (`-eq` instead of `=`)

scripts/verify_cluster_upgrades.sh
```diff
@@ -0,0 +1,57 @@
...
+  CONTROL_PLANE_STATUS=$(oc get clusterversion version -o json | jq -r '.status.conditions[] | select(.status == "True") | select(.type != "RetrievedUpdates") | .message')
+
+  echo "======================================="
+  echo "$cluster upgrade status"
+  echo -e "=======================================\n"
+  
+  if echo "$CONTROL_PLANE_STATUS" | grep -q -E "Done applying 4\.[0-9]+\.[0-9]+"
+  then
+    echo "$cluster control plane status: $CONTROL_PLANE_STATUS"
+  echo
```
NautiluX: `else`?

scripts/verify_cluster_upgrades.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
```
NautiluX: I'd propose to add `set -euo pipefail`

jewzaam:
@jharrington22 what is different in this script from what `cluster-upgrade.sh` is doing today?  From the description I do not see why we'd break this out as something different.

1. When CVO is done upgrading and `ClusterVersion` reports the cluster is on the target version it means the control plane is upgrade.
2. I think the upgrade script will make sure all nodes are Ready before completing (after every node is on the same version).  If not, upgrade script should be updated.
3. Upgrade script blocks until all nodes are on the same version.  Implication of a cluster version changing is that the version of master nodes has changed.  The script waits for all nodes to match implying all workers must be on the same version as master.

PR #148, opened 2019-12-02
SREP-2178 - Add perms for dedicated-admins to manage strimzi cluster operator

rogbas:
- Add permissions for dedicated-admins to deploy strimzi operator into the openshift-strimzi namespace
- Add permissions for dedicated-admins to manage the operator from that namespace

SREP-2178


deploy/osd-strimzi-operator/05-role.yaml
```diff
@@ -0,0 +1,76 @@
...
+  - pods/log
+  - events
+  verbs:
+  - list
+  - get
+  - watch
+- apiGroups:
+  - logging.openshift.io
+  resources:
+  - clusterloggings
```
jewzaam: `ClusterLogging` CR should not be allowed in this namespace.

deploy/osd-strimzi-operator/05-role.yaml
```diff
@@ -0,0 +1,76 @@
...
+  resources:
+  - "*"
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
jewzaam: Why not use "*" for the kafka verbs?

deploy/osd-strimzi-operator/05-role.yaml
rogbas: We grant that on codeready workspaces NS, I understood it was there to allow customers to manage logging of those components: https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-codeready-workspaces/05-role.yaml

deploy/osd-strimzi-operator/05-role.yaml
rogbas: Could be, it's the same thing.

rogbas: Just updated verbs to "*"

rogbas:
rebased with latest changes in the repo

deploy/osd-strimzi-operator/05-role.yaml
jewzaam: If we give it to crw we should remove it.  This CR will stand up a whole new logging stack.  Probably came with copying perms for the `openshift-logging` namespace.

rogbas:
@jewzaam I cleaned up logging related permissions from other folders as well

PR #149, opened 2019-12-03
Add dedicated-admin SA to logging rolebinding

wanghaoran1988:


maorfr:
This is required since in OSDv4 as per https://docs.openshift.com/container-platform/4.1/logging/efk-logging.html

wanghaoran1988:
PR rebased, @jewzaam could you lgtm again?

deploy/osd-logging/06-rolebinding.yaml
```diff
@@ -4,8 +4,12 @@ metadata:
   name: dedicated-admins-openshift-logging
   namespace: openshift-logging
 subjects:
-- kind: Group
+- apiGroup: rbac.authorization.k8s.io
```
cblecker: This is incorrect. Groups are not a part of the `rbac.authorization.k8s.io` apiGroup.

jewzaam: Huh, yea, thanks.  It's in `user.openshift.io`.  We have used it incorrectly in other places but it is functioning.  If we could confirm the proper apiGroup then we should circle back (separate jira/pr/etc) and fix everything else.

wanghaoran1988: @cblecker @jewzaam 
This should be correct according to the api definition here: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/rbac/types.go#L73
This is a admin rolebinding automatically created when you create a project, it's using the "rbac.authorization.k8s.io" group:


```
kind: RoleBinding
metadata:
  creationTimestamp: "2019-12-06T05:57:24Z"
  name: admin
  namespace: haowang
  resourceVersion: "536215"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/haowang/rolebindings/admin
  uid: 465c8df1-17ed-11ea-a7e9-02a23b63208b
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: haowang@redhat.com
```

jewzaam: If you review the actual resource definitions in OpenShift v4 clusters this isn't the `apiGroup` assigned to `Groups`.  I guess the cluster is handling it on the back-end.

This is a script I use to get a dump of everything available in a cluster: https://gist.github.com/jewzaam/b51c0ec1c0de896cdd0ba0434bfebb5d

wanghaoran1988: @jewzaam I know what you mean, the 'Group' resources is openshift specific, my understaning here, the Group here is diffrent with the one in openshift, when openshift finish authentication, it will bring the user/group attributes to k8s apiserver, and then  the rbac authorizer will then using the "groups" info of the current user do the authorization. 

Here for the Group kind, the apiGroup must be "rbac.authorization.k8s.io", or you cannot create the clusterrolebinding, it will fail with errors:

```
The ClusterRoleBinding "test-admin-1" is invalid: subjects[0].apiGroup: Unsupported value: "user.openshift.io": supported values: "rbac.authorization.k8s.io"
```
FYI, the validation code is here: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/rbac/validation/validation.go#L240

jewzaam:
/approve
/lgtm

@cblecker could we remove the hold?

jewzaam:
@wanghaoran1988 there are conflicts now, please update thanks!

PR #158, opened 2019-12-06
to support the removal of dedicated-admin and use of rbac-operator

dofinn:
Alertmanager throws an alert on the no longer present DedicatedAdminOperatorOfflineSRE. This removes the alert. We should also look at plugging in relative alerts for the rbac-operator.


jewzaam:
Should be:
```
rm deploy/sre-prometheus/100-dedicated-admin-operator.PrometheusRule.yaml
make
```
then commit the changes

jewzaam:
replace w/ https://github.com/openshift/managed-cluster-config/pull/159

PR #160, opened 2019-12-09
SREP-2178 - Give App-SRE permissions to do pod/poft-forward in customer monitoring NS

rogbas:
Adding pods/port-forwarding perms, so App-sre/customers can troubleshoot issues with the monitoring stack


rogbas:
cc @aditya-konarde

jewzaam:
/hold
Not sure yet why we need this since I expect prometheus to expose services via routes.  I haven't confirmed this with the customer monitoring setup we have in OSD v4, need instructions on how to set it up so I can poke around w/ it.  In-cluster prometheus has routes for AM, grafana, and prom already.  Customer could create routes in the NS to expose the services too.

jewzaam:
Been talking about this.  We would have granted access to do this if the operator were installed in a customer namespace as a deployment and given the ability to setup the route in this namespace there's no reason we should limit the port forwarding in `openshift-customer-monitoring`.

/approved
/lgtm
/hold cancel

PR #162, opened 2019-12-10
SREP-2140 - alert when sre stays in cluster-admin more than 2 hours

boranx:
Fixes: SREP-2140

Currently, SRE can temporarily elevate permissions to cluster-admin by adding themselves to the 'osd-sre-cluster-admins' group.
The alert will be fired if SRE stays in cluster-admin role for more than 2 hours.


jewzaam:
Note this won't catch if I added myself directly to the cluster-admin ClusterRole:
```bash
oc adm groups add-users osd-sre-cluster-admins $(oc whoami)
oc adm policy add-cluster-role-to-user cluster-admin $(oc whoami)
```

But it's better than nothing and will let us know if hive didn't sync the group to remove permissions in the expected 2 hours.

/lgtm
/approve

boranx:
@jewzaam how can we catch if someone adds himself directly into the cluster-admin?
We can also add alert for the situation you mentioned.
I assume currently we can't do it via 'openshift_group_user_account' nor `openshift_group_created`

cblecker:
You'd have to look for new objects being created. `add-cluster-role-to-user` creates a new ClusterRoleBinding object with the role and user you specify

PR #167, opened 2019-12-11
SREP-2532: Ignore Prometheus specific memory usage

cblecker:
Memory usage is linear with the number of metrics being monitored, and is not actionable.. This ignores the specific prometheus pods.


RiRa12621:
Still not a fan of the hard threshold but this should at least reduce the noise.

/lgtm
/approve

jewzaam:
agree, it's a good step

PR #169, opened 2019-12-12
Add dedicated-admin serviceaccounts permissions

NautiluX:
ServiceAccounts in dedicated-admin project get the same
permissions as the dedicated-admins group.


wanghaoran1988:
+1 for a quick fix of this, we should consider allow multiple subjects in the api.

deploy/rbac-permissions-operator-config/00-dedicated-admins-serviceaccounts.SubjectPermission.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: dedicated-admins
```
wanghaoran1988: The name here should be changed

PR #171, opened 2019-12-14
Remove managed-velero-operator

dak1n1:
Managed-velero-operator is now deployed using the app-sre pipeline.
openshift/managed-velero-operator#20


deploy/managed-velero-operator/100-velero.Namespace.yaml
```diff
@@ -1,6 +0,0 @@
-apiVersion: v1
-kind: Namespace
```
cblecker: I think deleting this will delete the NS and the Velero CRs

dak1n1: I tested in hive-stage, and it appears to keep the namespace and CR around when the SelectorSyncSet is deleted.

cblecker:
This was finally done in #393 (about 7 months longer than we expected, but I'm glad we got there).

We miss you @dak1n1! Hope you're well :)

PR #175, opened 2019-12-20
Allow dedicated admins to view router metrics

aditya-konarde:


cblecker:
Is there a reason that dedicated-admins need this directly? Normally this is only needed by prometheus to scrape those metrics.

aditya-konarde:
For the sake of people reading this later: 
- Dedicated admins already have node level status access.
- Router metrics permissions are needed so that dedicated admins can delegate these to a tool like Prometheus. Without this change, it was not possible as Kubernetes would deny it as a privilege escalation.

cblecker:
/lgtm
/approve

That seems reasonable to me.

PR #178, opened 2020-01-08
Ensure files are processed in alphabetic order

NautiluX:


jewzaam:
Verified by a couple of people.  Confirmed still updates the template if `deploy/` changes.  I'm good to merge, would like someone else to give a lgtm but I'll approve.
/approve

NautiluX:
to whoever wants to test this out quickly:
```
git clone https://github.com/NautiluX/managed-cluster-config ntlx-config; cd ntlx-config; git checkout order-files; make; git diff
```
`git diff` should give an empty result.

PR #182, opened 2020-01-13
allow dedicated-admin to create project-request template in openshift-config

maorfr:
according to the attached docs, we need to create a project-request template in openshift-config.
since the template does not exist, `create` needs to be added.

https://docs.openshift.com/container-platform/4.2/applications/projects/configuring-project-creation.html


RiRa12621:
Can you explain why we will need that?

PR #184, opened 2020-01-13
adding label to openshift-customer-monitoring

RiRa12621:
To allow for network polices to be used from and to openshift-customer-monitoring, we have to have a label identifying that namespace.
I propose to be straight forward and use `name`

### DIFF
--- a/deploy/osd-customer-monitoring/00-namespace.yaml
+++ b/deploy/osd-customer-monitoring/00-namespace.yaml
@@ -5,5 +5,6 @@ metadata:
   annotations:
     openshift.io/node-selector: ""
   labels:
+    name: "openshift-customer-monitoring"
     openshift.io/cluster-logging: "true"
     openshift.io/cluster-monitoring: "false"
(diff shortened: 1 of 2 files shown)


RiRa12621:
Sure, do you have anything in particular in mind?

maorfr:
fwiw, the following namespaces have the `name` label:

openshift-cluster-version
openshift-ingress
openshift-insights
openshift-kni-infra
openshift-machine-api
openshift-machine-config-operator
openshift-monitoring
openshift-multus
openshift-network-operator
openshift-openstack-infra
openshift-sdn

rogbas:
I agree it's confusing, but the **name** label is used by the networkPolicy label selector, and several other NSs have to use it too.

IMO, updating that across the board should be covered in a different card.  

/lgtm

cblecker:
> I agree it's confusing, but the **name** label is used by the networkPolicy label selector, and several other NSs have to use it too.

Ah, that makes sense. Yes, I'm okay with this then too 👍

RiRa12621:
great can someone /lgtm please ?

PR #198, opened 2020-01-30
Rename alertmanager configuration secret name

aditya-konarde:
https://issues.redhat.com/browse/APPSRE-1468


jewzaam:
If this used to work with `alertmanager-config` may want to add `alertmanager-instance` instead of replace the old value.

PR #207, opened 2020-02-05
Remove nodeSelectors on prom/am

cblecker:
This removes the nodeSelector on prometheus and alertmanager


jewzaam:
we are going to move to 3 infra which will make this go away
/close

PR #209, opened 2020-02-06
OSD-2809 - Enable missing Secrets permissions in openshift-strimzi NS

rogbas:
Per discussion with PM/BU, enabling Secrets permissions for dedicated-admins in openshift-strimzi NS


jewzaam:
/lgtm
confirmed we have approval to provide this

PR #211, opened 2020-02-10
Allow any "non-default", 'enabled' CIS to be a valid target

lisa:
This change allows any 'enabled' ClusterImageSet to be a valid target,
even if it's not the "default". We still check to see if we're doing an
upgrade, however.


lisa:
/cancel hold

Green light for merge.

lisa:
/hold cancel

helps if I use the right magic words.

PR #214, opened 2020-02-13
Script to rotate aws creds

jewzaam:
https://issues.redhat.com/browse/OSD-2873

### DIFF
--- a/scripts/rotate-aws-creds.sh
+++ b/scripts/rotate-aws-creds.sh
@@ -1,0 +1,214 @@
+#!/bin/bash
+
+ENVIRONMENT=$1
+CLUSTER_ID=$2
+
+# make sure we have no state for the aws cli
+unset AWS_PROFILE
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION=us-east-1
+
+if [ "$CLUSTER_ID" == "" ] || [ "$ENVIRONMENT" == "" ];
+then
+    echo "usage: $0 <environment> <cluster id>"
+    echo "example: $0 stage 18kshg9idnn687p1bstlh669hlod7ldr"
+    exit 1
+fi
+
+info() {
+    echo "INFO: $1"
+}
+
+aws_cli_setup() {
+    KEY_ID=$1
+    SECRET_KEY=$2
+    PROFILE_NAME=$3
+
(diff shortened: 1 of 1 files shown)


scripts/rotate-aws-creds.sh
```diff
@@ -0,0 +1,214 @@
+#!/bin/bash
+
+ENVIRONMENT=$1
+CLUSTER_ID=$2
+
+# make sure we have no state for the aws cli
+unset AWS_PROFILE
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION=us-east-1
```
RiRa12621: should have a comment like 
// This is a mandatory value to be set.
// It only has to be any valid region, not reflecting the region the cluster is in as the IAM interaction is globaal

scripts/rotate-aws-creds.sh
```diff
@@ -0,0 +1,216 @@
...
+CLUSTER_ID=$2
+
+# make sure we have no state for the aws cli
+unset AWS_PROFILE
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
+# Region is a mandatory value to be set when using AWS CLI.
+# All AWS CLI requests are related to IAM and are global, so picking any region will be OK.
+export AWS_DEFAULT_REGION=us-east-1
+
```
RiRa12621: if [ $1" == "-h" ] || [ "$1" == "--help" ]
then
    echo "usage: $0 <environment> <cluster id>"
    echo "example: $0 stage 18kshg9idnn687p1bstlh669hlod7ldr"
    exit 1
fi

scripts/rotate-aws-creds.sh
```diff
@@ -0,0 +1,216 @@
...
+if [ "$CLUSTER_ID" == "" ] || [ "$ENVIRONMENT" == "" ];
```
RiRa12621: display usage also if `-h` or `--help` flag is given

```
if [ $1" == "-h" ] || [ "$1" == "--help" ] || [ "$CLUSTER_ID" == "" ] || [ "$ENVIRONMENT" == "" ];
then
    echo "usage: $0 <environment> <cluster id>"
    echo "example: $0 stage 18kshg9idnn687p1bstlh669hlod7ldr"
    exit 1
fi
```

PR #221, opened 2020-02-17
added back the ingress pod landing on infra nodes

mwoodson:


jewzaam:
/hold
are we trying to test in non-prod prior to the hive update in hive?  if not, let's wait for prod deployment of hive update so we don't have issue of this change holding up MCC promotion

mwoodson:
My intention is to get this onto int/stg to verify this is working before we promote hive to prod.

PR #223, opened 2020-02-18
Added secret names to managed by openshift-acme

rporres:
openshift-acme uses secrets to keep tls info related to routes. We need to define which secrets it's going to manage in this namespace as dedicated admin only has rights to manage certain named resources here

This is a follow-up of #197 

cc @maorfr

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>


rporres:
cc @rogbas

PR #224, opened 2020-02-18
Add LP SRE Cluster admin and RHMI SRE RBAC

alanmoran:
As per https://issues.redhat.com/browse/OSD-2783

- Create new layered-sre-cluster-admins (breakglass)
- Create CS SRE admins group (RHMI)
- New monitoring check for  layered-sre-cluster-admins group

### DIFF
--- a/deploy/layered-sre-authorization/00-layered-sre-cluster-admins.Group.yaml
+++ b/deploy/layered-sre-authorization/00-layered-sre-cluster-admins.Group.yaml
@@ -1,0 +1,6 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: layered-sre-cluster-admins
+users: []
+# NOTE when created as a SelectorSyncSet, Hive will wipe out the list of users when cache expires because `users: []` and this is desired!
--- a/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
+++ b/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
@@ -1,0 +1,45 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: layered-cs-sre-admin-cluster
+rules:
+# CS SRE can manage projects
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - projects
+  verbs:
+  - '*'
+- apiGroups:
+  - project.openshift.io
+  resources:
+  - projects
+  verbs:
+  - '*'
+# CS SRE can manage namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  - namespaces/finalize
+  verbs:
+  - '*'
+# CS SRE can update existing groups (enables break glass)
+- apiGroups:
+  - user.openshift.io
+  resources:
+  - groups
+  verbs:
+  - patch
+  - update
+# CS SRE can interact with RHMI ClusterResourceQuotas (Only for LA)
+- apiGroups:
+  - quota.openshift.io
+  resources:
+  - clusterresourcequotas/status
+  - clusterresourcequotas
+  verbs:
+  - '*'
+  resourceNames:
+  - rhmi-loadbalancer-quota
+  - rhmi-persistent-volume-quota
--- a/deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
+++ b/deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
@@ -1,0 +1,16 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: layered-cs-sre-admin
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+  - dedicated-admins-cluster
+  - layered-cs-sre-admin-cluster
+  permissions:
+  - allowFirst: true
+    clusterRoleName: layered-cs-sre-admin-project
+    namespacesAllowedRegex: .*
+    namespacesDeniedRegex: (^openshift-.*|^kube-.*|^ops-health-monitoring$|^management-infra$|^default$|^logging$|^sre-app-check$)
+  subjectKind: Group
+  subjectName: layered-cs-sre-admins  
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -24,6 +24,22 @@ parameters:
   required: true
 - name: TELEMETER_SERVER_URL
   required: true
+- name: RHMI_SRE_USER1
+  required: false
+- name: RHMI_SRE_USER2
+  required: false
+- name: RHMI_SRE_USER3
+  required: false
+- name: RHMI_SRE_USER4
+  required: false
+- name: RHMI_SRE_USER5
+  required: false
+- name: RHMI_SRE_USER6
+  required: false
+- name: RHMI_SRE_USER7
+  required: false
+- name: RHMI_SRE_USER8
+  required: false
 metadata:
   name: selectorsyncset-template
 objects:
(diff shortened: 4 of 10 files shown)


deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: SubjectPermission
+metadata:
+  name: layered-cs-sre-admin
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+  - layered-cs-sre-admin
+  permissions:
+  - allowFirst: true
+    clusterRoleName: dedicated-admins-cluster
```
jewzaam:
```suggestion
    clusterRoleName: layered-cs-sre-admin
```
Suspect this is a copy/paste error, intent is to authorize with the LP admin ClusterRole

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
```diff
@@ -0,0 +1,15 @@
...
+    namespacesAllowedRegex: .*
```
jewzaam:
```suggestion
    namespacesAllowedRegex: ^redhat-.*
```

Do you really need all customer namespaces?  You can limit this to the LP namespaces if not.

deploy/layered-sre-authorization/00-layered-sre-cluster-admins.Group.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: layered-sre-cluster-admins
```
cblecker: Should we prefix this with osd, like our other groups?

alanmoran: Went the layered-* route so we don't confuse with SRE P groups. layered-* will be added to the webhook validation.

jewzaam:
Added support for `sss-config.yaml` in https://github.com/openshift/managed-cluster-config/pull/231
You can create this file in the directory with your resources and supply the RHMI label.
```
matchLabels:
    api.openshift.com/addon-rhmi-operator: "true"
```

alanmoran:
> Added support for `sss-config.yaml` in #231
> You can create this file in the directory with your resources and supply the RHMI label.
> 
> ```
> matchLabels:
>     api.openshift.com/addon-rhmi-operator: "true"
> ```

Thanks @jewzaam! I've applied the SSS config now. The build was also failing as I had added the RHMI_SRE_USER1 template to the selectorsyncset template directly. I've updated template.yaml to allow for these optional values.

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: layered-cs-sre-admin
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+  - dedicated-admins-cluster
```
jewzaam:
```suggestion
  - layered-cs-sre-admin
```

This should not be the `dedicated-admins-cluster` ClusterRole

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - project.openshift.io
+  resources:
+  - projects
+  verbs:
+  - '*'
+# CS SRE can manage namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
```
jewzaam: Same issue as `projects` resource.

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
```diff
@@ -0,0 +1,87 @@
...
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - projects
+  verbs:
+  - '*'
+- apiGroups:
+  - project.openshift.io
+  resources:
+  - projects
```
jewzaam: I believe this must be granted at a cluster scope.  If so, need to create a separate ClusterRole and a ClusterRoleBinding to the `layered-cs-sre-admins` group.

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - '*'
+# CS SRE can manage namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  - namespaces/finalize
+  verbs:
+  - '*'
+# CS SRE can use oc debug
```
jewzaam: Using `oc debug` requires being run in the `default` namespace.  Need to split this out to a separate permissions.
See https://github.com/openshift/ops-sop/blob/b40edcb9c761d7023b7a8ced30cc9c7aea05cad1/v4/support.asciidoc#23-access-node-via-oc

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
alanmoran: I'm trying to achieve giving the group layered-cs-sre-admin the dedicated-admins-cluster role available clusterwide, and the layered-cs-sre-admin role available in all the non redhat namespaces. Am I applying this wrong?

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
alanmoran: I don't really understand what needs to be done here. This is a ClusterRole?

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
alanmoran: The intention for this is to run debug containers in RHMI namespaces and not against nodes. Not sure if I still need to split out. If I do, what permissions do I need to split this to? CS SRE have access to the default NS.

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
jewzaam: You could add layered-cs-sre-admin to this, it's an array.  Why does CS SRE need dedicated-admin permissions in cluster?  I expect your `layered-cs-sre-admin` ClusterRole to have everything CS SRE needs in a customer namespace and that those permissions are something less than dedicated-admin.

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
jewzaam: A ClusterRole isn't applied until you bind it to something.  To grant project and namespace I think you have to use a ClusterRoleBinding.  Using SubjectPermission CR will create RoleBindings in the matching namespaces.  You need to create a ClusterRoleBinding as a part of this PR to get the cluster scope permissions.

For example, dedicated-admin permissions are split into two ClusterRoles: dedicated-admins-cluster [1] and dedicated-admins-project [2].  A ClusterRoleBinding is created for the dedicated-admins-cluster permissions [3], a SubjectPermission is created for the dedicated-admins-project permissions [4] which results in RoleBindings in each customer namespace.

[1] https://github.com/openshift/managed-cluster-config/blob/master/deploy/rbac-permissions-operator-config/00-dedicated-admins-cluster.ClusterRole.yaml
[2] https://github.com/openshift/managed-cluster-config/blob/master/deploy/rbac-permissions-operator-config/00-dedicated-admins-project.ClusterRole.yaml
[3] https://github.com/openshift/managed-cluster-config/blob/master/deploy/rbac-permissions-operator-config/00-dedicated-admins-cluster.ClusterRoleBinding.yaml
[4]

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
jewzaam: Good point, I think "node" when I see `oc debug`.  Appears to work on a pod, so yes this is fine as-is.

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
alanmoran: Ok, Interesting. My view is that we have the same as the customer and a subset more. This is purely from the nature of the type of work we do. We are managing middleware products with customers more than on behalf. An example here is products dynamically creating namespaces that hold config. Customers can manage this but so can CS SRE.

deploy/layered-sre-authorization/01-layered-cs-sre-admin.ClusterRole.yaml
alanmoran: I thought the RBAC operator created the ClusterRoleBinding too. I'll validate this tomorrow to check. I made a change to this piece earlier in the PR and now need to retest it.

alanmoran: Ok, I've tested this. The ClusterRoleBinding is created (if missing). I did lose the layered-cs-sre-admin ClusterRoleBinding though. I've re-added it to the SubjectPermission. Nice catch. Thank you.

deploy/layered-sre-authorization/02-layered-cs-sre-admin.SubjectPermission.yaml
jewzaam: Ok, I see my disconnect.  You're ClusterRole reads to me like a project level permissions.  I missed the clusterPermissions will create the ClusterRoleBinding.  Your layered-cs-sre-admin ClusterRole is too broad to grant as a ClusterRoleBinding, please split it into two ClusterRoles, one for cluster scope resources (project, namespace, user, etc) and project resources (everything else).  In the SubjectPermissions the cluster scope ClusterRole goes into `clusterPermissions`, the project scope ClusterRole goes into `permissions`.

alanmoran: I also see what you were originally asking.  I've splt the cluster scope and project resources out (layered-cs-sre-admin-cluster and layered-cs-sre-admin-project)

deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
```diff
@@ -0,0 +1,54 @@
...
+rules:
+# CS SRE can get, list, and watch all resources cluster wide
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
```
jewzaam: Why cluster-scope get/list/watch?  My concern with this is it enables access to all secrets in the cluster and I'd just like to check that this is actually needed.

jewzaam: Discussed on slack.  Basically it's there to enable access to the aws creds in case it's needed.  Discussed with @alanmoran about this being a break-glass thing, CS SRE could elevate privileges and get access.  Making elevating required means some thought has to go into the actions that are highly privileged.

jewzaam:
```suggestion
```

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -535,6 +551,159 @@ objects:
...
+        name: layered-cs-sre-admin-cluster
+      rules:
+      - apiGroups:
+        - '*'
+        resources:
+        - '*'
+        verbs:
+        - get
+        - list
+        - watch
```
jewzaam:
```suggestion
```

PR #226, opened 2020-02-19
Fix osd customer monitoring acme-related secrets

rporres:
There was a duplicate there.

Fixes #223

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>


rporres:
cc @maorfr @aditya-konarde @jewzaam

PR #227, opened 2020-02-19
Many changes to the cluster-upgrade script, added ClusterUpgradingSRE alert

jewzaam:
Added new ClusterUpgradingSRE alert
Split the control plane and worker PD maint windows into two things
Get current version from the cluster, not hive
Sign pending CSR while waiting for Nodes to provision
Don't skip if kubelet not updated yet
Log if waiting for worker to provision
Log ClusterOperator summary status periodically
Log Node upgrade summary status periodically
Exclude ClusterUpgradingSRE (new) and DNSErrors05MinSRE alerts from upgrade script checks
Tweaks on sleeps, timing, and fix to desired version check

### DIFF
--- a/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
@@ -1,0 +1,20 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-cluster-upgrade-alerts
+    role: alert-rules
+  name: sre-cluster-upgrade-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-cluster-upgrade-alerts
+    rules:
+    - alert: ClusterUpgradingSRE
+      expr: cluster_version{type="updating"} > 0
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+      annotations:
+        from_version: "$labels.from_version"
+        to_version: "$labels.version"
--- a/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
@@ -13,7 +13,7 @@ spec:
     - alert: ClusterUpgradingSRE
       expr: cluster_version{type="updating"} > 0
       labels:
-        severity: critical
+        severity: warning
         namespace: openshift-monitoring
       annotations:
         from_version: "$labels.from_version"
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -4273,7 +4273,7 @@ objects:
           - alert: ClusterUpgradingSRE
             expr: cluster_version{type="updating"} > 0
             labels:
-              severity: critical
+              severity: warning
               namespace: openshift-monitoring
             annotations:
               from_version: $labels.from_version
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -4262,6 +4262,26 @@ objects:
       kind: PrometheusRule
       metadata:
         labels:
+          prometheus: sre-cluster-upgrade-alerts
+          role: alert-rules
+        name: sre-cluster-upgrade-alerts
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-cluster-upgrade-alerts
+          rules:
+          - alert: ClusterUpgradingSRE
+            expr: cluster_version{type="updating"} > 0
+            labels:
+              severity: critical
+              namespace: openshift-monitoring
+            annotations:
+              from_version: $labels.from_version
+              to_version: $labels.version
+    - apiVersion: monitoring.coreos.com/v1
+      kind: PrometheusRule
+      metadata:
+        labels:
           prometheus: sre-configure-alertmanager-operator-offline-alerts
           role: alert-rules
         name: sre-configure-alertmanager-operator-offline-alerts
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -419,7 +419,6 @@ check_cluster_status() {
                 fi
             done
 
-            CLUSTER_STATUS=0
         fi
     fi
 }
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -4,16 +4,18 @@
 # make sure if the script is killed all child processes are killed:
 trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
 
-PD_MAINT_DESCRIPTION="OSD v4 Cluster Upgrade"
+PD_MAINT_DESCRIPTION_BASE="OSD v4 Cluster Upgrade Control Plane"
+PD_MAINT_DESCRIPTION_WORKERS="OSD v4 Cluster Upgrade Nodes"
 
-# base single-az cluster takes about 60 minutes to upgrade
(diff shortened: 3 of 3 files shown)


lisa:
@jewzaam What is the new alert for? Do we need to page people when there's an upgrade in progress?

scripts/cluster-upgrade.sh
```diff
@@ -118,33 +120,37 @@ setup() {
...
+    # Set the initial maintenance window to the base so control plane issues will page earlier for very large clusters.
+    setup_maintenance_window $OCM_NAME $CD_NAMESPACE $CD_NAME $PD_MAINT_BASE_MIN "$PD_MAINT_DESCRIPTION_BASE"
 }
 
 setup_maintenance_window() {
     OCM_NAME=$1
     CD_NAMESPACE=$2
     CD_NAME=$3
     MAINT_WINDOW_MIN=$4
+    MAINT_DESC=$5
```
lisa: Do you need to quote this assignment so that multiple words don't overlap into `$6`, `$7`, etc?

scripts/cluster-upgrade.sh
```diff
@@ -226,24 +232,25 @@ upgrade() {
     
     log $OCM_NAME "upgrade" "Checking $OCM_NAME..."
 
-    # - do we need to upgrade?
-    OCP_CURRENT_VERSION=$(oc -n $CD_NAMESPACE get clusterdeployment $CD_NAME -o json | jq -r '.status.clusterVersionStatus.history[] | select(.state == "Completed") | .version' | head -n1)
+    # do we need to upgrade?
+    OCP_CURRENT_VERSION=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusterversion version -o json | jq -r '.status.history[] | select(.state == "Completed") | .version' | grep $FROM)
+    KUBELET_VERSION_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers -o custom-columns=VERSION:.status.nodeInfo.kubeletVersion | sort | uniq | wc -l)
```
```suggestion
    KUBELET_VERSION_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers -o custom-columns=VERSION:.status.nodeInfo.kubeletVersion | sort -u | wc -l)
```

What's the count of the unique kubelet versions do for us?

Also, `sort -u` is one fewer pipe for `sort | uniq`

scripts/cluster-upgrade.sh
```diff
@@ -255,45 +262,90 @@ upgrade() {
...
+        sleep 60
+        OCP_CURRENT_VERSION=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusterversion version -o json | jq -r '.status.history[] | select(.verified == true and .state == "Completed") | .version' | grep $TO)
     done
 
     log $OCM_NAME "upgrade" "ClusterVersion on $TO"
-    log $OCM_NAME "upgrade" "checking kubelet versions"
+    log $OCM_NAME "upgrade" "checking node versions"
 
     # fun fact!  after clusterversion says it is done, individual nodes could still be updated.
+    MASTER_KUBELET_VERSION=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers | grep master | sort | uniq | awk '{print $5}' | sort | uniq)
```
```suggestion
    MASTER_KUBELET_VERSION=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers | grep master | sort | uniq | awk '{print $5}' | sort -u)
```

As above

jewzaam:
@lisa the cluster is in maint for the control plane upgrade for currently 90 minutes.  The alert will only page if we breach that window.  Expectation is no human is actively watching script output and the alert provides context for what is happening to the cluster, there will be other alerts too.

scripts/cluster-upgrade.sh
```diff
@@ -255,45 +262,90 @@ upgrade() {
...
+        teardown_maintenance_window $OCM_NAME $CD_NAMESPACE $CD_NAME "$PD_MAINT_DESCRIPTION_BASE"
+    fi
+
+    # log current state
+    log $OCM_NAME "upgrade" "nodes upgraded: $MACHINE_COUNT_UPGRADED/$MACHINE_COUNT_ALL"
+
     # make sure all nodes run same kubelet version
-    KUBELET_VERSION_COUNT=0
-    while [ "$KUBELET_VERSION_COUNT" != "1" ];
+    while [ ! $MACHINE_COUNT_UPGRADED -ge $MACHINE_COUNT_ALL ];
```
```suggestion
    while [ $MACHINE_COUNT_UPGRADED -le $MACHINE_COUNT_ALL ];
```
Not greater than is less than or equal, which reads better?

scripts/cluster-upgrade.sh
```diff
@@ -342,32 +393,34 @@ check_cluster_status() {
...
+            do
+                CO_MESSAGE=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusteroperator $CO -o json | jq -r '.status.extension.lastSyncError')
+
+                if [ "$CO_MESSAGE" != "" ];
+                then
+                    log $OCM_NAME $LOG_NAME "INFO: Operator $CO lastSyncError = $CO_MESSAGE"
+                fi
+            done
+
+            CLUSTER_STATUS=0
```
lisa: The comment being removed says not to set `CLUSTER_STATUS=0` (bad) as it could be a false positive. Do we still have that concern? Is it okay to set that here?

scripts/cluster-upgrade.sh
```diff
@@ -410,6 +467,7 @@ teardown_maintenance_window() {
     OCM_NAME=$1
     CD_NAMESPACE=$2
     CD_NAME=$3
+    MAINT_DESC=$4
```
lisa: Same as above. Does this need to be quoted to prevent extra words being set to `$5`, `$6`, etc?

scripts/cluster-upgrade.sh
jewzaam: When passed to this function, yes.  Hit that in early iterations, all args to calling setup and teardown maint window functions are quoted.

scripts/cluster-upgrade.sh
jewzaam: If we have more than one version on our nodes the worker upgrades are not done.  Thanks for the suggestion.. `sort | uniq` is muscle memory I should change :(

scripts/cluster-upgrade.sh
lisa: Ooh, I just noticed you had greater than or equals. My suggestion should be `lt`

scripts/cluster-upgrade.sh
jewzaam: This can be removed, it's already set after we find a critical alert (line 403)
```suggestion

```

deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: sre-cluster-upgrade-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-cluster-upgrade-alerts
+    rules:
+    - alert: ClusterUpgradingSRE
+      expr: cluster_version{type="updating"} > 0
+      labels:
+        severity: critical
```
lisa:
```suggestion
        severity: warning
```

Per Slack discussion, to keep context for it, but not to cause unnecessary metal association with a "critical"

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -4258,6 +4258,26 @@ objects:
...
+        name: sre-cluster-upgrade-alerts
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-cluster-upgrade-alerts
+          rules:
+          - alert: ClusterUpgradingSRE
+            expr: cluster_version{type="updating"} > 0
+            labels:
+              severity: critical
```
lisa:
```suggestion
              severity: warning
```

matches above

PR #228, opened 2020-02-21
Better validation for node upgrades.

jewzaam:
Check the status of node, not just version.
Check all replicasets are ready & available.
Check all daemonsets are ready & available.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -350,25 +350,15 @@ upgrade() {
     log $OCM_NAME "upgrade" "checking availability of ReplicaSets and DaemonSets"
 
     # verify all replicasets are at expected counts
-    TOTAL_DESIRED=0
-    TOTAL_SCHEDULED=0
-    while [ $TOTAL_DESIRED -eq 0 ] || [ $TOTAL_DESIRED -ne $TOTAL_SCHEDULED ];
+    while true;
     do
-        # don't sleep on first pass
-        if [ $TOTAL_DESIRED -gt 0 ];
-        then
-            sleep 15
-        fi
-
-        # reset counters
         TOTAL_DESIRED=0
-        TOTAL_SCHEDULED=0
-
         for DESIRED in $(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get replicasets --all-namespaces -o json | jq -r '.items[] | select(.metadata.namespace | startswith("default") or startswith("kube") or startswith("openshift")) | select(.status.replicas > 0) | .status.replicas');
         do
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -349,6 +347,68 @@ upgrade() {
...
+    # verify all replicasets are at expected counts
+    TOTAL_DESIRED=0
+    TOTAL_SCHEDULED=0
+    while [ $TOTAL_DESIRED -eq 0 ] || [ $TOTAL_DESIRED -ne $TOTAL_SCHEDULED ];
+    do
+        # don't sleep on first pass
+        if [ $TOTAL_DESIRED -gt 0 ];
+        then
+            sleep 15
+        fi
```
lisa: Maybe on lines 353 the initial values can be initialized (cf lines 367-370, 372-375) to avoid the special case here?

jewzaam: I could but it duplicates the whole loop.  I've pushed an update that doesn't duplicate things, moves the check into the loop, if/else.  Testing in stage to verify it's good.

jewzaam: The update works, keeps it from progressing until all RS and DS are ready.

scripts/cluster-upgrade.sh
```diff
@@ -349,6 +347,62 @@ upgrade() {
         log $OCM_NAME "upgrade" "deleted openshift-machine-api/deployment/cluster-autoscaler-operator"
     fi
 
+    log $OCM_NAME "upgrade" "checking availability of ReplicaSets and DaemonSets"
+
+    # verify all replicasets are at expected counts
+    while true;
+    do
+        TOTAL_DESIRED=0
+        for DESIRED in $(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get replicasets --all-namespaces -o json | jq -r '.items[] | select(.metadata.namespace | startswith("default") or startswith("kube") or startswith("openshift")) | select(.status.replicas > 0) | .status.replicas');
```
lisa: (Same question for line 362)

For OSD, do we permit customers to create a Project (`Namespace`) with a name that starts with `default`? If so, this is a minor-ish nit: we probably want to check for equality to `default`, and not if the Namespace starts with it.

jewzaam: Was trying to keep this from becoming more complicated by using startswith.  Technically yes a customer can create a namespace that starts with "default".  The impact would be as written the script waits for the RS and DS in those namespaces to be ready.  To get that in here I think we'd have to break out the conditions from jq, I didn't find an easy way to do the startswith OR equals.

PR #231, opened 2020-02-25
Add sss-config.yaml to support some overrides for SSS generation

jewzaam:
Related to, but not a solution for, https://issues.redhat.com/browse/OSD-2514
This change allows more nuanced and controlled configuration for SSS generation.
It does NOT support all use cases we might need, such as `matchExpressions` currently hard coded in the template.


jewzaam:
NOTE one SSS did move in the generated template because the parent `UPSERT` directory was removed (now set in configuration)

scripts/generate_syncset.py
```diff
@@ -21,7 +22,8 @@ def get_all_yaml_files(path):
     file_paths = []
     for r,d,f in os.walk(path):
         for file in f:
-            if file.endswith('.yml') or file.endswith('.yaml'):
+            print file
```
jharrington22: Left over debug?

scripts/generate_syncset.py
```diff
@@ -92,15 +94,17 @@ def process_yamls(name, directory, obj, platform="", sss_mode="sync"):
...
-            platform = "aws"
+        # load sss_config if it exists
+        sss_config = {}
+        path_sss_config = os.path.join(dirpath, sss_config_filename)
+        if os.path.exists(path_sss_config):
+            sss_config = get_yaml(path_sss_config)
+
+        # initialize defaults for sss_config
+        if "matchLabels" not in sss_config:
+            sss_config["matchLabels"] = {}
```
rogbas: Shouldn't we default to the `managed = true` label in case none is specified in `sss-config.yaml`?

Otherwise, we'll apply the SSS in non-managed clusters as well

scripts/generate_syncset.py
jewzaam: doh, yes, removing...

scripts/generate_syncset.py
jewzaam: Defaults are in the SSS template.  https://github.com/openshift/managed-cluster-config/blob/master/scripts/templates/selectorsyncset.yaml#L10-L13

rogbas: hmm, nice

PR #233, opened 2020-02-25
Update pruning configuration

cblecker:
- Updates the pruning update to use the imagestream again, so this is an 4.3 version of `oc`. BZ 1755155 (https://bugzilla.redhat.com/show_bug.cgi?id=1755115) was fixed as of 4.3, so the `cli:latest` imagestream on cluster should work, and should always track the cluster version now.
- Use the `service-ca.crt` certificate for the image registry, so that we don't need to allow insecure connections


cblecker:
I've tested this on my own cluster, and it works as expected.

/assign @jewzaam

PR #234, opened 2020-02-25
Create new CronJob to test builds

jewzaam:
https://issues.redhat.com/browse/OSD-2965

Won't rely on customer builds anymore so will be more resiliant.


deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -0,0 +1,39 @@
...
+          serviceAccountName: sre-build-test
+          restartPolicy: Never
+          containers:
+          - name: sre-build-test
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - /bin/sh
+            - -ec
+            - "export NS=openshift-build-test-$(date +%s) ; oc create ns $NS && oc -n $NS new-build nodejs~https://github.com/openshift/nodejs-ex --name=sre-build-test; echo $(date): Waiting for build to complete.; while :; do oc -n $NS get build -o custom-columns=STATUS:.status.phase --no-headers | grep Complete && break; echo $(date): Checking build again in 15 seconds, NS=$NS; sleep 15; done; echo $(date): Done, deleting build NS=$NS; oc delete ns $NS"
```
lisa: I think it would be more readable if this was contained in a proper shell script, perhaps mounted into the container via a ConfigMap.

Also, I think the node build uses a lot of resources (ISTR a discussion around the resource usage) and that a Ruby-based one might be better suited?

deploy/sre-build-test/70-stuck-builds.PrometheusRule.yaml
```diff
@@ -11,7 +11,12 @@ spec:
   - name: sre-alerts-stuck-builds
     rules:
     - alert: StuckNewBuilds3MinSRE
-      expr: count(openshift_build_active_time_seconds{phase="New"} < time() - 180) > 0
+      expr: count(openshift_build_active_time_seconds{phase="New",exported_namespace=~"^openshift-build-test.*"} < time() - 180) > 0
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+    - alert: StuckPendingBuilds3MinSRE
+      expr: count(openshift_build_active_time_seconds{phase="Pending",exported_namespace=~"^openshift-build-test.*"} < time() - 180) > 0
```
lisa: Is it possible to write these using `for` instead of the time calculation?

(I'll take the blame for using a time calculation in the first place, in other rules. 😅 )

deploy/sre-build-test/50-build-test.CronJob.yaml
jewzaam: Is this request for a CM a blocker to merge?

Re resources, happy to pick something else but am just basing this on what we do for all OSD v3.. not concerned w/ the resource utilization.

deploy/sre-build-test/70-stuck-builds.PrometheusRule.yaml
jewzaam: haha, thanks for taking the blame ;)

Didn't look at changing this one other than adding the Pending variant.

deploy/sre-build-test/50-build-test.CronJob.yaml
lisa: It's a blocker in the sense that reviewing a 400+ character one-liner is hard to review. Splitting it into a script lets us get away from such a long line, and grant readability. Yes, I could cut and paste it into an editor and turn it into a multi-liner, but future changes to this will also be challenging to do.

jewzaam: Updated to use `>-` so it's broken into multiple lines in the repo.  When it's deployed to the cluster it's munged together again with `\n` added in, but this makes it easier to reason about in code.

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -0,0 +1,51 @@
...
+          serviceAccountName: sre-build-test
+          restartPolicy: Never
+          containers:
+          - name: sre-build-test
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - /bin/sh
+            - -ec
+            - >-
```
lisa: this works? very cool.

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -0,0 +1,51 @@
...
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: sre-build-test
+  namespace: openshift-build-test
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Forbid
+  schedule: "11 */1 * * *"
```
lisa:
```suggestion
  schedule: "11 * * * *"
```

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -0,0 +1,51 @@
...
+            effect: NoSchedule
+          serviceAccountName: sre-build-test
+          restartPolicy: Never
+          containers:
+          - name: sre-build-test
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - /bin/sh
+            - -ec
```
lisa: The `-e` option here (which I like) will cause the CronJob to litter namespaces should the build fail for whatever reason prior to line 51. I'm not sure how you want to handle that.

jewzaam: Good point, this can be removed.  I want the job to wait forever if there's a problem.  Updating..

lisa: Is it desirable to wait forever? `Job`s (and by extension `CronJob`s) have an innate retry (by default, it's 6 times) if the thing fails.

I don't know what happens with `CronJob` failure accounting if the wait is very long. I do know that at one point failed `CronJob`s could stop triggering _at all_ after a certain number of failures (I think 100), although that might have been changed, as that was ca. 2017 I ran into it.

deploy/sre-build-test/50-build-test.CronJob.yaml
jewzaam: Yes, confirmed the full thing on a stage cluster I had never touched.

deploy/sre-build-test/70-stuck-builds.PrometheusRule.yaml
jewzaam: To be clear, if we want to change the rules I recommend tracking elsewhere, I didn't want to get into changing alerts except to point at limited namespaces.  I noticed the "Pending" state so added that alert..

lisa: No, I agree. I had thought a "clean as you go" approach would work, but a wholistic approach is probably better.

deploy/sre-build-test/50-build-test.CronJob.yaml
jewzaam: Set `concurrencyPolicy: Forbid` [1] expecting that to keep it from trying if the job is already running.  I didn't want it to delete the generated test namespace.  If the build is stuck we need it to stay around for the alert to trigger and persist.  If there's some other problem I think we have `KubeJobCompletion` alert that will fire for this job after 1 hour.

[1] https://github.com/openshift/managed-cluster-config/pull/234/files#diff-016fe875d079f89b3ad58cdfcd27de45R10

lisa: If the build is stuck won't the existing `KubeJobCompletion` rule already cover us?

jewzaam: No, I see this alert in a prod cluster for a broken customer build.  There's no other alerts in cluster.  The reason I want to create a new build for a targeted alert is so broken customer builds don't page SRE.

lisa: I understand.

PR #238, opened 2020-02-28
Add GC options to custom-kubelet

jewzaam:
https://issues.redhat.com/browse/OSD-3010

### DIFF
--- a/deploy/kubelet-config/03-kubelet-config.yaml
+++ b/deploy/kubelet-config/03-kubelet-config.yaml
@@ -10,3 +10,5 @@ spec:
     systemReserved:
       cpu: 1000m
       memory: 1Gi
+    ImageGCLowThresholdPercent: 30 # https://issues.redhat.com/browse/OSD-3010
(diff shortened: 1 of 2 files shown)


deploy/kubelet-config/03-kubelet-config.yaml
```diff
@@ -10,3 +10,5 @@ spec:
     systemReserved:
       cpu: 1000m
       memory: 1Gi
+    imageMinimumGCAge: 30 # https://issues.redhat.com/browse/OSD-3010
```
cblecker: This is going from "2m" default to 30 seconds

deploy/kubelet-config/03-kubelet-config.yaml
```diff
@@ -10,3 +10,5 @@ spec:
...
+    imageGCHighThresholdPercent: 35 # https://issues.redhat.com/browse/OSD-3010
```
cblecker: And this is going from 85% down to 35? This seems like a significant drop, especially without changing `ImageGCLowThresholdPercent`

PR #243, opened 2020-03-06
dedicated-admin: dns fix and dns config map

mwoodson:


deploy/rbac-permissions-operator-config/03-dedicated-admins-dns.RoleBinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: dedicated-admins-openshift-dns
+  namespace: openshift-dns
+subjects:
```
rogbas: This Rolebinding is not granting access to dedicated-admin ServiceAccounts, only to real users? 
Is that expected?

PR #245, opened 2020-03-09
Remove hard coded namespace label from sre-filesysfull alerts

jewzaam:
https://issues.redhat.com/browse/OSD-3085


jewzaam:
/close

Not sure why we have this alert, we cannot act if it's a customer PV.

PR #246, opened 2020-03-09
Delete sre-filesysfull alerts, SRE will not get alerted on customer namespaces

jewzaam:
https://issues.redhat.com/browse/OSD-3085


jewzaam:
/hold

Need to be sure this is the right path forward, so holding it for a bit.

PR #252, opened 2020-03-11
Removing resourceNames from CS SRE clusterresourcequotas  RBAC

alanmoran:
TIL: Create requests cannot be restricted by resourceName, as the object name is not known at authorization time.

CS SRE creates two clusterresourcequotas at provisioning time.

### DIFF
--- a/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
+++ b/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
@@ -59,6 +59,3 @@ rules:
   - clusterresourcequotas
   verbs:
   - '*'
-  resourceNames:
-  - rhmi-loadbalancer-quota
-  - rhmi-persistent-volume-quota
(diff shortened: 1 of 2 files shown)


deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
```diff
@@ -59,6 +59,3 @@ rules:
   - clusterresourcequotas
   verbs:
   - '*'
-  resourceNames:
-  - rhmi-loadbalancer-quota
-  - rhmi-persistent-volume-quota
```
jewzaam: Removing this means CS SRE can manage any clusterresourcequota.  For non-CCS clusters we have service load balancer and storage quotas in place.  Would it work to split out "create" without resourceName limitations and other verbs restricted to these two resources?  I assume these two are the ones that need to be managed.

alanmoran: So this will be removed for GA. It is a temporary measure for LA (CCS). RHMI won't be available on non-CCS clusters during LA. I'll add a JIRA to the GA epic to track removing this.

jewzaam: Generally would prefer we don't open up this gap but hive will reconcile any changes in core OSD quotas for non-CCS clusters so it will be a mitigated risk.

I'll approve but will ask someone else in SREP to sanity check for a lgtm.

PR #256, opened 2020-03-19
Grant customer installed opeartor's CRD RBAC to dedicated-admins

jewzaam:
https://issues.redhat.com/browse/OSD-3153


deploy/rbac-permissions-operator-config/05-dedicated-admins-aggregate-cluster.ClusterRole.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  annotations:
-    authorization.openshift.io/system-only: "true"
```
jewzaam: This is probably intended to ensure the clusterrole doesn't show up in the web console.  Finding a reference is tough.  Best I found is https://github.com/openshift/origin/issues/16862 .  As a dedicated-admin, I can see this in the console so it's not doing that.  And we don't use it on other clusterroles.  So it's gone!

PR #258, opened 2020-03-19
allow dedicated-admins to run `oc adm top {node,pod}`

yithian:
for https://issues.redhat.com/browse/OSD-2023


yithian:
`make` deleted a **_TON_** of stuff from `hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl`

yithian:
I found the right combination of SELinux contexts for files in the repo to allow `make` to run correctly. I've ammended my commit to include the stuff that was removed from `hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl` due to the missing contexts.

PR #262, opened 2020-03-23
Wait for 5 minutes before firing the DNS error alerts. Also add messa…

karthikperu7:
…ge annotations to debug the issue faster.

https://issues.redhat.com/browse/OSD-2770


deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
```diff
@@ -17,6 +17,9 @@ spec:
         namespace: openshift-monitoring
     - alert: DNSErrors05MinSRE
       expr: rate(dns_failure_failure_total[5m]) > 0
+      for: 5m
```
cblecker: Is this necessary? The rate in the expression is looking for failures over the last 5 minutes.

karthikperu7: yes. However as soon as there is data point for rate(dns_failure_total[5m]) > 0 it will go to straight 'firing' (which means a single failure in the last minute will fire the alert) without waiting in 'pending' state. This results in unnecessary flapping alerts where SRE cannot do much. Therefore adding 'for: 5m' makes the alert wait for 5m with dns test failures before firing and sending a PagerDuty alert.

PR #265, opened 2020-03-26
Added custom alert for tracking monthly SLA

superfunhappytime:
Added a custom alert to track monthly SLA breaches. This alert is then used in thanos/grafana to calculate monthly SLA values.

### DIFF
--- a/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  name: sre-uptime-sla
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-uptime-sla
+    rules:
+    - alert: SLAUptimeSRE
+      annotations:
+        message: The API server has had 100 percent request failures for 300 seconds.
+      expr: |
+        (sum(rate(apiserver_request_count{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_count{job="apiserver"}[5m])) >= 1)
+      for: 5m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -4377,6 +4377,30 @@ objects:
       kind: PrometheusRule
       metadata:
         labels:
+          prometheus: k8s
+          role: alert-rules
+        name: sre-uptime-sla
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-uptime-sla
+          rules:
+          - alert: SLAUptimeSRE
+            annotations:
+              message: The API server has had 100 percent request failures for 300
+                seconds.
+            expr: '(sum(rate(apiserver_request_count{job="apiserver", code=~"(400|5..)"}[5m]))
+              / sum(rate(apiserver_request_count{job="apiserver"}[5m])) >= 1)
+
+              '
+            for: 5m
+            labels:
+              severity: warning
+              namespace: openshift-monitoring
+    - apiVersion: monitoring.coreos.com/v1
+      kind: PrometheusRule
+      metadata:
+        labels:
           prometheus: osd-sre-cluster-admins
           role: alert-rules
         name: osd-sre-cluster-admins


deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
```diff
@@ -0,0 +1,21 @@
...
+    rules:
+    - alert: SLAUptimeSRE
+      annotations:
+        message: The API server has had 100 percent request failures for 300 seconds.
+      expr: |
+        (sum(rate(apiserver_request_count{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_count{job="apiserver"}[5m])) >= 1)
+      for: 5m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
```
jewzaam: SRE will be alerted in PD because it's a warning and it's in a core namespace.  Based on the PR description it appears this is not desired, this is for reporting elsewhere.  Can we set the severity to something other than `warning` or `critical` and have it go to telemeter still?  If not, need to see this paired with a PR to configure-alertmanager-operator to suppress (route to `null`) this alert.

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-uptime-sla
+    rules:
+    - alert: SLAUptimeSRE
+      annotations:
+        message: The API server has had 100 percent request failures for 300 seconds.
+      expr: |
+        (sum(rate(apiserver_request_count{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_count{job="apiserver"}[5m])) >= 1)
```
jewzaam: @c-e-brumm I don't think this aligns with the way we are measuring uptime.  Could you comment on this with respect to the language in the contract appendix?

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
oddtazz: Would it be possible to use `silent-test` alerts for this, considering they already exist?

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
oddtazz: For reference this is what it says in the appendix:
`Monthly Availability = 
100 * [(Total number of 5 minute intervals in a calendar month - Number of Downtime intervals) /  (Total number of 5 minute intervals in a calendar month)] ` 

https://docs.google.com/document/d/1Qa3tVePF94ZzrThMyvqNySwLwtQGxVshqIHS8D_N3XQ/edit?skip_itp2_check=true&pli=1#

c-e-brumm: @jewzaam After some discussion we went back to measuring by 5 minute increments with no extra math. This alert is part of measuring that.

jewzaam: Synced w/ @c-e-brumm off PR.  Summary is we do expect to report on outages less than 5 minutes and using `rate(...[5m])` does not meet that need.

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
superfunhappytime: Thanks @jewzaam - Thanos requires either `warning` or `critical` so please find the paired PR for configure-alertmanager-operator here - https://github.com/openshift/configure-alertmanager-operator/pull/71; this dev/nulls the alert

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
superfunhappytime: @jewzaam 

From the text in the link:

```“Monthly Availability” means the total number of five minute intervals in a calendar month, subtracting the number of Downtime intervals during the calendar month and dividing that number by the total number of five minute intervals in the calendar month.  ```

Where a Downtime interval is defined as:
```“Downtime” means a period of more than five minutes in which all connection requests to the OpenShift Dedicated Cluster’s API server fails. ```

Could you please clarify a little more as I can't find anywhere that we must report on outages less then 5 minutes.
Happy to have a sync off-PR too

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
cblecker: side node: warning/critical alerts, even if routed away in alert manager, will still show up in the cluster health view in OCM. I don't know if this is desired or not.

c-e-brumm:
We've sync'd offline that this alert is what we're looking for. But if there is still concern we can modify it later.

PR #267, opened 2020-03-27
Add PrometheusRules for managed-velero-operator

mbarnes:
Initial alerting rules for `managed-velero-operator`.

Note, the rule severity is set to "warning" until we have an SOP for each alert.

### DIFF
--- a/deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
@@ -1,0 +1,49 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-managed-velero-operator-alerts
+    role: alert-rules
+  name: sre-managed-velero-operator-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-managed-velero-operator-alerts
+    rules:
+    # Execution times for Velero backup jobs are fairly consistent, so these
+    # alerts measure the time since the last successful backup finished (or
+    # the deployment's creation time if there's been no backup yet), plus a
+    # small static buffer to account for some variation.
+    - alert: VeleroHourlyObjectBackupMissed
+      # Hourly backup should complete about 3600 seconds (1 hour) after the
+      # last successful backup, with a 600 second (10 minute) buffer.
+      expr: |
+        time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="hourly-object-backup"},
+        scalar(max(kube_deployment_created{namespace="openshift-velero",deployment="velero"}))) > 3600 + 600
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+      annotations:
+        message: The hourly Velero backup has not successfully completed
+    - alert: VeleroDailyFullBackupMissed
+      # Daily backup should complete about 86,400 seconds (1 day) after the
+      # last successful backup, with a 600 second (10 minute) buffer.
+      expr: |
+        time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="daily-full-backup"},
+        scalar(max(kube_deployment_created{namespace="openshift-velero",deployment="velero"}))) > 86400 + 600
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+      annotations:
+        message: The daily Velero backup has not successfully completed
+    - alert: VeleroWeeklyFullBackupMissed
+      # Weekly backup should complete about 604,800 seconds (1 week) after the
+      # last successful backup, with a 600 second (10 minute) buffer
+      expr: |
+        time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="weekly-full-backup"},
+        scalar(max(kube_deployment_created{namespace="openshift-velero",deployment="velero"}))) > 604800 + 600
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+      annotations:
+        message: The weekly Velero backup has not successfully completed
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -4757,6 +4757,54 @@ objects:
       kind: PrometheusRule
       metadata:
         labels:
+          prometheus: sre-managed-velero-operator-alerts
+          role: alert-rules
+        name: sre-managed-velero-operator-alerts
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-managed-velero-operator-alerts
+          rules:
+          - alert: VeleroHourlyObjectBackupMissed
+            expr: 'time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="hourly-object-backup"},
+
+              scalar(max(kube_deployment_created{namespace="openshift-velero",deployment="velero"})))
+              > 3600 + 600
+
+              '
+            labels:
+              severity: warning
+              namespace: openshift-monitoring
+            annotations:
+              message: The hourly Velero backup has not successfully completed
+          - alert: VeleroDailyFullBackupMissed
+            expr: 'time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="daily-full-backup"},
+
+              scalar(max(kube_deployment_created{namespace="openshift-velero",deployment="velero"})))
+              > 86400 + 600
+
+              '
+            labels:
+              severity: warning
+              namespace: openshift-monitoring
+            annotations:
+              message: The daily Velero backup has not successfully completed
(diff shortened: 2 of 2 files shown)


cblecker:
/hold
We need to have the metrics available before this can merge

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,50 @@
...
+spec:
+  groups:
+  - name: sre-managed-velero-operator-alerts
+    rules:
+    - alert: VeleroPodOfflineSRE
+      expr: kube_deployment_status_replicas_available{namespace="openshift-velero",deployment="velero"} == 0
+      for: 15m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
```
cblecker, line 18: @jewzaam if these alerts relate to a specific operator, is it better for them to be in the operators NS? or do we prefer openshift-monitoring?

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,50 @@
...
+      annotations:
+        message: Velero pod has been missing or unavailable for 15m
+    - alert: VeleroHourlyObjectBackupMissed
+      expr: |
+        velero_backup_attempt_total{namespace="openshift-velero",schedule="hourly-object-backup"} > 0 and
+        time() - velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="hourly-object-backup"} > 3600
+      for: 30m
```
cblecker, line 25: It would be great to have some comments to outline what these are actually doing.

If I read this right.. we are verifying that we've at least attempted a backup, and that the "time since last successful" is greater than 1 hour.

Why do we have the 30 minute pending delay on that?

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+      annotations:
+        message: The hourly Velero backup has not successfully completed
+    - alert: VeleroDailyFullBackupMissed
+      expr: |
+        velero_backup_attempt_total{namespace="openshift-velero",schedule="daily-full-backup"} > 0 and
+        time() - velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="daily-full-backup"} > 86400
+      for: 12h
```
cblecker, line 35: Why the 12 hour pending delay?

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+      annotations:
+        message: The daily Velero backup has not successfully completed
+    - alert: VeleroWeeklyFullBackupMissed
+      expr: |
+        velero_backup_attempt_total{namespace="openshift-velero",schedule="weekly-full-backup"} > 0 and
+        time() - velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="weekly-full-backup"} > 604800
+      for: 84h
```
cblecker, line 45: Why the 84h pending delay?

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
mbarnes: The delay is there to add wiggle room for backup jobs, on the assumption that one backup may take a little longer than the previous to complete.  The delay is arbitrary but consistent: half the backup frequency (hourly -> 30m / daily -> 12h / weekly -> 84h).

So for example if an hourly backup hasn't completed in the last hour-and-a-half, assume something's wrong and raise an alert.

Those values can certainly be tweaked; it was just an initial guess.

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,50 @@
...
+    prometheus: sre-managed-velero-operator-alerts
+    role: alert-rules
+  name: sre-managed-velero-operator-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-managed-velero-operator-alerts
+    rules:
+    - alert: VeleroPodOfflineSRE
+      expr: kube_deployment_status_replicas_available{namespace="openshift-velero",deployment="velero"} == 0
```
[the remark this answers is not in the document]
mbarnes: Possibly this is redundant if there's some standard set of alerts for protected namespaces.

mbarnes:
> I noted we aren't tracking the failures metrics. Do we get appropriate coverage using the ones you've chosen?

I was concerned about how alerts would ever clear when using the failure metrics, which as I understand only increment.  Open to suggestions there.

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
cblecker: The duration of jobs should be fairly consistent. The way that `for` is used here is "the alert needs to be active for this long before we take action". We want to leave enough time for the backup to complete, but we can find that out with something like `rate(velero_backup_duration_seconds_sum{namespace="openshift-velero",schedule="hourly-object-backup"}[5h])/rate(velero_backup_duration_seconds_count{namespace="openshift-velero",schedule="hourly-object-backup"}[5h])` (right now it looks like about 60 seconds, but the daily/weekly might be a bit longer with the snapshot API calls).

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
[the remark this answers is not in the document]
cblecker: Check out the `KubeDeploymentReplicasMismatch` default alert

mbarnes:
Second iteration make the following changes:

- Remove `VeleroPodOfflineSRE` since it's covered by `KubeDeploymentReplicasMismatch`.
- Remove the `for` clause from the `*BackupMissed` alerts.
- Attempt a different strategy to avoid alerting immediately on startup: use the Velero deployment's creation timestamp as a fallback when there's no "last successful backup" timestamp.  (`velero_backup_attempt_total > 0` was problematic because it's incremented when the backup _begins_, and `velero_backup_last_successful_timestamp` would still be 0 before the first attempt completes)
- Use the average backup duration in estimating when the next backup should complete.

deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,45 @@
...
+    role: alert-rules
+  name: sre-managed-velero-operator-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-managed-velero-operator-alerts
+    rules:
+    - alert: VeleroHourlyObjectBackupMissed
+      expr: |
+        time() - clamp_min(velero_backup_last_successful_timestamp{namespace="openshift-velero",schedule="hourly-object-backup"},
```
cblecker, line 15: so if I understand this correctly:
current time - (the lower of the last successful backup, OR when the deployment was created): in a working cluster, this should be somewhere between 0 and 3600 (one hour)

so we take that, and measure if it's greater than

3600 + the average backup duration for the last 5 hours

This is really cool. Two thoughts.. 
Should we add a buffer? Say like a static extra 5 mins.. or some added percentage above the average.. just in case it's slightly longer than the average?
Would it be good to also clamp that average to a max, so if it can never go over a certain amount (say a static 30 minutes on each of these jobs)?

mbarnes: You got it right except the average backup time itself acts as the buffer, because `velero_backup_last_successful_timestamp` marks when the last backup _finished_, not started.  So this actually gives it roughly the span of _two_ backup durations to finish: the hourly backup should be _finishing_ about 3600 seconds after the last one finished (since the durations are pretty consistent), plus the average time tacked on as a buffer.

The average calculation might be overkill though if, like you suggest, a constant buffer would suffice.  Or we could tighten the buffer and make it a fraction of the average, say like `3600 + (AVERAGE * 0.25)`.  Also good idea on clamping the average.

Sounds like it's just a matter of tuning now.

cblecker: Ohhh okay. That makes sense.

My gut says it probably doesn't have to be that complex. Right now the average time for a base cluster is about 60 seconds.. if we give a static buffer of 5 or 10 minutes from the last complete time, that might be enough. That way, we don't have to clamp it, it's just a static number + the last completion time.

cblecker:
The other thing I'd love is a `# comment` just describing the logic of what we are doing with these alerts. That will help future SREs understand how things are calculated.

mbarnes:
Simplified the expressions and added comments.

cblecker:
/lgtm
/approve
/hold cancel

All backups will first kick off when velero is initially provisioned, and then they go to the cron schedule after that.

PR #273, opened 2020-03-31
Adding SelectorSyncIdentityProvider for Google Auth

rogbas:
- Adding SelectorIdentityProvider for Google IDP - applying to production only
- Updating the existing LDAP SelectorIdentityProvider to match non-production

PR #274, opened 2020-03-31
Only curate operators for non-CCS clusters

jewzaam:
https://issues.redhat.com/browse/OSD-2681


jewzaam:
/hold
let https://github.com/openshift/managed-cluster-config/pull/273 merge first (it's more important and I don't want to cause a rebase or anything)

PR #279, opened 2020-04-03
Allow generating sss per matchlabel

alanmoran:
Allows the use of a matchLabelsApplyMode flag (OR). When set all matchLabels are applied as separate SSS.

This is to tackle a situation where we want to apply configuration across multiple labels. (RHMI as an example)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ There is a limited configuration available at this time.  The file `sss-config.y
 
 * matchLabels (default: `{}`) - adds additional `matchLabels` conditions to the `clusterDeploymentSelector`
 * resourceApplyMode (default: `"Sync"`) - sets the `resourceApplyMode`
+* matchLabelsApplyMode (optional, default: not set) - When set as `"OR"` generates additional SSS per `matchLabels` conditions. This is to tackle a situation where we want to apply configuration to multiple labels.
 
 # Selector Sync Sets included in this repo
 
--- a/scripts/generate_syncset.py
+++ b/scripts/generate_syncset.py
@@ -130,11 +130,15 @@ if __name__ == '__main__':
         if "matchLabelsApplyMode" in sss_config and sss_config["matchLabelsApplyMode"] == "OR":
             # generate new SSS per matchLabels line
             for key, value in sss_config['matchLabels'].items():
+
                 o = copy.deepcopy(sss_config)
                 o['matchLabels'].clear()
                 o['matchLabels'].update({key:value})
                 del o["matchLabelsApplyMode"]
-                process_yamls(sss_name, dirpath, selectorsyncset_data, o)
+
+                # SSS objects require unique names
+                unique_sss_name = sss_name + key.replace('api.openshift.com/', '-')
+                process_yamls(unique_sss_name, dirpath, selectorsyncset_data, o)
         else:
             # Catches anyone with a rouge value
             process_yamls(sss_name, dirpath, selectorsyncset_data, sss_config)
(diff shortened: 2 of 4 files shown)


scripts/generate_syncset.py
```diff
@@ -126,7 +126,22 @@ def process_yamls(name, directory, obj, sss_config):
...
+                # generate new SSS per matchLabels line
+                for key, value in sss_config['matchLabels'].items():
+                    o = copy.deepcopy(sss_config)
+                    o['matchLabels'].clear()
+                    o['matchLabels'].update({key:value})
+                    del o["matchLabelsApplyMode"]
+                    process_yamls(sss_name, dirpath, selectorsyncset_data, o)
+            else:
+                 # Catches anyone with a rouge value
+                process_yamls(sss_name, dirpath, selectorsyncset_data, sss_config)
```
jewzaam, line 144:
```suggestion
        if "matchLabelsApplyMode" in sss_config and sss_config["matchLabelsApplyMode"] == "OR":
            # generate new SSS per matchLabels line
            for key, value in sss_config['matchLabels'].items():
                o = copy.deepcopy(sss_config)
                o['matchLabels'].clear()
                o['matchLabels'].update({key:value})
                del o["matchLabelsApplyMode"]
                process_yamls(sss_name, dirpath, selectorsyncset_data, o)
        else:
            # Catches anyone with a rouge value
            process_yamls(sss_name, dirpath, selectorsyncset_data, sss_config)
```

alanmoran: Changed. Much cleaner. thank you

README.md
```diff
@@ -17,6 +17,7 @@ There is a limited configuration available at this time.  The file `sss-config.y
 
 * matchLabels (default: `{}`) - adds additional `matchLabels` conditions to the `clusterDeploymentSelector`
 * resourceApplyMode (default: `"Sync"`) - sets the `resourceApplyMode`
+* matchLabelsApplyMode (optional, default: not set) - When set as `"OR"` generates additional SSS per `matchLabels` conditions. This is to tackle a situation where we want to apply configuration to multiple labels.
```
jewzaam, line 20:
```suggestion
* matchLabelsApplyMode (optional, default: not set) - When set as `"OR"` generates a separate SSS per `matchLabels` conditions. Default behavior creates a single SSS with all `matchLabels` conditions.  This is to tackle a situation where we want to apply configuration for one of many label conditions.
```

alanmoran:
@cblecker  - Good catch with the name. I'll have a think about how to address that one.

>  I'm also concerned about the future extensibility of this design.. what if you need to match on a set of three labels (managed=true, rhmi=true, cloud-provider=aws, for example)

I'm not sure what you mean here. If it is to match all 3 then you wouldn't use `matchLabelsApplyMode` and it would use the default behaviour where the labels are ANDed.

jewzaam:
> @cblecker - Good catch with the name. I'll have a think about how to address that one.

Append index?

> 
> > I'm also concerned about the future extensibility of this design.. what if you need to match on a set of three labels (managed=true, rhmi=true, cloud-provider=aws, for example)
> 
> I'm not sure what you mean here. If it is to match all 3 then you wouldn't use `matchLabelsApplyMode` and it would use the default behaviour where the labels are ANDed.

What you couldn't do with this design is have a mix of OR and AND conditions, other than the managed=true which is enforced outside of this in the base SSS template.  If we need that we could restructure the sss-config.yaml to allow a better expression of conditions.  Do we need that?  I think not but...

alanmoran:
> Append index?

I thought about this but if we were doing this to several "add-ons" it would be confusing but happy to change to index if needed. 

I don't think a mix of OR and AND conditions is really in scope of this PR. I don't see a use case for it right now. We do have one for RHMI :)

jewzaam:
/approve

@cblecker if OK with this please lgtm and remove the hold, it seems like a reasonable compromise

PR #282, opened 2020-04-07
use hive.openshift.io/v1

staebler:


cblecker:
This LGTM, but I'd like to have @jewzaam review it specifically for the cluster upgrade script bits

/lgtm
/approve
/hold
/assign @jewzaam 

@jewzaam: if it looks good to you, feel free to remove the hold

jewzaam:
looks reasonable, we'll try it out before upgrading to verify nothing is broken.  thanks!
/hold cancel

PR #283, opened 2020-04-08
Deploy managed-velero-operator:v0.2.133-c6455c5 to stage

cblecker:
This upgrades the managed-velero-operator to `v0.2.133-c6455c5`.

The key feature being deployed is GCP support.

Detailed changes:
https://github.com/openshift/managed-velero-operator/compare/ededc3e...c6455c5

### DIFF
--- a/deploy/managed-velero-operator/130-velero.CredentialRequest.yaml
+++ b/deploy/managed-velero-operator/130-velero.CredentialRequest.yaml
@@ -1,7 +1,7 @@
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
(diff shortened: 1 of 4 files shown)


lisa:
@cblecker How do the clusters fare with having both CredentialsRequests (one for AWS and one for GCP)? I'm guessing they will not have the opposite cloud provider configured and silently ignore the one not for them?

cblecker:
@lisa Yup! That's exactly right: https://github.com/openshift/cloud-credential-operator/blob/01b2676518a76119fe9f09f18fe78663b34df717/pkg/controller/credentialsrequest/credentialsrequest_controller.go#L310-L324

PR #284, opened 2020-04-09
Grant cluster wide imagestream READ permission to dedicated-admins

wanghaoran1988:
https://issues.redhat.com/browse/OSD-3294


cblecker:
/lgtm
/approve
/hold

This looks good.. the one concern I have is double checking with the RHMI folks that they are cool with this change too. By default on OSD, we just have the system imagestreams in the `openshift` namespace.. I don't know if adding on RMHI changes that or if they have any concerns.

wanghaoran1988:
Thanks @alanmoran
/hold cancel

PR #285, opened 2020-04-09
OSD-3193 Create SSS for cluster-admin Rolebinding

rogbas:
Create SelectorSyncSet for cluster-admin rolebinding for clusters with CD labelled as `api.openshift.com/cluster-admin: "true"`


jewzaam:
/lgtm
/hold

remove hold when ready to merge, this should be fine to promote too given it requires the label and there's no way it's going to be applied unless manually

PR #287, opened 2020-04-13
Add efs-csi

2uasimojo:
efs-csi is a SelectorSyncSet that enables EFS on AWS clusters via the
CSI driver at the behest of label `managed.openshift.io/efs-csi: "true"`.


README.md
```diff
@@ -73,6 +69,24 @@ To opt-in to logging, the customer must:
 2. install the `elasticsearch` operator
 3. create `ClusterLogging` CR in `openshift-logging`
 
+## EFS Enablement via CSI
+
+[`efs-csi`](deploy/efs-csi) enables AWS EFS via the CSI driver. Customer
+opts in by opening a SNow ticket, whereupon SRE must add the appropriate
```
dustman9000, line 75: I think we format it as `SNOW` instead of `SNow`...

deploy/efs-csi/01-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
+# Source: https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/51d19a433dcfc47fbb7b7a0e1c8ff6ab98ce87e9/deploy/kubernetes/base/node.yaml
```
dustman9000, line 1: Can you add a YAML document separator `---` at the top of all of these manifests?

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -643,6 +643,180 @@ objects:
...
+        selector:
+          matchLabels:
+            app: efs-csi-node
+        template:
+          metadata:
+            labels:
+              app: efs-csi-node
+          spec:
+            nodeSelector:
+              beta.kubernetes.io/os: linux
```
dustman9000, line 676: To be on only worker nodes:
```suggestion
              node-role.kubernetes.io/worker: ""
```

deploy/efs-csi/01-daemonset.yaml
2uasimojo: I was going to, but am following precedent. Only four out of the existing 133 yaml files under `deploy` do so. The builder is putting the chunks into a single yaml file anyway, so arguably sans-separator makes more sense. :shrug:

README.md
2uasimojo: I just expanded it to ServiceNow

deploy/efs-csi/sss-config.yaml
```diff
@@ -0,0 +1,2 @@
+matchLabels:
+  hive.openshift.io/efs-csi: "true"
```
cblecker, line 2: we should also be checking for `hive.openshift.io/cluster-platform: "aws"`

deploy/efs-csi/05-scc.yaml
```diff
@@ -0,0 +1,39 @@
+# A clone of the `anyuid` scc allowing dedicated-admins to create pods
```
cblecker, line 1: This is concerning, as this may allow root pods.

deploy/efs-csi/04-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+# Lets dedicated-admins create a PersistentVolume
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: system:persistent-volume-provisioner
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:persistent-volume-provisioner
```
cblecker, line 9: What does this role grant that dedicated admins don't already have?

deploy/efs-csi/01-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: efs-csi-node
+  namespace: kube-system
```
cblecker, line 6: is `kube-system` the right NS for this?

deploy/efs-csi/01-daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+        node-role.kubernetes.io/worker: ''
+      hostNetwork: true
+      priorityClassName: system-node-critical
+      tolerations:
+        - operator: Exists
+      containers:
+        - name: efs-plugin
+          securityContext:
+            privileged: true
+          image: amazon/aws-efs-csi-driver:latest
```
cblecker, line 26: This should be a fully qualified image name

deploy/efs-csi/05-scc.yaml
jewzaam, line 1: Is this a literal clone?  If yes, what keeps us from using `anyuid` as it is?

deploy/efs-csi/sss-config.yaml
jewzaam, line 2: What is the mechanisim we expect to have enabling this feature on OSD clusters?  If driven through OCM then label should be `api.openshift.com/efs-csi` as that's the prefix used from that system.  If we never expect it to be driven through OCM then perhaps `managed.openshift.io/efs-csi`.  The `hive.openshift.io` prefix implies it is a value set by hive and it is not.

deploy/efs-csi/04-clusterrolebinding.yaml
jewzaam: Dedicated admins cannot create PVs today, this setup requires the customer to create PVs.  Else SRE has to do it for them.. middle man kind of role.

deploy/efs-csi/sss-config.yaml
2uasimojo: Could do, though presumably SRE wouldn't set the `efs-csi` label unless it was an AWS cluster, so is it necessary?

deploy/efs-csi/01-daemonset.yaml
2uasimojo: This is a straight copy from the source (L1) (except for the nodeSelector on L17) and is how we tested. If `kube-system` isn't right, what would it be?

deploy/efs-csi/01-daemonset.yaml
2uasimojo: What repo should we use? Is docker.io acceptable?

deploy/efs-csi/04-clusterrolebinding.yaml
2uasimojo: Right.

deploy/efs-csi/05-scc.yaml
2uasimojo: That's its purpose. NFS mounts are created with root ownership, so out of the box they would be unusable to the customer. We could go fix it for them, but again that puts us in a middleman role. So instead we open this up for them and show them how to create a temporary root pod to change the permissions, and then they promise not to do it again.

2uasimojo: I'm glad you asked.

I couldn't get a patch (to add `dedicated-admins` to `anyuid`) to work idempotently. I tried all three `patchType`s. It would either replace the `groups` list completely, or add `dedicated-admins` every time it ran (so the list kept growing), or fail at the REST request for a bad header (`Accept` or `Content-Type`, can't remember which). It's possible there's a bug in whatever implementation of strategic merge we're using.

deploy/efs-csi/sss-config.yaml
2uasimojo: Okay, I had no idea, just guessed based on the fact that this was going to need to be done to a hive-managed cluster.

The mechanism is "SRE labels by hand" using `oc label clusterdeployment`. See https://github.com/openshift/ops-sop/pull/744/files?short_path=c81e05c#diff-c81e05c90920950575849b3827d2fbfc

As to whether this will _ever_ be done via OCM, I don't know, but probably not. This SSS solution is really kind of an intermediate step towards having an operator to do the heavy lifting (whereupon the label will be unnecessary? or at least different?)

So, given all that, what do you recommend?

cblecker: In short: yes. In no case should this go onto a non-AWS cluster, so adding it as a match will not be harmful, and will prevent operator error.

deploy/efs-csi/01-daemonset.yaml
jharrington22: `registry.hub.docker.com/amazon/aws-efs-csi-driver:latest` - We may want to pin the version here too @2uasimojo so they we can be sure each install is the same, use the image we used for testing.

deploy/efs-csi/05-scc.yaml
jharrington22: @cblecker this is an accepted risk, the customer will also be provided with an enhanced dedicated-admin. This is to remove SRE's involvement from the process.

deploy/efs-csi/sss-config.yaml
jharrington22: :+1: @cblecker

jharrington22: I would recommend changing it to `managed.openshift.io/efs-csi` since this feature may not be an optional addition through OCM but a default from SRE (thinking ahead). For the moment SRE will be setting it so thats what makes sense to me.

deploy/efs-csi/01-daemonset.yaml
cblecker: We should also note that if we are using docker up, we will have AWS China issues with this. cc: @dofinn @c-e-brumm

jharrington22: @cblecker what is the alternative? Quay proxy?

cblecker: There's a docker hub mirror in china, but it would require an operator on the cluster to detect it's in china and change the image

deploy/efs-csi/01-daemonset.yaml
cblecker: Probably a new NS for just this purpose. I would be reluctant to put things in kube-system that don't need to be there. `openshift-csi-drivers` maybe?

deploy/efs-csi/05-scc.yaml
cblecker: We shouldn't be modifying default SCCs to add dedicated-admins to them.. it'll break upgrades in 4.3 and will be impossible in 4.4+ (ref: https://bugzilla.redhat.com/show_bug.cgi?id=1821905)

Cloning the SCC is the right approach, if the SCC is required.

deploy/efs-csi/01-daemonset.yaml
c-e-brumm: We have not conducted a speed test of dockerhub.io from china but we have pulled images from there successfully. To be certain though, is there a way we can toggle this on/off for regions like china?

dofinn: Thus far, docker appears to be more reliable then quay.

deploy/efs-csi/01-daemonset.yaml
jewzaam: Agree, we should create a new namespace and use that if possible.  Any changes to an OCP provided namespace should only be done if we have no other options.

deploy/efs-csi/01-daemonset.yaml
jewzaam: If we push to a repo we managed on quay.io we can control the version available to OSD and have a single registry to work with for OSD in China.

deploy/efs-csi/05-scc.yaml
jewzaam: I know very little about SCCs but the docs show you can add SCC in a role.  We should try not to clone the SCC as it may change in future OCP releases.  You cannot modify it because something on cluster is managing it, either CVO or another operator. 

https://docs.openshift.com/container-platform/4.3/authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth

2uasimojo:
I worry that we may be trying to "perfect" this too much. It was supposed to be kind of a quick-and-dirty bridge solution to bank the fires of customer demand while we develop an operator to do it right. However we resolve the pending issues, we're still allowing the customer to manage PVs and create root pods, which is hacky.

Now, I know as well as anyone how quick and dirty hacks can wind up being supported for eternity. But we should also keep in mind that it's easy to iterate -- we can polish later.

deploy/efs-csi/05-scc.yaml
2uasimojo: It doesn't really matter if the `anyuid` SCC diverges from this one in the future. The important thing is that this one allows creation of a root pod, which it will continue to do regardless. This approach doesn't stop `dedicated-admins` being added to other SCCs (including `anyuid` for that matter). And we don't *want* this one to get modified in situ, which is another advantage of having it separate from `anyuid` (patch problems aside).

Reading what you linked, it does look like we could create a Role and (Cluster)RoleBinding to give access to the `anyuid` SCC. But I don't really see what difference it would make.

deploy/efs-csi/01-daemonset.yaml
jharrington22: `openshift-csi-drivers` makes sense to me :+1:

jewzaam: +1 if customer doesn't administer the namespace

2uasimojo:
dnm: I haven't regenerated the main sss yaml yet.

2uasimojo:
> dnm: I haven't regenerated the main sss yaml yet.

Now I have. This is ready for review again.

jewzaam:
> I worry that we may be trying to "perfect" this too much. It was supposed to be kind of a quick-and-dirty bridge solution to bank the fires of customer demand while we develop an operator to do it right. However we resolve the pending issues, we're still allowing the customer to manage PVs and create root pods, which is hacky.
> 
> Now, I know as well as anyone how quick and dirty hacks can wind up being supported for eternity. But we should also keep in mind that it's easy to iterate -- we can polish later.

Yea, some of these things can be shelfed such as SCC creation.. I resolved that conversation.  Where it's deployed is important when looking at this landing on our managed platform.  We need to keep day-2 operations like this cleanly isolated from core platform components unless it's not possible or it's an unreasonable amount of work.  Quick and dirty or not we will see tickets for this and have to be ready to support it.

2uasimojo:
> Where it's deployed is important when looking at this landing on our managed platform.

Okay, that's done.

deploy/efs-csi/01-daemonset.yaml
2uasimojo: I tried this and it makes the pod fail to start:

```
  Warning  FailedMount  2m37s                 kubelet, ip-10-0-139-172.ec2.internal  Unable to attach or mount volumes: unmounted volumes=[efs-mnt], unattached volumes=[default-token-6mjb2 efs-mnt]: timed out waiting for the condition
  Warning  FailedMount  30s (x10 over 4m40s)  kubelet, ip-10-0-139-172.ec2.internal  MountVolume.MountDevice failed for volume "efs-pv" : kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient: driver name efs.csi.aws.com not found in the list of registered CSI drivers
```

What now?

2uasimojo:
DNM: the non-`kube-system` namespace for the DaemonSet doesn't work.

2uasimojo:
Another ick I just discovered: the SCC kicks in by default, so your pods are rooty unless you specify otherwise. That's really not how we want to roll here.

deploy/efs-csi/01-daemonset.yaml
2uasimojo: I put this back for now so we can make forward progress.

deploy/efs-csi/05-scc.yaml
2uasimojo: Update: we don't need the SCC after all, if we use access points to set up the PV. I removed it in the latest commit.

deploy/efs-csi/01-daemonset.yaml
2uasimojo: Dug a bit more here.

The daemonset fails to create in a generic namespace with
```Error creating: pods "efs-csi-node-" is forbidden: pods with system-node-critical priorityClass is not permitted in efsap namespace```

It needs to be in the `system-node-critical` priority class because otherwise:
```
Error creating: pods "efs-csi-node-" is forbidden: unable to validate against any security context constraint: [
  provider restricted:
    .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used
    spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used
    <same for spec.volumes[1, 2, 3]>
    spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed
    spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used
    spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 9809: Host ports are not allowed to be used
    <same for spec.containers[1, 2]>
]
```

So I gather that we would need to create some kind of SCC/(Cluster)Role(Binding) etc. and attach it to our bespoke namespace. I doddled around with a couple of things (like cloning `kube-system`'s annotations) to no avail. Without a deeper understanding of the RBAC-isms, I'm shooting in the dark.

Comment (2uasimojo) [reply deploy/efs-csi/01-daemonset.yaml]: @drow and I spent some time working through this and we think we have a path forward. It entails:
- Creating our namespace, let's call it `openshift-efs-csi`.
- Creating a ServiceAccount in that namespace, let's call it `efs-csi-sa`
- Creating a SCC that
  - allows at least*:
    - privileged containers
    - hostNetwork and hostPorts
    - volumes.hostPath
  - includes user `system:serviceaccount:openshift-efs-csi:efs-csi-sa` (that last bit is `$namespace:$serviceaccount`)
- In the DaemonSet:
  - Using `metadata.namespace: openshift-efs-csi`
  - In `spec.template.spec`:
     - Removing `priorityClassName`
     - Adding `serviceAccount: efs-csi-sa`

I'll run through it a bit more to make sure, then update the PR accordingly.

This shouldn't require changes to the customer or SRE instructions.

*NB: when experimenting, we got this to work by cloning the `privileged` SCC. It is unclear whether we can reduce from there and still have stuff work.

Comment (2uasimojo) [reply deploy/efs-csi/01-daemonset.yaml]: Added a commit to the PR for this.

**Comment by 2uasimojo** [comment]
> DNM: the non-`kube-system` namespace for the DaemonSet doesn't work.

Fixed in 6bb5067d46b37f170308da64555adecb2ff8a821

> Another ick I just discovered: the SCC kicks in by default, so your pods are rooty unless you specify otherwise. That's really not how we want to roll here.

N/A since 853320d845d0ef35b88441801cff3e0c2ba5e7c0

File: README.md [hunk 9]
```diff
@@ -73,6 +69,26 @@ To opt-in to logging, the customer must:
...
 3. create `ClusterLogging` CR in `openshift-logging`
 
+## EFS Enablement via CSI
+
+[`efs-csi`](deploy/efs-csi) enables AWS EFS via the CSI driver. Customer
+opts in by opening a ServiceNow ticket, whereupon SRE must add the appropriate
+[label](deploy/efs-csi/sss-config.yaml) to the cluster. The
+SelectorSyncSet:
+
+- Installs a DaemonSet into the `kube-system` namespace, running the CSI driver
```
dustman9000, line 79: s/kube-system/openshift-efs-csi/g

dustman9000, line 79:
```suggestion
- Installs a DaemonSet into the `openshift-efs-csi` namespace, running the CSI driver
```

2uasimojo: Good catch, thanks @dustman9000

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  annotations:
+    kubernetes.io/description: 'Highly privileged SCC for the EFS CSI driver DaemonSet.'
+  name: efs-csi-scc
```
cblecker, line 6: Is this a custom SCC? Or a clone of an existing one that we are providing for this service account?

2uasimojo: Well... yes. As noted [here](https://github.com/openshift/managed-cluster-config/pull/287#discussion_r411656007) I started by cloning `privileged` to get it working. It may be possible to distill it further.

deploy/efs-csi/01-daemonset.yaml
jharrington22: This is clean to me, we are not modifying any default SCC's and we have a logically separate namespace for the ESI driver, I'm +1 good work @2uasimojo and @dustman9000 

@jewzaam @cblecker could I get a +1 from you, we'd like to get this merged today.

deploy/efs-csi/03-scc.yaml
jharrington22: :+1:

jharrington22:
@cblecker @jewzaam @2uasimojo I'd like to get this merged today so we can test a cluster in stage and promote to prod on Monday. Could we call review again please?

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+allowedCapabilities:
+- '*'
+allowedUnsafeSysctls:
+- '*'
+fsGroup:
+  type: RunAsAny
+groups:
+- system:cluster-admins
+- system:nodes
+- system:masters
```
cblecker, line 23:
```suggestion
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+seccompProfiles:
+- '*'
+supplementalGroups:
+  type: RunAsAny
+users:
+- system:admin
```
cblecker, line 34:
```suggestion
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+- system:serviceaccount:openshift-efs-csi:efs-csi-sa
+volumes:
+- '*'
```
cblecker, line 37:
```suggestion
volumes:
- hostPath
- projected
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+- system:cluster-admins
+- system:nodes
+- system:masters
+readOnlyRootFilesystem: false
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+seccompProfiles:
+- '*'
```
cblecker, line 30:
```suggestion
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+allowHostIPC: true
+allowHostNetwork: true
+allowHostPID: true
+allowHostPorts: true
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: true
+allowedCapabilities:
+- '*'
+allowedUnsafeSysctls:
+- '*'
```
cblecker, line 17:
```suggestion
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+apiVersion: security.openshift.io/v1
+metadata:
+  annotations:
+    kubernetes.io/description: 'Highly privileged SCC for the EFS CSI driver DaemonSet.'
+  name: efs-csi-scc
+allowHostDirVolumePlugin: true
+allowHostIPC: true
+allowHostNetwork: true
+allowHostPID: true
+allowHostPorts: true
```
cblecker, line 11:
```suggestion
allowHostIPC: false
allowHostNetwork: false
allowHostPID: false
allowHostPorts: false
```

deploy/efs-csi/03-scc.yaml
```diff
@@ -0,0 +1,37 @@
...
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: true
```
cblecker, line 13: This one may or may not be needed

cblecker:
We can scope down the SCC in a follow up.

/lgtm
/approve
/hold cancel

deploy/efs-csi/03-scc.yaml
cblecker: oh, we *do* need host network after all

jharrington22:
Thanks @cblecker

PR #290, opened 2020-04-15
OSD-3286 Extended dedicated-admin support for non-ccs customers using another label

rogbas:
https://issues.redhat.com/browse/OSD-3286

Adding independent `extended-dedicated-admin` label, using the new OR functionality on the `sss-config.yaml` file


jewzaam:
/lgtm

Added more done criteria around managing the old SSS if necessary.

PR #293, opened 2020-04-17
Support ClusterRole label changes in OCP 4.3.12+

jewzaam:
https://issues.redhat.com/browse/OSD-3431

Labels applied by OLM to ClusterRoles for CR RBAC changed.
This uses a different approach and fixes a gap for extended dedicated-admin when operators are installed in a customer namespace.


jewzaam:
Caused by https://github.com/operator-framework/operator-lifecycle-manager/pull/1348

PR #295, opened 2020-04-20
OSD-3429 loop for consistency in cluster_upgrade

mrbarge:
Commit c6554aced769b7d3d197ecb5ccc1c944c3521a2f unfortunately still had issues with pre-emptively declaring a cluster upgraded whilst nodes were yet to be upgraded.

This commit is by no means a great solution but attempts to overcome that problem by performing the check of the aforementioned commit N times until all N tries report consistent information. This change was successfully tested on a staging cluster going from 4.3.0 to 4.3.12 and behaved as expected.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -17,6 +17,11 @@ PD_MAINT_BASE_INFRA_COUNT=3
 # additional time to upgrade per node beyond the base count
 PD_MAINT_ADDITIONAL_NODE_MIN=8
 
+# minimum number of times the cluster must report consistent version state before proceeding
+CONSISTENT_STATE_REQUIRED_COUNT=5
+# seconds to wait in-between consistent state checks
+CONSISTENT_STATE_WAIT=60
+
 OCP_VERSION_FROM=$1
 OCP_VERSION_TO=$2
 CLUSTER_NAMES=${@:3}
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -327,24 +332,30 @@ upgrade() {
         teardown_maintenance_window $OCM_NAME $CD_NAMESPACE $CD_NAME "$PD_MAINT_DESCRIPTION_BASE"
     fi
 
-    # log current state
-    log $OCM_NAME "upgrade" "nodes upgraded: $MACHINE_COUNT_UPGRADED/$MACHINE_COUNT_ALL"
-
-    # make sure all nodes run same kubelet version
-    while [ $MACHINE_COUNT_UPGRADED -lt $MACHINE_COUNT_ALL ];
+    # only consider the upgrade done if we a successful run of consistent state reporting,
```
2uasimojo, line 335: s/if we a/if we have a/ ?

scripts/cluster-upgrade.sh
```diff
@@ -327,24 +332,30 @@ upgrade() {
...
+        then
+            # inconsistent state, reset the counter
+            CONSISTENT_STATE_COUNTER=0
+        else
+            CONSISTENT_STATE_COUNTER=$((CONSISTENT_STATE_COUNTER+1))
+        fi
+
+        sleep $CONSISTENT_STATE_WAIT
 
         MACHINE_COUNT_UPGRADED=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes -o json | jq -r ".items[].status | select(.annotations[\"machineconfiguration.openshift.io/currentConfig\"] == .annotations[\"machineconfiguration.openshift.io/desiredConfig\"]) | .conditions[] | select(.type == \"Ready\" and .status == \"True\") | .reason" | wc -l)
```
2uasimojo, line 351: This is chunky enough, and duplicated (at least from L318) to be factored out into a helper. Not really related to this change, though.

jewzaam:
/lgtm
/approve

Basically a forced wait and hope that MCO starts the next worker upgrade while waiting.  Rather we look at the desiredConfig being consistent/same for all Nodes with the same machine type.  Will lave OSD-3429 open for that change, but this PR looks good enough for the current (in-progress) round of upgrades.

ravitri:
As shared on JIRA task, I confirmed with MCO team and this is expected behaviour for MCO as of now where it will only report the upgrade of master nodes to CVO and the worker nodes will still continue to update. This would be covered in future release of OCP.

From automation perspective, it's suggested by MCO team to track progress of the worker pool's status and monitor the field "Updated" to be set as "True" to mark completion of all worker nodes.

mrbarge:
Thank you for your review, folks.

As per Ravi's notes, we can use the MachineConfigPool as a way of reliably determining when all nodes are upgraded. I have submitted a new PR that uses that method (https://github.com/openshift/managed-cluster-config/pull/301) and will close this one as the changes are substantially different.

PR #301, opened 2020-04-21
OSD-3429 Use MachineConfigPool for upgrade completion check

mrbarge:
For OSD4 cluster upgrades, we can use the MachineConfigPool data as a way of confidently identifying when all nodes are upgraded, and hence can conclude the upgrade knowing that everything is complete.

This change has been successfully tested on a fresh 4.3.0 cluster and concluded the upgrade at the right time.

Refs OSD-3429

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -310,18 +310,24 @@ upgrade() {
     log $OCM_NAME "upgrade" "ClusterVersion on $TO"
     log $OCM_NAME "upgrade" "checking node versions"
 
-    # fun fact!  after clusterversion says it is done, individual nodes could still be updated.
-    MACHINE_COUNT_ALL=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers | wc -l)
+    MASTER_MACHINEPOOL_TOTAL_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"master\" ) | .status.machineCount")
+    MASTER_MACHINEPOOL_UPDATED_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"master\" ) | .status.updatedMachineCount")
 
-    MACHINE_COUNT_UPGRADED=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes -o json | jq -r ".items[].status | select(.annotations[\"machineconfiguration.openshift.io/currentConfig\"] == .annotations[\"machineconfiguration.openshift.io/desiredConfig\"]) | .conditions[] | select(.type == \"Ready\" and .status == \"True\") | .reason" | wc -l)
+    while [ $MASTER_MACHINEPOOL_UPDATED_COUNT -lt $MASTER_MACHINEPOOL_TOTAL_COUNT ];
+    do
+        # wait for all masters to be upgraded
+        log $OCM_NAME "upgrade" "master nodes upgraded: $MASTER_MACHINEPOOL_UPDATED_COUNT/$MASTER_MACHINEPOOL_TOTAL_COUNT"
+        sleep 60
+    done
 
-    MACHINE_COUNT_PENDING=$((MACHINE_COUNT_ALL-MACHINE_COUNT_UPGRADED))
+    WORKER_MACHINEPOOL_TOTAL_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"worker\" ) | .status.machineCount")
+    WORKER_MACHINEPOOL_UPDATED_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"worker\" ) | .status.updatedMachineCount")
 
-    # start maintenance only if we need to
-    if [ ! $MACHINE_COUNT_UPGRADED -ge $MACHINE_COUNT_ALL ];
+    if [ $WORKER_MACHINEPOOL_UPDATED_COUNT -lt $WORKER_MACHINEPOOL_TOTAL_COUNT ];
     then
         # Create a new maintenance window for the kubelet (non master) upgrades based on the number of nodes.
-        MAINT_WINDOW_MIN=$((MACHINE_COUNT_PENDING*PD_MAINT_ADDITIONAL_NODE_MIN))
+        WORKERS_PENDING=$((WORKER_MACHINEPOOL_TOTAL_COUNT-WORKER_MACHINEPOOL_UPDATED_COUNT))
+        MAINT_WINDOW_MIN=$((WORKERS_PENDING*PD_MAINT_ADDITIONAL_NODE_MIN))
 
         # create worker maintenance window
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -310,18 +310,24 @@ upgrade() {
     log $OCM_NAME "upgrade" "ClusterVersion on $TO"
     log $OCM_NAME "upgrade" "checking node versions"
 
-    # fun fact!  after clusterversion says it is done, individual nodes could still be updated.
-    MACHINE_COUNT_ALL=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes --no-headers | wc -l)
+    MASTER_MACHINEPOOL_TOTAL_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"master\" ) | .status.machineCount")
+    MASTER_MACHINEPOOL_UPDATED_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"master\" ) | .status.updatedMachineCount")
```
2uasimojo, line 314: I know we're not super concerned about performance and resource consumption here, but why make two API calls when one will do?

Also, using `-o jsonpath` rather than `-o json | jq` not only saves us some subprocesses, but does all the filtering on the server side, saving network traffic. (Again, in this case the amount of data is tiny, so not really an issue.)

```
DATA=($(oc get machineconfigpools -o jsonpath="{.items[?(@.metadata.name=="'"master"'")]['.status.machineCount', '.status.updatedMachineCount']}"))
MASTER_MACHINEPOOL_TOTAL_COUNT=${DATA[0]}
MASTER_MACHINEPOOL_UPDATED_COUNT=${DATA[1]}
```

kind of thing.

Not suggesting a change here, really, just musing.

scripts/cluster-upgrade.sh
```diff
@@ -310,18 +310,24 @@ upgrade() {
...
 
-    MACHINE_COUNT_UPGRADED=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes -o json | jq -r ".items[].status | select(.annotations[\"machineconfiguration.openshift.io/currentConfig\"] == .annotations[\"machineconfiguration.openshift.io/desiredConfig\"]) | .conditions[] | select(.type == \"Ready\" and .status == \"True\") | .reason" | wc -l)
+    while [ $MASTER_MACHINEPOOL_UPDATED_COUNT -lt $MASTER_MACHINEPOOL_TOTAL_COUNT ];
+    do
+        # wait for all masters to be upgraded
+        log $OCM_NAME "upgrade" "master nodes upgraded: $MASTER_MACHINEPOOL_UPDATED_COUNT/$MASTER_MACHINEPOOL_TOTAL_COUNT"
+        sleep 60
+    done
```
2uasimojo, line 321: If the initial check determines that not everything is upgraded, this will spin forever, won't it? We need to recalculate (repeat L313-4 -- though preferably pull that out into a helper function so it's not copy/pasted) inside the loop.

scripts/cluster-upgrade.sh
```diff
@@ -310,18 +310,24 @@ upgrade() {
...
 
-    MACHINE_COUNT_PENDING=$((MACHINE_COUNT_ALL-MACHINE_COUNT_UPGRADED))
+    WORKER_MACHINEPOOL_TOTAL_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"worker\" ) | .status.machineCount")
+    WORKER_MACHINEPOOL_UPDATED_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"worker\" ) | .status.updatedMachineCount")
```
2uasimojo, line 324: This is similar enough to the code on L313-4 that the helper function suggested above could take a parameter ("worker" or "master") and DRY even more.

scripts/cluster-upgrade.sh
```diff
@@ -330,22 +336,14 @@ upgrade() {
...
-    # make sure all nodes run same kubelet version
-    while [ $MACHINE_COUNT_UPGRADED -lt $MACHINE_COUNT_ALL ];
+    while [ $WORKER_MACHINEPOOL_UPDATED_COUNT -lt $WORKER_MACHINEPOOL_TOTAL_COUNT ];
     do
-        # sleep at beginning because we've already checked that we are not done.
-        SLEEP_DURATION="$((PD_MAINT_ADDITIONAL_NODE_MIN*60/4))s"
-        sleep "$SLEEP_DURATION"
-
-        MACHINE_COUNT_UPGRADED=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get nodes -o json | jq -r ".items[].status | select(.annotations[\"machineconfiguration.openshift.io/currentConfig\"] == .annotations[\"machineconfiguration.openshift.io/desiredConfig\"]) | .conditions[] | select(.type == \"Ready\" and .status == \"True\") | .reason" | wc -l)
+        WORKER_MACHINEPOOL_UPDATED_COUNT=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get machineconfigpools -o json  | jq -r ".items[] | select(.metadata.name == \"worker\" ) | .status.updatedMachineCount")
```
2uasimojo, line 341: This should be moved below the `sleep` so we don't a) do the same call twice in a row on the first iteration, and b) do an extra unnecessary sleep on the last iteration.

scripts/cluster-upgrade.sh
ravitri: Here, we are computing the upgradecount and total count separately and monitor in a loop. That would work no doubt. But I was even thinking along the lines of fetching the status directly from machineconfigpool. Something like:

`oc get mcp -ojson | jq -r '.items[].status.conditions[] | select(.type == "Updated" and .status == "True")`

So we can check for all mcp that are configured and check for their "Updated"="True" status. This way, even if there is any separate mcp added (Eg. for infra), we can still get the work done with same logic. Plus, it required less lines of script too.

scripts/cluster-upgrade.sh
ravitri: From a point of view, running a check loop over the master machineconfigpool won't be necessary as the CVO would already handle that for the control plane. I checked this with MCO team yesterday. So, we can even simply introduce the check for worker pool check the "Updated" = True field to happen. For the timeout variable, will need to factor in the time needed for rolling update to apply change, mark node unschedulable and reboot the node. I think that would be 5-10 minutes per node. So can set it for cluster accordingly on a safer note.

scripts/cluster-upgrade.sh
jewzaam: +1, have a PR coming to this PR in a bit w/ that update. Just want to test it out.

scripts/cluster-upgrade.sh
jewzaam: I think it's safe to include this check, I will put  in a PR to update the upgraded count soon, after the sleep.

scripts/cluster-upgrade.sh
jewzaam: Not going to spend time to refactor as it's "good enough" and our end-goal is to delete this.. we don't need it to be as optimized as we can get it, just need it to work and be a reference for the upgrade operator we're designing.

scripts/cluster-upgrade.sh
jewzaam: Could but then have to add the various values together.  Given it's already written to handle master and worker separate I'm in favor of keeping it that way.  Other option is drop master and simply keep the worker check but it's not really critical to tweak that.  I like having a fail-safe to check master node upgrades.

mrbarge:
@jewzaam 's PR of the PR has been merged in, /lgtm

PR #302, opened 2020-04-21
OSD-2681 Disable curation for extended-dedicated-admin clusters

rogbas:


jewzaam:
well that's simple...

/lgtm

PR #309, opened 2020-04-24
Scope down security pieces for efs-csi

cblecker:
/assign @2uasimojo @jharrington22


cblecker:
@2uasimojo still something worth looking at?

jharrington22:
@2uasimojo could you make sure these are included in the operator? 

@cblecker we'll be using an operator to install these pieces in cluster so this can be closed and can be added to the operator

cblecker:
@jharrington22 sounds good!

PR #311, opened 2020-04-27
efs-csi: use `latest` CSI driver image

2uasimojo:
In testing it was discovered that aws-efs-csi-driver:release-0.3 was
built using amazon-efs-utils 1.21, which doesn't include a dependency on
openssl, and therefore doesn't support TLS, which is required in order
to mount with access points.

Requesting a new tagged build via [this
issue](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/152),
whereupon we'll revert this commit.

### DIFF
--- a/deploy/efs-csi/04-daemonset.yaml
+++ b/deploy/efs-csi/04-daemonset.yaml
@@ -30,8 +30,11 @@ spec:
         - name: efs-plugin
           securityContext:
             privileged: true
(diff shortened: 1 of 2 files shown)


2uasimojo:
Pin to 0.4 proposed via https://github.com/openshift/managed-cluster-config/pull/312 (on hold until there actually is a 0.4)

PR #313, opened 2020-04-28
Improved in-cluster uptime alert to detect gaps in requests

superfunhappytime:
Change to the in-cluster alert to mark any 5m blocks of missing metrics _after_ cluster creation as a downtime event. During api outages, sometimes it will be unable to send metrics to prometheus, resulting in a false negative.

### DIFF
--- a/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
@@ -10,12 +10,14 @@ spec:
   groups:
   - name: sre-uptime-sla
     rules:
+    # This alert is used to mark any 5m block of 100% requests failure to api 
+    # or any 5m blocks of missing metrics after cluster creation as a downtime event in SLA calculations.
+    # cluster_version is used to detect the creation time of the cluster
     - alert: SLAUptimeSRE
       annotations:
         message: The API server has had 100 percent request failures for 300 seconds.
       expr: |
-        (sum(rate(apiserver_request_total{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_total{job="apiserver", code=~"([1-5]..)"}[5m])) >= 1)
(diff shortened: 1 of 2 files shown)


cblecker:
/cc @mbarnes 
Would you mind taking a look at the PromQL here, Matt?

deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
```diff
@@ -14,7 +14,7 @@ spec:
       annotations:
         message: The API server has had 100 percent request failures for 300 seconds.
       expr: |
-        (sum(rate(apiserver_request_total{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_total{job="apiserver", code=~"([1-5]..)"}[5m])) >= 1)
+        (sum(rate(apiserver_request_total{job="apiserver", code=~"(5..|400|410|0)"}[5m])) / sum(rate(apiserver_request_total{job="apiserver"}[5m])) >= bool 1) or (sum_over_time(absent(apiserver_request_total) and (count_over_time((topk(1, time() - min(cluster_version)) > bool 0)[1d:]))[5m:5m])) >= bool 1
```
mbarnes, line 17: I had to look up what `cluster_version` was since it's not obvious it's actually a time stamp and the subtraction from `time()` looked odd at first.  Now that I understand it, the expression looks good to me (AFAICT).

However, for the benefit of others -- and especially for complex expressions like this -- can you please add a comment here explaining what this is doing?  (see [this example](https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml))

superfunhappytime: Added some comments :) Please let me know if there isn't enough detail

mbarnes:
Thanks for the update!
/lgtm

PR #322, opened 2020-05-06
Only warn if critical alerts prior to upgrade

cblecker:
ref: https://coreos.slack.com/archives/CCX9DB894/p1588788459236400?thread_ts=1588786626.228300&cid=CCX9DB894

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -53,7 +53,7 @@ log() {
     MESSAGE=$3
 
     MESSAGE=$(echo $MESSAGE | sed 's/\(.*\),$/\1/g')
-    
+
     echo "$(date "+%Y-%m-%d_%H.%M.%S") - $NAME - $STAGE - $MESSAGE"
 }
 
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -469,32 +469,12 @@ check_cluster_status() {
...
-                # get clusteroperator error if we can
-                CO_MESSAGE=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusteroperator $CO -o json 2>/dev/null | jq -r '.status.extension.lastSyncError')
-
-                if [ "$CO_MESSAGE" != "" ];
-                then
-                    log $OCM_NAME $LOG_NAME "INFO: Operator $CO lastSyncError = $CO_MESSAGE"
-                fi
-            done
-
-        fi
```
jewzaam, line 497: Looking at this again, this will not change the upgrade script.  This is going to only be triggered when there's a critical alert already active.  This is only in here to report any degraded operators when there is a critical alert to help provide context to whoever is doing the upgrade.  I'm not in favor of removing this since it's not blocking OSD upgrades.

mrbarge: Agreed - this output was actually the main way I was able to tell that something was not right with some of the upgrades, when an operator reported as degraded for an extended period of time. During the mid-upgrade process the cluster_status() call is for informational purposes only, it doesn't affect progression.

jewzaam:
Guess the question is if we update the upgrade script to ignore degraded operator alerts before starting an upgrade.  Technically it doesn't stop an upgrade from starting. But because we put cluster in maint so we don't get alerted by noisy things I'm not in favor of ignoring any alert that can page SRE.

jewzaam:
@cblecker do we want to keep this PR around or close it?

cblecker:
@jewzaam Changed this around so that `check_cluster_status` doesn't actually block the upgrade

jewzaam:
Seems reasonable.  Would like some more eyes.  I read back through the prior conversation linked and adding in latest context and it boils down to two things:
1. clusters are assumed healthy by virtue of SRE on-call fixing things
2. we'll soon allow upgrades to be initiated by customers

You have my verbal +1 but I'd like @dofinn to have some eyes on this and give a final review.

cblecker:
/assign @dofinn 
Please have a look when you can

cblecker:
Closing as no longer applicable

PR #329, opened 2020-05-11
Do not allow customers to adminster RHMI CRs

jewzaam:
https://issues.redhat.com/browse/OSD-3636

Switched from using namespace to using operator group name (label olm.owner).
Include the list of current RHMI operator group names.

Tested in Stage, works as expected.  Dedicated-admins can "get" but CANNOT "edit" `Rhmis` CRs.

### DIFF
--- a/deploy/rbac-permissions-operator-config/03-dedicated-admins-cluster.ClusterRole.yaml
+++ b/deploy/rbac-permissions-operator-config/03-dedicated-admins-cluster.ClusterRole.yaml
@@ -20,12 +20,32 @@ aggregationRule:
       operator: In
       values:
       - OperatorGroup
+    # exclude specific operator groups by name (aka owner)
+    - key: olm.owner
+      operator: NotIn
+      values:
+      # platform operator groups
+      - openshift-cluster-monitoring
+      - olm-operators
+      # OSD operator groups
+      - cloud-ingress-operator
+      - rbac-permissions-operator
+      - splunk-forwarder-operator-og
+      # Layred Product operator groups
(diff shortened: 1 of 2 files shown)


deploy/rbac-permissions-operator-config/03-dedicated-admins-cluster.ClusterRole.yaml
```diff
@@ -20,17 +20,37 @@ aggregationRule:
...
+    - key: olm.owner
+      operator: NotIn
+      values:
+      # platform operator groups
+      - openshift-cluster-monitoring
+      - olm-operators
+      # OSD operator groups
+      - cloud-ingress-operator
+      - rbac-permissions-operator
+      - splunk-forwarder-operator-og
```
rogbas, line 33: Shouldn't `openshift-velero` be listed here?

jewzaam: It's not deployed via OLM yet, is deployed by MCC.
https://github.com/openshift/managed-cluster-config/tree/master/deploy/managed-velero-operator

jewzaam: The exclusion of `openshift-velero` namespace will catch it in the future until we get an operatorgroup.

rogbas: :+1:

PR #331, opened 2020-05-11
Update the ClusterUpgradingSRE alert

cblecker:
Changes severity to info, and adds a message.

How it looks:
Screenshot 2020-05-11 15 28 45
Screenshot 2020-05-11 15 29 02

### DIFF
--- a/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
@@ -13,8 +13,7 @@ spec:
     - alert: ClusterUpgradingSRE
       expr: cluster_version{type="updating"} > 0
       labels:
-        severity: warning
+        severity: info
         namespace: openshift-monitoring
       annotations:
-        from_version: "$labels.from_version"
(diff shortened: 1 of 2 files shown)


deploy/sre-prometheus/100-cluster-upgrade.PrometheusRule.yaml
```diff
@@ -13,8 +13,7 @@ spec:
     - alert: ClusterUpgradingSRE
       expr: cluster_version{type="updating"} > 0
       labels:
-        severity: warning
+        severity: info
         namespace: openshift-monitoring
       annotations:
-        from_version: "$labels.from_version"
-        to_version: "$labels.version"
```
jewzaam, line 20: Why remove these annotations?  I assume they're available in PD and we could use it to query telemeter data, but maybe they should be labels for that?

cblecker: The labels already show up in the PD alert, but the annotations don't render. They literally show up in PD as:
```
Annotations:
 - from_version = $labels.from_version
 - to_version = $labels.version
```

PR #333, opened 2020-05-14
Remove bad SOP link from README

dustman9000:
This removes a bad link in the README pointing to a private OSD SOP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,6 @@ SelectorSyncSet:
 - Sets up a ClusterRoleBinding allowing dedicated-admins to create
   PersistentVolumes.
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
Okay, let's merge this and put up another PR with the right link. We can continue the discussion there.

/hold cancel
/approve
/lgtm

2uasimojo:
I guess my /approve isn't powerful enough (which makes sense).

https://github.com/openshift/managed-cluster-config/pull/334 fixes the link instead of removing it. If this one merges, I'll rebase to re-add it. Either way, let's continue the discussion there.

PR #334, opened 2020-05-14
Use correct link to EFS CSI ops-sop doc

2uasimojo:
The doc was ported to .md

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ SelectorSyncSet:
 - Sets up a ClusterRoleBinding allowing dedicated-admins to create
   PersistentVolumes.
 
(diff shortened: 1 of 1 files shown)


2uasimojo:
This or PR #333

2uasimojo:
/hold

So the question here is whether it's okay for this doc (which is public?) to contain a link into a private repository.

If the intended audience of this doc is SREs anyway, then the benefit of having an arrow back to the master SOP that makes this work might outweigh the inconvenience/confusion of a 403 in the rare case that someone else tries to hit it.

cblecker:
The SOP can link here.. we don't necessarily need a link the other way IMO

2uasimojo:
Okay, thanks for the input @cblecker. Closing.

PR #335, opened 2020-05-14
Initial effort at Group, ClusterRole and ClusterRoleBinding for osd-devaccess

robotmaxtron:
As part of OSD-3282/3284


deploy/osd-devaccess/01-osd-devaccess.ClusterRole.yaml
```diff
@@ -0,0 +1,90 @@
...
+  # devaccess can get, list, and watch all resources
+  - apiGroups:
+      - "*"
+    resources:
+      - "*"
+    verbs:
+      - get
+      - list
+      - watch
+  # devaccess can get and list nodes
```
cblecker, line 15: Are all these get/list entries for other objects redundant if the first rule is "get/list/watch all"?

robotmaxtron: Yep, thank you.

PR #338, opened 2020-05-19
Additional Resources Required for Metering Operator

dustman9000:
Adds additional resources required for full 'reporting-admin' permissions. This is required for exposing and configuring reports as documented here: https://github.com/kube-reporting/metering-operator/blob/master/Documentation/configuring-reporting-operator.md


dustman9000:
No longer needed as the OLM install takes care of these.

PR #343, opened 2020-05-24
Add AlertManagerSilenceSRE alert

mrbarge:
This PR implements [OSD-3494](https://issues.redhat.com/browse/OSD-3494) to add a new alert "AlertManagerSilenceSRE" which raises if active AlertManager silences are detected on the cluster.

The following exclusions apply to when this alert will fire:
- The alert will not fire if the cluster is in the process of upgrading. Alert manager silences are expected during the upgrade process via the Upgrade Operator (see [OSD-3426](https://issues.redhat.com/browse/OSD-3426))
- The alert will not fire if alertmanager is set up to silence the AlertManagerSilenceSRE alert itself. I can't think how to get around this one.

SOPs for the alert have been undertaken [as part of this ops-sop PR](https://github.com/openshift/ops-sop/pull/787) and should be reviewed alongside this PR.

### DIFF
--- a/deploy/sre-prometheus/100-alertmanager-silence-active.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-alertmanager-silence-active.PrometheusRule.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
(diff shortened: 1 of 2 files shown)


mrbarge:
> Can we please add a more verbose message annotation to this?

Thanks for the pick-up @cblecker - an annotation has now been added.

PR #126, opened 2019-10-28
[WIP] Set default service-lb quota to 0, exclude openshift-ingress

jewzaam:
https://jira.coreos.com/browse/SREP-2307

Only allows the single default value.

PR #350, opened 2020-06-01
This exclude the service loadbalancer quota from  openshift-kube-apiserver

wanghaoran1988:
It looks like we have a apischeme CR by default, and this will create a service using loadbalancer, we should exclude this.


lisa:
> @wanghaoran1988 I would expect to see changes in https://github.com/openshift/managed-cluster-config/blob/master/deploy/resource-quotas/values.mk for this. This change might not be persisted, check out the readme in https://github.com/openshift/managed-cluster-config/tree/master/deploy/resource-quotas

Nevermind. It looks like #126 broke the behaviour I quoted. (cc @jewzaam)

cblecker:
This was done in #352
/close

PR #357, opened 2020-06-02
OSD-3282 rbac

robotmaxtron:
rbac for v4 tiered access


hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -3304,6 +3304,165 @@ objects:
...
+        name: osd-devaccess
+      subjects:
+      - kind: Group
+        name: osd-devaccess
+    - apiVersion: v1
+      data:
+        ca.crt: '-----BEGIN CERTIFICATE-----
+
+          MIIENDCCAxygAwIBAgIJANunI0D662cnMA0GCSqGSIb3DQEBCwUAMIGlMQswCQYD
+
```
lisa, line 3414: This is formatted oddly? I wonder why it's like this and not the `|-` style as above?

PR #358, opened 2020-06-03
CSSRE-996: Updated RBAC for CSSRE to be able to list/get/create proce…

pepedocs:
[Jira issue](https://issues.redhat.com/browse/CSSRE-996)

Updated RBAC for CS-SRE to be able to list/create/get `processedtemplates` resources. This access is required for creating RH SSO IDPs.

Please review.

### DIFF
--- a/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
+++ b/deploy/layered-sre-authorization/01-layered-cs-sre-admin-cluster.ClusterRole.yaml
@@ -83,3 +83,13 @@ rules:
   - patch 
   - update 
   - watch
+
+# CS SRE can manage processedtemplates
+- apiGroups:
+  - "template.openshift.io"
+  resources:
+  - processedtemplates
+  verbs:
+  - create
+  - list
+  - get
--- a/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
@@ -1062,6 +1062,14 @@ objects:
         - patch
         - update
         - watch
+      - apiGroups:
+        - template.openshift.io
+        resources:
+        - processedtemplates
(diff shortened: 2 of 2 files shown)


jewzaam:
/hold

The script is adding an IDP to the cluster in the OAuth cluster CR.  Hive manages this and will rewrite identity providers if any SelectorSyncIdentityProvider or SyncIdentityProvider changes happen.  I have verified all in-cluster changes to OAuth .spec.identityProviders are lost.

pepedocs:
> /hold
> 
> The script is adding an IDP to the cluster in the OAuth cluster CR. Hive manages this and will rewrite identity providers if any SelectorSyncIdentityProvider or SyncIdentityProvider changes happen. I have verified all in-cluster changes to OAuth .spec.identityProviders are lost.

When this script uses the ocm api does this still hold true? When CLUSTER_ID var is set it will use the ocm api.

jewzaam:
> When this script uses the ocm api does this still hold true? When CLUSTER_ID var is set it will use the ocm api.

When using ocm api the config will not be overwritten, it is then managed by hive correctly.  Thank you for clarification.  There was additional commentary around the OAuth config on slack yesterday which raised this as a concern.

Regarding this PR, is the RBAC required at a cluster scope?  Much prefer if we can add this to project scope if that works in https://github.com/openshift/managed-cluster-config/blob/master/deploy/layered-sre-authorization/01-layered-cs-sre-admin-project.ClusterRole.yaml

alanmoran:
I agree with @jewzaam the scope needs to be reduced to project (customer ns, and rhmi ns). @pepedocs please move this to 01-layered-cs-sre-admin-project.ClusterRole.yaml

Now that the Oauth permissions are removed. I'm happy.

dofinn:
@pepedocs is this ok to close out?

pepedocs:
> @pepedocs is this ok to close out?

Yes I don't think we need it anymore.

PR #364, opened 2020-06-08
Add AlertManagerConfiguration CRs for OSD

grdryn:
This change requires openshift/configure-alertmanager-operator/pull/78
to have already rolled out the AlertManagerConfiguration CRD I
believe. I'm not really sure of the ordering of how everything gets
rolled out.


grdryn:
/cc @jewzaam @cblecker this is part of the change in https://github.com/openshift/configure-alertmanager-operator/pull/78

I'm not sure what's failing in the job as I can't access it. Also not sure about naming of directories or files here, or if I should have added these files to some existing directory.

grdryn:
No longer relevant, closing

PR #367, opened 2020-06-17
use of :z for bindmount

dofinn:
podman makes were failing with `/bin/sh: 1: scripts/generate_syncset.py: Permission denied`. 

Added :z as per:

```
     To  change  a  label  in  the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Podman to relabel file objects on the shared volumes. The z option
       tells Podman that two containers share the volume content. As a result, Podman labels the content with a shared content label. Shared volume labels allow all containers to read/write content.   The  Z
       option tells Podman to label the content with a private unshared label.  Only the current container can use a private volume.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ generate-oauth-templates:
 .PHONY: generate-syncset
 generate-syncset: generate-oauth-templates
 	if [ -z ${IN_CONTAINER} ]; then \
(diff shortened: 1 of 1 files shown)


dofinn:
This also also been tested with docker.

lisa:
works on Macs.

/lgtm

PR #373, opened 2020-06-22
OHSS-390

robotmaxtron:
Correcting the role aggregation rules for osd-devaccess, no extra permissions added.


sam-nguyen7:
@robotmaxtron @jewzaam any reason why this is in its own folder rather than `rbac-permissions-operator-config` ?

jewzaam:
@sam-nguyen7 separation of concerns.  The fewer things we have in a SSS the better.  Any failure to sync a resource results in subsequent resources to not be attempted.  I'd like to see us grouping only related items.  Tbh, rbac-permissions-operator-config bugs me because of the name and I put dedicated-readers in there.  Probably because of the name.  I'd rather split out dedicated-admins and dedicated-readers and not name it on the operator unless it's something specific to that operator's deployment or general configuration.. at which point it should be separate too.

PR #374, opened 2020-06-26
Add 'name' label to OLM's namespace

jewzaam:
https://issues.redhat.com/browse/OSD-4230

### DIFF
--- a/deploy/osd-namespace-labels/README.md
+++ b/deploy/osd-namespace-labels/README.md
@@ -1,0 +1,13 @@
+# Overfiew
+
+We needed in OSD-4230 to patch the `openshift-operator-lifecycle-manager` namespace to add a label that could be used in NetworkPolicy.  We don't do anything else directly with that namespace and may need a way to do similar tweaks in the future.  So this will become a single SSS to patch namespace labels for these use cases.
+
+# Guideline
+
+The general guideline for these labels are:
+- label key is `name`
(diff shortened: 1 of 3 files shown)


deploy/osd-namespace-labels/README.md
```diff
@@ -0,0 +1,13 @@
+# Overfiew
+
+We needed in OSD-4230 to patch the `openshift-operator-lifecycle-manager` namespace to add a label that could be used in NetworkPolicy.  We don't do anything else directly with that namespace and may need a way to do similar tweaks in the future.  So this will become a single SSS to patch namespace labels for these use cases.
```
lisa, line 3:
```suggestion
We needed in [OSD-4230](https://issues.redhat.com/browse/OSD-4230) to patch the `openshift-operator-lifecycle-manager` namespace to add a label that could be used in NetworkPolicy.  We don't do anything else directly with that namespace and may need a way to do similar tweaks in the future.  So this will become a single SSS to patch namespace labels for these use cases.
```

deploy/osd-namespace-labels/README.md
```diff
@@ -0,0 +1,13 @@
+# Overfiew
```
RiRa12621, line 1: Typo in “overview”

jewzaam: https://github.com/openshift/managed-cluster-config/pull/375

PR #382, opened 2020-07-03
Add pruner for removing old alpha MHC resources

cblecker:
ref OSD-3967


cblecker:
/hold
Hold pending having a cluster ready to test post-merge

PR #385, opened 2020-07-08
Added a Prometheus rule to fire an info level alert when console routes are unavailable

sulmer-redhat:
@RiRa12621 Is that the correct implementation? 
In this rule, all "response error total" metrics bigger than 0 will be firing an alert.
Can we later use this to calculate SLO in the telemeter dashboard?

### DIFF
--- a/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
@@ -18,7 +18,7 @@ spec:
         severity: info
         namespace: openshift-monitoring
 
-    - alert: API-Route-SLO-Breach
+    - alert: OpenShift-API-Server-Route-SLO-Breach
       expr: haproxy_server_response_errors_total{exported_service=~"(console)"} > 0
       for: 5m
       labels:
--- a/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
@@ -12,15 +12,11 @@ spec:
   - name: sre-important-routes-slo-alerts
     rules:
     - alert: ConsoleRoute-SLO-Breach
-      expr: haproxy_server_response_errors_total{exported_service=~"(oauth-openshift)"} > 0
-      for: 5m
-      labels:
-        severity: info
-        namespace: openshift-monitoring
-
-    - alert: OpenShift-API-Server-Route-SLO-Breach
-      expr: haproxy_server_response_errors_total{exported_service=~"(console)"} > 0
-      for: 5m
+      annotations:
+        message: The console route has had 100 percent request failures for 60 seconds.
+      expr: |
+        (sum(rate(haproxy_backend_http_responses_total{route=~"console",code=~"(5..)"}[1m])) / sum(rate(haproxy_backend_http_responses_total{route=~"console",code=~"([1-5]..)"}[1m])) >= 1)
+      for: 1m
       labels:
         severity: info
         namespace: openshift-monitoring
--- a/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
@@ -11,7 +11,7 @@ spec:
   groups:
   - name: sre-important-routes-slo-alerts
     rules:
-    - alert: ConsoleRoute-SLO-Breach
+    - alert: ConsoleRequestErrorsSRE
       annotations:
         message: The console route has had 100 percent request failures for 60 seconds.
       expr: |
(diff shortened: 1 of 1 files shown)


sulmer-redhat:
[WIP]

deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: sre-important-routes-slo-alerts
+    rules:
+    - alert: ConsoleRoute-SLO-Breach
+      expr: haproxy_server_response_errors_total{exported_service=~"(oauth-openshift)"} > 0
+      for: 5m
+      labels:
+        severity: info
+        namespace: openshift-monitoring
+
+    - alert: API-Route-SLO-Breach
```
RiRa12621, line 21:
```suggestion
- alert: OpenShift-API-Server-Route-SLO-Breach
```

RiRa12621:
/hold
I think they should both be a rate so that you can have the errors/s and then the alert should fire if it's > x percent of total requests/s to that route

then your you'd basically fire if x % of requests per second are errors

sulmer-redhat:
/hold

The only example I have is from 100-sre-api-errors.PrometheusRule.yaml:

_"        (sum(rate(apiserver_request_total{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_total{job="apiserver", code=~"([1-5]..)"}[5m])) >= 1) "_

I tried to craft something similar:

_"        sum(rate((haproxy_server_response_errors_total{exported_service=~"(oauth-openshift|console)"})[5m:1s])) / sum(rate((haproxy_server_http_responses_total{exported_service=~"(oauth-openshift|console)"})[5m:1s])) "_

Do you think that can be enough? I am not sure if my math is correct, the _haproxy_server_response_errors_total_ and _haproxy_server_http_responses_total_ are not the same as filtering individual http return codes with the _apiserver_request_total_ parameter, as in the example above.

sulmer-redhat:
@RiRa12621 PTAL

deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
```diff
@@ -0,0 +1,22 @@
...
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-important-routes-slo-alerts
+    rules:
+    - alert: ConsoleRoute-SLO-Breach
+      annotations:
+        message: The console route has had 100 percent request failures for 60 seconds.
+      expr: |
+        (sum(rate(haproxy_backend_http_responses_total{route=~"console",code=~"(5..)"}[1m])) / sum(rate(haproxy_backend_http_responses_total{route=~"console",code=~"([1-5]..)"}[1m])) >= 1)
```
RiRa12621, line 18: why the extra parenthesis at the beginning?

sulmer-redhat: [Looks the same here](https://github.com/openshift/managed-cluster-config/blob/ecc37336cecb461e78cc111eea429fa53d5d4005/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml#L17) so I thought to save the same syntax.

RiRa12621: can you make sure that's really needed and drop them if they're not please

sulmer-redhat: done :)

deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
```diff
@@ -0,0 +1,22 @@
+
```
cblecker, line 1:
```suggestion
---
```

deploy/sre-prometheus/100-important-routes-slo-PrometheusRule.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    prometheus: sre-important-routes-slo
+    role: alert-rules
+  name: sre-important-routes-slo
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-important-routes-slo-alerts
+    rules:
+    - alert: ConsoleRoute-SLO-Breach
```
cblecker, line 14:
```suggestion
    - alert: ConsoleRequestErrorsSRE
```

sulmer-redhat:
Anything else or we can merge?

jeremyeder:
Is there a corresponding "upstream" commit for this?  Upstream being OCP itself.

RiRa12621:
Not at this point.
Mainly because it's not very straight forward where it needs to go.
Technically we're getting extra information about how the console performs on the edge (as far as we can measure), so from that perspective I'd say it should be part of the console.
On the other hand it's measured on the router so that would point for it to be part of the router.

Also I would like this to get a review from the rest of WG-MSO before it's proposed against an OCP component

cblecker:
@sulmer-redhat Requires a `make` to rebuild the applied selectorsyncset

RiRa12621:
@sulmer-redhat are you following up on this one?

sulmer-redhat:
> @sulmer-redhat are you following up on this one?
> from my perspective, we probably want to be closing this, in turn define an SLO for the console and then start alerting on it accordingly

we're going to use the new operator so yes, this PR is not relevant anymore.
closing

PR #388, opened 2020-07-09
Adds system:admin impersonation

iamkirkbater:
This PR aims to add the ability for SREs and customers alike to impersonate system:admin rather than to always have cluster-admin permissions enabled.  OSD-4234

This change gives SREs the ability to escalate commands with impersonation and removes the sre-cluster-admin group, as well as creating the impersonator group for the customers.

Escalation would look something like this:

`oc apply -f myClusterRoleBinding.yaml --as=system:admin`


cblecker:
/hold
Holding for discussion. This will change a number of SRE workflows.

iamkirkbater:
Closing due to #390 being merged in lieu of this one.

PR #389, opened 2020-07-10
Minor updates to the AWS credentials rotation script

arjunrn:
The script to rotate the AWS credentials has been updated with the following changes:

1. pods which mount the generated secrets are deleted in parallel but the script waits for them to complete. This is to prevent the script from exiting before the deletions have completed.
2. The names of the AWS users are now generated and have a randomized suffix. The script now takes that into account.

### DIFF
--- a/scripts/rotate-aws-creds.sh
+++ b/scripts/rotate-aws-creds.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -euo pipefail
+set -x
 
 ENVIRONMENT=$1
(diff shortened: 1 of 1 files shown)


2uasimojo:
> 1. pods which mount the generated secrets are deleted in parallel but the script waits for them to complete. This is to prevent the script from exiting before the deletions have completed.

We could use the bash `wait` primitive instead, and still get the advantage of the parallelism.

PR #390, opened 2020-07-13
Osd 4234

robotmaxtron:
https://issues.redhat.com/browse/OSD-4234

Introduced a new impersonators group, and enabling the 'sudoers' ClusterRole with a ClusterRoleBinding. This is designed to only allow existing ClusterAdmin clusters, the ability to leverage impersonation of the system:admin role without required changes to existing workflows.

### DIFF
--- a/deploy/osd-cluster-admin/06-osd-impersonators.Group.yaml
+++ b/deploy/osd-cluster-admin/06-osd-impersonators.Group.yaml
@@ -1,0 +1,4 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: osd-impersonators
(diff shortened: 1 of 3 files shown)


lisa:
/hold

for review

Do we need a webhook to protect the group?

deploy/osd-cluster-admin/06-osd-impersonators.Group.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: osd-impersonators
```
cblecker, line 4: Do we want to prefix this group, if we want the customer to be able to edit it? Will this group's membership be erased by the SelectorSyncSet?

robotmaxtron: No, the group membership will not be reset automatically. I'm not sure that we _want_ the customers to edit it, but with impersonation they will have access. The specifics on how the workflow will interface with OCM hasn't been fully realized so I'm hesitant to default to an unprotected namespace.

PR #391, opened 2020-07-13
Add osd-sre-admins group per OSD-4455

csheremeta:
https://issues.redhat.com/browse/OSD-4455

See also https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/6439


hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -4722,8 +4727,8 @@ objects:
     - apiVersion: user.openshift.io/v1
       kind: Group
       metadata:
-        name: osd-sre-cluster-admins
-      users: []
+        name: osd-sre-admins
+      users: ${{SREP_USERS}}
```
jewzaam, line 4731: `SREP_USERS` needs to be defined in the template.  Similar to this:
https://github.com/openshift/managed-cluster-config/blob/master/scripts/templates/template.yaml#L31-L32

csheremeta: thanks for catching this! I've added it in.

jewzaam:
`SREP_USERS` is created and ready for this PR.

/approve
/lgtm

PR #392, opened 2020-07-14
OSD-4493 Pre-update ClusterVersion.spec.channel with the desired channel if necessary

rogbas:
Update the ClusterVersion.spec.channel in case of a cluster upgrade to a different minor version (e.g 4.3 to 4.4).

If the channel is already the desired one, don't do anything.

### DIFF
--- a/scripts/cluster-upgrade.sh
+++ b/scripts/cluster-upgrade.sh
@@ -249,6 +249,19 @@ upgrade() {
         return
     fi
 
+    # Bump current ClusterVersion Channel in case it's pointing to a previous minor version
+    DESIRED_CHANNEL=$(get_channel $TO)
+    CURRENT_CHANNEL=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusterversion version -o json | jq ".spec.channel")
+
+    # If the channel is already set properly, don't do anything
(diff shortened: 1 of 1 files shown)


scripts/cluster-upgrade.sh
```diff
@@ -249,6 +249,19 @@ upgrade() {
...
+    # Bump current ClusterVersion Channel in case it's pointing to a previous minor version
+    DESIRED_CHANNEL=$(get_channel $TO)
+    CURRENT_CHANNEL=$(KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc get clusterversion version -o json | jq ".spec.channel")
+
+    # If the channel is already set properly, don't do anything
+    if [ "$DESIRED_CHANNEL" != "$CURRENT_CHANNEL" ];
+    then
+        KUBECONFIG=$TMP_DIR/kubeconfig-${CD_NAMESPACE} oc patch clusterversion version --type merge -p "{\"spec\":{\"channel\": \"$DESIRED_CHANNEL\"}}"
+
+        # Wait for CVO to identiofy the update and refresh clusterVersion.status.availableUpdates
```
cblecker, line 261:
```suggestion
        # Wait for CVO to identify the update and refresh clusterVersion.status.availableUpdates
```

PR #395, opened 2020-07-20
define the alert rules for the upgrade operator

bmeng:
Create the prometheus alert rules for managed upgrade operator based on the generated metrics.


deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,77 @@
...
+  labels:
+    prometheus: sre-managed-upgrade-operator-alerts
+    role: alert-rules
+  name: sre-managed-upgrade-operator-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-managed-upgrade-operator-alerts
+    rules:
+    - alert: UpgradeConfigValidationFailed
```
mrbarge, line 13: We seem to have an existing naming convention of appending "SRE" to our alerts, I guess to distinguish them from built-in k8s/OCP alerts. Should we do similar for these?

On a related note, what do you think about establishing a common naming convention so that they're easier to allow-filter in the silences we create? ie. making our silence allow anything matching a regex of "Upgrade.*SRE". cc @sedroche 

```suggestion
    - alert: UpgradeConfigValidationFailedSRE
```

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,77 @@
...
+      expr: |
+        upgradeoperator_upgradeconfig_validation_failed > 0
+      for: 5m
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+      annotations:
+        summary: "Upgrade config validation failed"
+        description: "Upgrade config validation failed"
+    - alert: ClusterCheckFailed
```
mrbarge, line 24: This alert name sounds a bit generic, maybe 'UpgradeClusterCheckFailedSRE' or something to give it more context?

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,77 @@
...
+      expr: |
+        upgradeoperator_cluster_check_failed > 0
+      for: 5m
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+      annotations:
+        summary: "cluster check failed"
+        description: "basic cluster checks failed on either before the upgrade or after the upgrade"
+    - alert: NodeScalingFailed
```
mrbarge, line 35: Again maybe some additional context to the alert name here to indicate it's scaling due to a cluster upgrade. eg "UpgradeNodeScalingFailedSRE"

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,77 @@
...
+      expr: |
+        upgradeoperator_scaling_failed > 0
+      for: 5m
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+      annotations:
+        summary: "node scaling failed"
+        description: "The extra machine/node was not ready before the upgrade started"
+    - alert: ClusterVerificationFailed
```
mrbarge, line 46: Same comments as above, it sounds a bit generic as-is. "UpgradePostVerificationFailureSRE" or something maybe?

mrbarge:
@bmeng Could we also add an additional alert for breaching the upgrade window? I've written up what I think it should look like:
https://gist.github.com/mrbarge/dde03418562f65cdc0bbff9c54b66ed0

hack/00-osd-managed-cluster-config.selectorsyncset.yaml.tmpl
```diff
@@ -4906,6 +4906,101 @@ objects:
...
+          role: alert-rules
+        name: sre-managed-upgrade-operator-alerts
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-managed-upgrade-operator-alerts
+          rules:
+          - alert: UpgradeConfigValidationFailedSRE
+            expr: 'upgradeoperator_upgradeconfig_validation_failed > 0
+
```
wanghaoran1988, line 4923: Why here is an empty line?

bmeng: caused by the format of the expr.
If we write it as 
```
expr: |
 xxxx
```
it will add the line break in the end. 
I can see the format is supported in the cluster.
Fixed as we do not need the expression to be multiple lines.

mrbarge:
@bmeng we should remove the `deploy/sre-prometheus/100-cluster-ugprade.PrometheusRule.yaml` alert as part of this PR too, what do you think?

cblecker:
/hold

Is the upgrade operator in production?

bmeng:
Yes, it is not in production yet. But will be soon :smile:

cblecker:
/hold cancel

Haven't reviewed this, but now that it's in prod, removing the hold I put on earlier

PR #396, opened 2020-07-20
Update rotate aws credentials script

arjunrn:
The script currently does not handle CCS/MOA clusters where the names of the managed admin users have a suffix. The script has been updated to handle this case.


scripts/rotate-aws-creds.sh
```diff
@@ -91,17 +115,32 @@ update_aws_secret() {
 }
 
 NAMESPACE=$(ssh root@hive-$ENVIRONMENT-master "oc get clusterdeployment --all-namespaces --no-headers | grep \"$CLUSTER_ID \" | awk '{print \$1}'")
+CLUSTER_NAME=$(ssh root@hive-$ENVIRONMENT-master "oc get clusterdeployment --all-namespaces --no-headers | grep \"$CLUSTER_ID \" | awk '{print \$2}'")
 
-ACCESS_KEY_ID=$(ssh root@hive-$ENVIRONMENT-master "oc get secrets -n $NAMESPACE aws -o json | jq -r '.data.aws_access_key_id' | base64 --decode")
-SECRET_ACCESS_KEY=$(ssh root@hive-$ENVIRONMENT-master "oc get secrets -n $NAMESPACE aws -o json | jq -r '.data.aws_secret_access_key' | base64 --decode")
+if [ $(ssh root@hive-production-master "oc get clusterdeployment -n ${NAMESPACE} ${CLUSTER_NAME} -o yaml -o json" | jq -r '.metadata.labels["api.openshift.com/ccs"]') == "true" ]; then
```
jewzaam, line 120:
```suggestion
if [ $(ssh root@hive-$ENVIRONMENT-master "oc get clusterdeployment -n ${NAMESPACE} ${CLUSTER_NAME} -o yaml -o json" | jq -r '.metadata.labels["api.openshift.com/ccs"]') == "true" ]; then
```

arjunrn:
@jewzaam My original task was to write an SOP on how to run the script. Midway through writing it I found out that the hives are going to be sharded. I didn't update the script to handle this. We need to consider how to move forward with this script anyway because it is quite logic heavy and bash may not be the right choice. So I've created a ticket to rewrite the script which would also take care of the dependency on the CLI.

scripts/rotate-aws-creds/common.sh
```diff
@@ -0,0 +1,132 @@
...
+            S=$(jq -r '.spec.secretRef.name' "$F")
+
+            # cleanup any pods that referenced the secret (note we're ignoring namespace, exceptionally small risk)
+            IFS_OLD=$IFS
+            IFS=
+            for X in $(KUBECONFIG=$KUBECONFIG $OC get pods --all-namespaces -o json | jq -r ".items[] | select(.spec.volumes[].secret.secretName == \"$S\") | \"$OC -n \" + .metadata.namespace + \" delete pod \" + .metadata.name");
+            do
+                X="KUBECONFIG=$KUBECONFIG $X"
+                # do not wait, just move on
+                eval "$X" &
```
jharrington22, line 48: We're ignoring if they fail?

scripts/rotate-aws-creds/common.sh
```diff
@@ -0,0 +1,132 @@
...
+            done
+            IFS=$IFS_OLD
+
+            info "Rotated SRE managed CredentialsRequest '$CR' in '$NS'."
```
jharrington22, line 52: You're not expecting `$CR` and `$NS` to evaluate here?

scripts/rotate-aws-creds/common.sh
```diff
@@ -0,0 +1,132 @@
...
+
+iam_user_create_accesskey() {
+    local AWS_USERNAME=$1
+
+    info "IAM User: $AWS_USERNAME"
+
+    local ORIG_ACCOUNTID
+    ORIG_ACCOUNTID=$(AWS_PROFILE=$AWS_PROFILE aws sts get-caller-identity | jq -r '.Account')
+
+    info "Account ID: $ORIG_ACCOUNTID"
```
jharrington22, line 65: Not sure if you get two new lines here but `jq -j` will print the raw output without the trailing newline.

scripts/rotate-aws-creds/common.sh
```diff
@@ -0,0 +1,132 @@
...
+    # get encoded access key data
+    NEW_ACCESS_KEY_ID=$(echo "$ACCESS_KEY_JSON" | jq -r '.AccessKey.AccessKeyId' | tr -d '\n')
+    ENCODED_ACCESS_KEY_ID=$(echo -n "$NEW_ACCESS_KEY_ID" | base64)
+    NEW_SECRET_ACCESS_KEY=$(echo "$ACCESS_KEY_JSON" | jq -r '.AccessKey.SecretAccessKey' | tr -d '\n')
+    ENCODED_SECRET_ACCESS_KEY=$(echo -n "$NEW_SECRET_ACCESS_KEY" | base64)
+
+    info "Created new AccessKey for IAM User '$AWS_USERNAME'."
+
+    # verify it works (need to sleep a bit)
+    echo "Waiting 10 seconds for new AccessKey to activate..."
```
jharrington22, line 80: I hate AWS' eventual consistency. This was a _big_ issue with the aws-account-operator

scripts/rotate-aws-creds/run.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -euo pipefail
+
+ENVIRONMENT=${1:-}
+CLUSTER_NAME=${2:-}
```
jharrington22, line 5: :+1:

scripts/rotate-aws-creds/sharded.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+set -euo pipefail
+
+source "$(dirname $0)/common.sh"
+
+SHARD_SSH="ssh root@hive-production-master"
+OC_CMD="oc --insecure-skip-tls-verify=true"
+
```
jharrington22, line 8: Should these go to common?

jharrington22: Also `SHARD_SSH` should be `SHARD_PROD_SSH` do these script support multiple ENV's?

scripts/rotate-aws-creds/sharded.sh
```diff
@@ -0,0 +1,134 @@
...
+CCS_ANNOTATION=$(KUBECONFIG="${SHARD_KUBECONFIG}" ${OC_CMD} get clusterdeployment -n "$CLUSTER_NAMESPACE" "$CLUSTER_NAME" -o json | jq -r '.metadata.labels["api.openshift.com/ccs"]')
+if [[ "$CCS_ANNOTATION" == "true" ]]; then
+    CCS_CLUSTER="1"
+    CREDENTIALS_SECRET_NAME="byoc"
+else
+    CCS_CLUSTER="0"
+    CREDENTIALS_SECRET_NAME="aws"
+fi
+
+ACCESS_KEY_ID=$(KUBECONFIG="$SHARD_KUBECONFIG" $OC_CMD get secrets -n "$CLUSTER_NAMESPACE" $CREDENTIALS_SECRET_NAME -o json | jq -r '.data.aws_access_key_id' | base64 --decode)
```
jharrington22, line 61: nit `base64 -d` to be consistent.

scripts/rotate-aws-creds/sharded.sh
```diff
@@ -0,0 +1,134 @@
...
+### Changes on the hive cluster
+
+# Replace aws-account-operator secret
+SECRET_NAME=$(KUBECONFIG=$SHARD_KUBECONFIG $OC_CMD -n "$CLUSTER_NAMESPACE" get accountclaim -o jsonpath='{.items[].spec.accountLink}')-secret
+update_aws_secret "$SHARD_KUBECONFIG" aws-account-operator "$SECRET_NAME"
+
+info "Replaced AWS Account Operator Secret."
+
+# Replace aws secret (in cluster's hive namespace)
+update_aws_secret "$SHARD_KUBECONFIG" "$CLUSTER_NAMESPACE" $CREDENTIALS_SECRET_NAME
```
jharrington22, line 87: All of these should be quoted?

jharrington22:
@arjunrn my only other comment here is to be consistent using `${}` and quote all vars since its bash.

scripts/rotate-aws-creds/common.sh
2uasimojo: They'll evaluate. The single quotes are literal since the whole thing is surrounded by double quotes.

```shell
[efried@efried ops-sop]$ a=foo
[efried@efried ops-sop]$ echo "this is '$a'."
this is 'foo'.
```

scripts/rotate-aws-creds/common.sh
arjunrn: This is how it was in the original script.

scripts/rotate-aws-creds/common.sh
arjunrn: It does not because it probably only adds the newline when printing to the terminal.
```bash
TEST=$(echo '{"name": "Arjun"}' | jq -j '.name' ); echo -n $TEST
```

scripts/rotate-aws-creds/sharded.sh
arjunrn: They support the both staging and production because the production hive(`hive-production`) is used to create both the staging and production sharded hives.

arjunrn: Also these variable are only used when accessing sharded hives so they should stay here IMO.

scripts/rotate-aws-creds/run.sh
```diff
@@ -0,0 +1,23 @@
...
+if [[ ${ENVIRONMENT} == "" || ${CLUSTER_NAME} == "" ]]; then
+    echo "Usage: $0 <environment> <cluster name> "
+    exit 1
+fi
+
+CLUSTER_ID=$(ocm list clusters | grep "${CLUSTER_NAME}" | awk '{print $1}')
```
jewzaam, line 11: Cluster names are not unique, let's use ID as the argument.  Also use `--managed` argument to list only OSD clusters.  We can switch arg 2 to be CLUSTER_ID and support both internal ID and EXTERNAL_ID with:
```suggestion
CLUSTER_ID=$(ocm list clusters --managed --columns id,external_id | grep "${CLUSTER_ID}" | awk '{print $1}')
```

scripts/rotate-aws-creds/sharded.sh
```diff
@@ -0,0 +1,134 @@
...
+
+CLUSTER_NAMESPACE=$(KUBECONFIG=${SHARD_KUBECONFIG} ${OC_CMD} get accountclaims --all-namespaces | grep "${CLUSTER_ID}" | awk '{print $1}')
+CLUSTER_KUBECONFIG_SECRET=$(KUBECONFIG=${SHARD_KUBECONFIG} ${OC_CMD} get secret -n "${CLUSTER_NAMESPACE}" | grep kubeconfig | awk '{print $1}')
+KUBECONFIG="${SHARD_KUBECONFIG}" ${OC_CMD} get secret -n "${CLUSTER_NAMESPACE}" "${CLUSTER_KUBECONFIG_SECRET}" -o json | jq -r '.data["raw-kubeconfig"]' | base64 -d >"${CLUSTER_KUBECONFIG}"
+fixup_kubeconfig "${CLUSTER_KUBECONFIG}"
+
+CCS_ANNOTATION=$(KUBECONFIG="${SHARD_KUBECONFIG}" ${OC_CMD} get clusterdeployment -n "${CLUSTER_NAMESPACE}" "${CLUSTER_NAME}" -o json | jq -r '.metadata.labels["api.openshift.com/ccs"]')
+if [[ "${CCS_ANNOTATION}" == "true" ]]; then
+    CCS_CLUSTER="1"
+    CREDENTIALS_SECRET_NAME="byoc"
```
jewzaam, line 55: There is also an `aws` secret for CCS.  Do we need to rotate it?

arjunrn: When I first implemented this script the secret had a different name. Since then things might have changed. Maybe @fahlmant might know more.

scripts/rotate-aws-creds.sh
```diff
@@ -1,216 +0,0 @@
...
-export AWS_DEFAULT_REGION=us-east-1
-
-if [ "$CLUSTER_ID" == "" ] || [ "$ENVIRONMENT" == "" ];
-then
-    echo "usage: $0 <environment> <cluster id>"
-    echo "example: $0 stage 18kshg9idnn687p1bstlh669hlod7ldr"
-    exit 1
-fi
-
-info() {
```
clcollins, line 21: This function is called frequently in the script, but no longer exists, so it executes the builtin `info` command rather than printing a log message.

scripts/rotate-aws-creds/sharded.sh
```diff
@@ -0,0 +1,134 @@
...
+update_aws_secret() {
+    local KUBECONFIG=$1
+    local SECRET_NAMESPACE=$2
+    local SECRET_NAME=$3
+    # assumes encoded vars are set already by iam_user_create_accesskey
+
+    KUBECONFIG=${KUBECONFIG} ${OC_CMD} get secret -n "${SECRET_NAMESPACE}" "${SECRET_NAME}" -o json | jq -r ".data.aws_access_key_id = \"${ENCODED_ACCESS_KEY_ID}\" | .data.aws_secret_access_key = \"${ENCODED_SECRET_ACCESS_KEY}\"" | KUBECONFIG=${KUBECONFIG} ${OC_CMD} replace -f -
+}
+
+CLUSTER_NAME=${1:-}
```
clcollins, line 28: CLUSTER_ID=${1:-}
SHARD_NAME=${2:-}
if [[ ${CLUSTER_ID} == "" || ${SHARD_NAME} == "" ]]; then
    echo "Usage: $0 CLUSTER_NAME SHARD_NAME"
    exit 1
fi

CLUSTER_NAME=$(ocm list clusters --managed | grep "${CLUSTER_ID}" | awk '{print $2}')

arjunrn:
@clcollins This PR is now quite outdated and should be updated to use the latest procedures and tools. If this is a blocker, please disregard the PR.

jharrington22:
@clcollins @arjunrn @jewzaam lets merge this PR so that is uses the break glass procedure as we don't expect this script to be used everyday. Once merged we can submit another PR to update the script to use osdctl credentials.

jharrington22:
Thanks @jewzaam 

@arjunrn could you please update the PR with relation to @clcollins's comments. As for the "byoc" secret please make sure you update the script now to also rotate the credentials in the "aws" secret and we can remove the "byoc" piece in a separate PR when [this card](https://issues.redhat.com/browse/SDA-2976) is resovled.

scripts/rotate-aws-creds/sharded.sh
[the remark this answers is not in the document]
arjunrn: Fixed.

scripts/rotate-aws-creds.sh
arjunrn: Fixed

clcollins (approved):
This looks good to me.  I haven't tested the changes, but it looks like it addresses my points.  I think it will accomplish what we need.

/lgtm

PR #398, opened 2020-07-20
Revert "Change velero alerts to info"

cblecker:
This reverts commit ca0704878f817081de6cc4666c006b3467eafbb2.

As https://github.com/openshift/managed-velero-operator/pull/87 pulls in the long term fix for https://issues.redhat.com/browse/OSD-3656, reverting these alerts back to warning, so that we can identify when we have issues with our backups


lisa:
See #371 

/lgtm

PR #401, opened 2020-07-21
Remove alert AlertmanagerSilencesActiveSRE

yeya24:

https://issues.redhat.com/browse/OSD-4432

This alert prevents us from silencing anything to the cluster.


RiRa12621:
/unhold
this was discussed in a meeting on monday with @jharrington22 @jewzaam and @mwoodson and needs to be dropped as our general agreement is move to silences to "on cluster" in the future.
This alerting rule would prevent such workflow.
/lgtm

yeya24:
@NautiluX @jewzaam Hello, would you mind taking a look at this issue?

jewzaam:
@RiRa12621 @cblecker we discussed this a few weeks ago, I didn't realize at the time we do suppress this alert while in upgrade.  We do need to create silences in cluster but should only need that for upgrades.  Would SRE not use PD maint for anything outside of an upgrade?

Note that the metric used for detecting upgrade `cluster_version{type="upgrading"}` will only cover control plane, not node upgrades.  @dofinn 

Given this, I think it's reasonable to keep the alert.  We do not want customers creating silences and SRE has other tools to create ad-hoc silences in a central system (PagerDuty).

RiRa12621:
There's two things here: one I remember seeing a card that said to move all silences to in cluster and two how would we silence a single alert for a cluster
A potential case here is a flapping alert a customer is causing but is informed about and working on.
We could easily silence that in cluster but I'm not sure how to in PagerDuty

RiRa12621:
/close
https://github.com/openshift/managed-cluster-config/pull/401#issuecomment-665171172
Clearly no perspective of this ever getting merged

PR #402, opened 2020-07-21
Remove ExcessiveContainerMemory alerts

yeya24:

https://issues.redhat.com/browse/OSD-4439

The alerts `ExcessiveContainerMemoryWarningSRE` and `ExcessiveContainerMemoryCriticalSRE` have hardcoded limits which are not good for alerting.


cblecker:
These alerts were added for a reason as well.

cc @jewzaam

RiRa12621:
iirc the reason was a memory leak in one of the OCP components @cblecker , which should be fixed by now.

RiRa12621:
@cblecker can we move ahead with this or not?

NautiluX:
@cblecker I don't think having an alert for a static amount of memory used is a good long-term solution to detect memory leaks. Should we create a card to ensure every pod we run in the watched namespaces have limits set before we remove the alert?

RiRa12621:
@cblecker can we move ahead with this or not?

RiRa12621:
Yes, no, maybe?

PR #403, opened 2020-07-22
feat(sre-build-test): add permissions

georgettica:
add to the ServiceAccount more permissions to do auto cleanup

related to https://issues.redhat.com/browse/OSD-4540

this can also fit the deletion as requested in PR

### DIFF
--- a/deploy/sre-build-test/50-build-test.CronJob.yaml
+++ b/deploy/sre-build-test/50-build-test.CronJob.yaml
@@ -34,10 +34,14 @@ spec:
             image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
             imagePullPolicy: Always
             env:
-              - name: JOB_NAME
+              - name: POD_NAME
                 valueFrom:
                   fieldRef:
                     fieldPath: metadata.name
+              - name: POD_NS
+                valueFrom:
+                  fieldRef:
+                    fieldPath: metadata.namespace
             command:
             - /bin/bash
             - -c
@@ -55,15 +59,18 @@ spec:
               trap "cleanup" EXIT SIGINT
 
               # set NS to include job name, to ease linking namespace to a specific job
-              export NS="openshift-build-test-${JOB_NAME##sre-build-test-}"
+              JOB_PREFIX=sre-build-test
+              NS="${POD_NS}-${POD_NAME##${JOB_PREFIX}-}"
+              JOB_NAME=$(echo "${POD_NAME}" | rev | cut -d- -f2- | rev)
+
               oc create ns "${NS}"
 
               # run build
-              oc -n "${NS}" new-build nodejs~https://github.com/openshift/nodejs-ex --name=sre-build-test
+              oc -n "${NS}" new-build nodejs~https://github.com/openshift/nodejs-ex --name="${JOB_PREFIX}"
               echo "$(date): Waiting for build to complete."
               while :
               do
-                export ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers)
+                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers)
                 case ${ST} in
                   "")
                     # if build status is blank, assume we are still starting the build
@@ -79,6 +86,11 @@ spec:
                     ;;
                   Complete)
                     echo "$(date): Build Complete"
+                    # Get all job names that have exceeded failed retries
+                    JOBS_TO_DELETE=$(oc -n "${POD_NS}" get job -o \
+                                       jsonpath='{.items[?(@.status.failed==@.spec.backoffLimit)].metadata.name}' || true)
+                    echo "$(date): Selected jobs for deletion: ${JOBS_TO_DELETE}"
+                    echo "${JOBS_TO_DELETE}" | xargs oc -n "${POD_NS}" delete job
                     break
                     ;;
                 esac
--- a/deploy/sre-build-test/40-build-test-RoleBinding.yaml
+++ b/deploy/sre-build-test/40-build-test-RoleBinding.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: sre-build-test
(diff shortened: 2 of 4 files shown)


georgettica:
as we need to change the PR for doing more, holding for now

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -55,15 +59,17 @@ spec:
...
               oc create ns "${NS}"
 
               # run build
-              oc -n "${NS}" new-build nodejs~https://github.com/openshift/nodejs-ex --name=sre-build-test
+              oc -n "${NS}" new-build nodejs~https://github.com/openshift/nodejs-ex --name="${CRONJOB_NAME}"
               echo "$(date): Waiting for build to complete."
               while :
               do
-                export ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers)
+                # fixes shellcheck error SC2155
```
NautiluX, line 71:
```suggestion
```

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -79,6 +85,12 @@ spec:
                     ;;
                   Complete)
                     echo "$(date): Build Complete"
+                     oc -n "${JOB_NS}" get job --no-headers | \
```
NautiluX, line 88: if this fails, the build fails. I this the right behaviour?

deploy/sre-build-test/50-build-test.CronJob.yaml
NautiluX: do we actually need to export ST?

deploy/sre-build-test/50-build-test.CronJob.yaml
georgettica: not sure, but will add an `||true to it`

deploy/sre-build-test/50-build-test.CronJob.yaml
georgettica: I don't think any of them need to be exported

deploy/sre-build-test/50-build-test.CronJob.yaml
georgettica: done

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -79,6 +86,12 @@ spec:
                     ;;
                   Complete)
                     echo "$(date): Build Complete"
+                    JOBS_TO_DELETE=$(oc -n "${POD_NS}" get job --no-headers || true | \
```
cblecker, line 89: Should these deletion steps go in the cleanup trap instead?

georgettica: that could happen but my bash-foo is not that string and my filtering do not work inside the trap 😊 , maybe I will be able to solve it later

georgettica: Actually @cblecker, we want this to happed only on success, so putting it in trap is counter intuative

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -79,6 +86,12 @@ spec:
...
+                                                              grep -v "${JOB_NAME}" | \
+                                                              grep -e "${JOB_PREFIX}" -e '0/1' | \
```
jewzaam, line 91: This will delete all jobs that match the prefix regardless of status.  Ideally we parse the json to be specific.  But a quick update will match what I think the expectations are, clear the failed jobs so the alerts go away:

```suggestion
                                                              grep "${JOB_PREFIX}" | \
                                                              grep '0/1' | \
```

jewzaam, line 91: Intent based on slack conversation is to clear the alert.  To do that we need to delete only the failed jobs.  This line will match any job that has failed (assuming "0/1" is a failure) AND any job with the prefix.  I.e. it will delete all the jobs regardless of status.

Recommend this instead:

```bash
JOBS_TO_DELETE=$(oc -n "${POD_NS}" get jobs -o json | jq -r ".items[] | select(.metadata.name | contains(\"$JOB_PREFIX\")) | select(.status.succeeded != 1) | .metadata.name")
```

jewzaam: Hmm. If we have `jq` available..

Since we only care about failed jobs and the job this is running in is _not_ failed, could also do:
```bash
JOBS_TO_DELETE$(oc -n "${POD_NS}" get jobs -o jsonpath='{.items[?(@.status.failed)].metadata.name}')
```

georgettica: I can add this, it sounds fair, also tested and the latter command works (no `jq` on this container)

georgettica: did it @jewzaam, please review to your satisfaction

deploy/sre-build-test/50-build-test.CronJob.yaml
georgettica: resolving according to @NautiluX response on slack

jewzaam:
/lgtm

@cblecker you put in the latest hold, wdyt?

cblecker:
I'm good with testing this.

@georgettica Please remove the hold when you are ready to be able to test

georgettica:
> I'm good with testing this.
> 
> @georgettica Please remove the hold when you are ready to be able to test

tested and now it does as expected, when the job is ready to terminate (after the failed amount threashold) it will be collected by the script and sent to deletion

@jewzaam I would like another /lgtm when you are ready

lisa:
/hold

can you squash the commits into 1?

georgettica:
sure, but I will need another lgtm after :/

lisa:
@georgettica happy to /lgtm a squashed commit.

georgettica:
@lisa

lisa:
/lgtm

/hold cancel

:shipit:

PR #406, opened 2020-07-23
Remove alert DNSLatencyOver200ms5MinSRE

yeya24:

https://issues.redhat.com/browse/OSD-4438

This alert never worked since it uses the wrong metric name https://github.com/openshift/managed-prometheus-exporter-dns/blob/master/monitor/main.py#L12. 

Also, we are planning to add a new DNS latency alert based on CoreDNS metrics. So it is okay to remove this old alert.

### DIFF
--- a/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
@@ -10,11 +10,6 @@ spec:
   groups:
   - name: sre-dns-alerts
     rules:
-    - alert: DNSLatencyOver200ms5MinSRE
-      expr: max_over_time(dns_latency_milliseconds_sum[5m]) > 0.2
-      labels:
-        severity: critical
-        namespace: openshift-monitoring
     - alert: DNSErrors05MinSRE
       expr: rate(dns_failure_failure_total[5m]) > 0
       for: 5m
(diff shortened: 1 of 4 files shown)


RiRa12621:
/lgtm
/assign @jewzaam 
this alert has never worked since the metrics it's relying on is not what we're getting from the exporter

jewzaam:
> /lgtm
> /assign @jewzaam
> this alert has never worked since the metrics it's relying on is not what we're getting from the exporter

Why not fix the alert `expr` to use the metric we do export?  It's `dns_latency_milliseconds`.

/hold

yeya24:
> > /lgtm
> > /assign @jewzaam
> > this alert has never worked since the metrics it's relying on is not what we're getting from the exporter
> 
> Why not fix the alert `expr` to use the metric we do export? It's `dns_latency_milliseconds`.
> 
> /hold

The metric itself is not correct. The name is `dns_latency_milliseconds` but the actual value is seconds.

RiRa12621:
From my perspective there are two things here: 
1. This has never worked since we introduced it and the metrics itself is declared wrong
2. There are metrics directly in [coreDNS](https://github.com/coredns/coredns/tree/master/plugin/metrics)

If we really think we need this, it should probably be part of the product as well and using `coredns_dns_request_duration_seconds{server, zone, type}`

RiRa12621:
@jewzaam seems the alert still hasn't worked all the time
I think we should merge this can focus on the conversation with @openshift/openshift-team-monitoring about adding coredns mixin for all of OCP
thoughts?

PR #407, opened 2020-07-23
Update IOPs burst balance alert rules

gsleeman:
* These alerts should be using min_over_time rather than avg
* Inverted balance for clarity when viewing Prometheus graphs

### DIFF
--- a/deploy/sre-prometheus/aws/100-ebs-iops-burstbalance.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/aws/100-ebs-iops-burstbalance.PrometheusRule.yaml
@@ -10,15 +10,17 @@ spec:
   groups:
   - name: sre-ebs-iops-burstbalance
     rules:
-    - alert: EbsVolumeBurstBalanceLT40PctSRE
-      expr: avg(ebs_iops_credits) < 40
+    - alert: EbsVolumeBurstBalanceLowSRE
+      expr: 100 - min_over_time(ebs_iops_credits[30m:5m]) > 80
       for: 60m
       labels:
         severity: warning
         namespace: openshift-monitoring
-    - alert: EbsVolumeBurstBalanceLT20PctSRE
-      expr: avg(ebs_iops_credits) < 20
-      for: 60m
+    - alert: EbsVolumeBurstBalanceEmptySRE
+      expr: 100 - min_over_time(ebs_iops_credits[3h:15m]) == 100
+      annotations:
+        message: "The EBS volume {{ $labels.vol_id }} has exhausted its IOPs burst credit for over 2 hours"
+      for: 120m
       labels:
         severity: critical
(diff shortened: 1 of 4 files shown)


gsleeman:
For reference, [the incident](https://issues.redhat.com/browse/OHSS-651) that prompted this change

And a [comparison of the alert rules](https://prometheus-k8s-openshift-monitoring.apps.ci.l2s4.p1.openshiftapps.com/graph?g0.range_input=3d&g0.end_input=2020-07-23%2003%3A00&g0.stacked=0&g0.expr=100%20-%20min_over_time(ebs_iops_credits%5B3h%3A15m%5D)%20%3D%3D%20100&g0.tab=0&g1.range_input=3d&g1.end_input=2020-07-23%2003%3A00&g1.stacked=0&g1.expr=100%20-%20min_over_time(ebs_iops_credits%5B30m%3A5m%5D)%20%3E%2080&g1.tab=0&g2.range_input=3d&g2.end_input=2020-07-23%2003%3A00&g2.expr=avg(ebs_iops_credits)%20%3C%2020&g2.tab=0) on the cluster in question

RiRa12621:
@jewzaam can you approve/deny?

deploy/sre-prometheus/aws/100-ebs-iops-burstbalance.PrometheusRule.yaml
```diff
@@ -10,15 +10,17 @@ spec:
...
         severity: warning
         namespace: openshift-monitoring
-    - alert: EbsVolumeBurstBalanceLT20PctSRE
-      expr: avg(ebs_iops_credits) < 20
-      for: 60m
+    - alert: EbsVolumeBurstBalanceEmptySRE
+      expr: 100 - min_over_time(ebs_iops_credits[3h:15m]) == 100
+      annotations:
+        message: "The EBS volume {{ $labels.vol_id }} has exhausted its IOPs burst credit for over 2 hours"
+      for: 120m
```
jewzaam, line 23: What does SRE do when this happens?  If we have this as a critical alert paging SRE we need an SOP.  I don't see that in ops-sop.

jewzaam (changes requested):

deploy/sre-prometheus/aws/100-ebs-iops-burstbalance.PrometheusRule.yaml
RiRa12621: That's true.
Sadly the existing alert didn't have one either.
I guess we need a card

gsleeman: It's hard to know how often this occurs and to what volumes - and I think that's going to be a factor in how we put together a SOP for this. I can try to put together a stub at least for tracing the volume to its workload and general debugging.

PR #409, opened 2020-07-27
OSD-4325 Add SyncSet generator script for MUO-driven upgrades

mrbarge:
This PR adds a script to generate the SyncSet bundles used for scheduling upgrades via the managed-upgrade-operator.

OSD4 SOP updates and documentation have been performed in ops-sop PR https://github.com/openshift/ops-sop/pull/886

See also: "managed-upgrade-operator Concept of Operations" document in SRE Google Drive.

### DIFF
--- a/scripts/generate_upgrade_syncset.py
+++ b/scripts/generate_upgrade_syncset.py
@@ -1,0 +1,301 @@
+#!/bin/env python
+
+import argparse
+import datetime
+import json
+import logging
+import os
+import re
+import subprocess
+import sys
+import ruamel.yaml as yaml
+from ruamel.yaml.scalarstring import DoubleQuotedScalarString as dq
+
+CS = yaml.comments.CommentedSeq
+CM = yaml.comments.CommentedMap
+
+# Script to generate the cluster list
+CLUSTER_LIST_EXEC = 'get-cluster-list.sh'
+# SyncSet CR constants
+SS_API_VERSION = 'hive.openshift.io/v1'
+SS_NAME = 'osd-upgrade-config'
+# UpgradeConfig CR constants
+UC_NAME = 'osd-upgrade-config'
+UC_NAMESPACE = 'openshift-managed-upgrade-operator'
+UC_API_VERSION = 'upgrade.managed.openshift.io/v1alpha1'
+UC_PDB_DRAIN_TIMEOUT = 60
+UC_FORCE = False
+
+def get_cluster_info(cluster_ids, groups, subgroup, environment):
+    """
+    Returns a list of clusters matching the supplied group and subgroup
+    :param group: Cluster group
+    :param subgroup: Cluster subgroup
+    :return: list of tuple pairs of cluster ID and cluster name
+    """
+
+    cluster_id_to_name = dict()
+
+    if cluster_ids is not None:
+        for cluster_id in cluster_ids:
+            try:
+                cluster_name = get_cluster_name_for_id(cluster_id, environment)
+                cluster_id_to_name[cluster_id] = cluster_name
+            except Exception as err:
+                logging.error('Unable to determine cluster name for cluster ID \'{}\': {}'.format(cluster_id, str(err)))
+                sys.exit(1)
+
+    if groups is not None:
+        for g in groups:
+            try:
+                # pout = subprocess.check_output(['cat', '/home/mbargenq/scratch/' + g + '.txt']).decode('utf-8')
+                pout = subprocess.check_output([CLUSTER_LIST_EXEC, g, subgroup]).decode('utf-8')
+                output_as_dict = json.loads(pout)
+
+                if 'items' not in output_as_dict:
+                    logging.error('Unexpected cluster info returned, unable to determine id/name')
+                    sys.exit(1)
+
+                cluster_items = output_as_dict['items']
+                for cluster in cluster_items:
+                    cluster_id_to_name[cluster['id']] = cluster['longName']
+
+            except OSError as err:
+                logging.info('Command \'{}\' failed with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
+            except subprocess.CalledProcessError as err:
+                logging.info('Command \'{}\' failed with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
+            except ValueError as err:
+                logging.info(
+                    'Output of command \'{}\' failed to parse with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
+
+    return cluster_id_to_name
+
+
+def get_cluster_name_for_id(cluster_id, environment):
+    """
+    Determines the clusterdeployment name associated with the supplied clusterid
+    :param cluster_id: cluster-id of cluster
+    :return: clusterdeployment name associated with cluster ID
+    """
+    cluster_ns = 'uhc-{}-{}'.format(environment, cluster_id)
+    try:
+        # pout = subprocess.check_output(['cat', '/home/mbargenq/scratch/cluster.txt']).decode(
+        #     'utf-8')
+        pout = subprocess.check_output(['oc', 'get', 'clusterdeployment', '-n', cluster_ns, '-o', 'json']).decode(
+            'utf-8')
+        output_as_dict = json.loads(pout)
+        if 'items' not in output_as_dict:
+            raise Exception('fetching clusterdeployment returned invalid response')
+
+        cd = output_as_dict['items']
+        if len(cd) == 0:
+            raise Exception('clusterdeployment not found')
+
+        return cd[0]['metadata']['name']
+
+    except Exception as err:
+        raise err
+
+
+def generate_upgradeconfig(start_time, version, channel):
+    """
+    Generates the UpgradeConfig CR in python dict form
+    :param version: version to upgrade to
+    :param channel: channel for upgrade
+    :param force: force upgrade
+    :return: dict representation of upgradeconfig CR
+    """
+    uc = {
+        'apiVersion': UC_API_VERSION,
+        'kind': 'UpgradeConfig',
+        'metadata': {
+            'name': UC_NAME,
+            'namespace': UC_NAMESPACE,
+        },
+        'spec': {
+            'type': dq('OSD'),
+            'upgradeAt': dq(start_time),
+            'proceed': True,
+            'PDBForceDrainTimeout': UC_PDB_DRAIN_TIMEOUT,
+            'desired': {
+                'version': dq(version),
+                'channel': dq(channel),
+                'force': UC_FORCE,
+            }
+        }
+    }
+    return uc
+
+
+def generate_syncset(upgradeconfig, cluster_id, cluster_name='', environment='production'):
+    """
+    Generates the SelectorSyncSet
+    :param upgradeconfig: UpgradeConfig CR as dict
+    :param cluster_id: Cluster ID to target
+    :param cluster_name: Cluster name to target
+    :return: SelectorSyncSet as dict
+    """
+
+    # generate the namespace for the SyncSet in hive format
+    # uhc-<environment>-<clusterID>
+    ss_namespace = 'uhc-{}-{}'.format(environment, cluster_id)
+
+    data = CM()
+    data['apiVersion'] = SS_API_VERSION
+    data['kind'] = 'SyncSet'
+    data['metadata'] = {
+        'labels': {
+            'api.openshift.com/id': cluster_id,
+            'api.openshift.com/name': cluster_name,
+        },
+        'name': SS_NAME,
+        'namespace': ss_namespace,
+    }
+
+    data['spec'] = {
+        'clusterDeploymentRefs': [
+            {
+                'name': dq(cluster_name),
+            }
+        ],
+        'resourceApplyMode': 'Upsert',
+        'resources': [
+            upgradeconfig
+        ],
+    }
+
+    yaml.preserve_quotes = True
+    yaml.dump(data, sys.stdout, Dumper=yaml.RoundTripDumper)
+
+
+def init_logging():
+    logging.basicConfig(level=logging.INFO, format='%(asctime)s %(levelname)s:%(name)s:%(message)s')
+
+
+def generate_syncsets(clusters, start_time, version, channel, environment='production'):
+    outer = CM()
+    outer['apiVersion'] = 'v1'
+    outer['kind'] = 'List'
+
+    items = CS()
+    for cluster_id, cluster_name in clusters.items():
+        uc = generate_upgradeconfig(start_time, version, channel)
+        ss_namespace = 'uhc-{}-{}'.format(environment, cluster_id)
+        data = CM()
+        data['apiVersion'] = SS_API_VERSION
+        data['kind'] = 'SyncSet'
+        data['metadata'] = {
+            'labels': {
+                'api.openshift.com/id': cluster_id,
+                'api.openshift.com/name': cluster_name,
+            },
+            'name': SS_NAME,
+            'namespace': ss_namespace,
+        }
+
+        data['spec'] = {
+            'clusterDeploymentRefs': [
+                {
+                    'name': dq(cluster_name),
+                }
+            ],
+            'resourceApplyMode': 'Upsert',
+            'resources': [
+                uc
+            ],
+        }
+        items.append(data)
+
+    outer['items'] = items
+
+    yaml.preserve_quotes = True
+    yaml.dump(outer, sys.stdout, Dumper=yaml.RoundTripDumper)
+
+
+def init_argparse():
+    """
+    Initialises the command-line argument parser
+    :return: parser instance
+    """
+    parser = argparse.ArgumentParser(description='UpgradeConfig SelectorSyncSet generator',
+                                     usage='%(prog)s [options]')
+
+    parser.add_argument('--subgroup', required=False, choices=['all', 'prod', 'nonprod'], default='all',
+                        help='Sub-cluster-group to target (default=all)')
+    parser.add_argument('--start-time', required=False, type=valid_date,
+                        help='Timestamp at which upgrade should commence (eg. 2020-01-25T00:00:00Z)')
+    parser.add_argument('--version', required=True, type=valid_version, help='Target upgrade version [required]')
+    parser.add_argument('--channel', required=True, help='Target upgrade channel [required]')
+    parser.add_argument('--force', '-f', required=False, default=False, help='Force upgrade [optional, default false]')
+    parser.add_argument('--environment', '-e', required=False, choices=['production', 'staging', 'integration'],
+                        default='production', help='Hive environment [optional, default production]')
+
+    cluster_group = parser.add_argument_group()
+    cluster_group.add_argument('--cluster-id', '-c', required=False, action='append', help='Cluster ID to target')
+    cluster_group.add_argument('--group', '-g', required=False, action='append',
+                               choices=['sd', 'sre', 'cicd', 'internal', 'snowflake', 'external'],
+                               help='Cluster group to target')
+
+    return parser
+
+
+def valid_version(v):
+    """
+    Verify the supplied version adheres to formatting conventions
+    :param v: version to verify
+    :return: the unmodified version if it parses successfully
+    :raises ArgumentError if version is invalid
+    """
+    p = re.compile('^4\\.[0-9]+\\.[0-9]+')
+    result = p.match(v)
+    if not result:
+        raise argparse.ArgumentError('Version must be in format 4.X.Y, eg. 4.3.25')
+    return v
+
+
+def valid_date(d):
+    """
+    Verify the supplied timestamp adheres to formatting conventions
+    :param d: timestamp to verify
+    :return: the unmodified timestamp if it parses successfully
+    :raises ArgumentError if timestamp is invalid
+    """
+    try:
+        datetime.datetime.strptime(d, '%Y-%m-%dT%H:%M:%SZ')
+    except ValueError:
+        raise argparse.ArgumentError('Time must be in format YYYY-MM-DDTHH:MM:SSZ, eg. 2020-01-25T00:00:00Z')
+    return d
+
+
+def check_environment():
+    """
+    Checks the system environment for the presence of the cluster-list fetcher
+    :return: bool indicating success
+    """
+    for path in os.environ["PATH"].split(os.pathsep):
+        exe_file = os.path.join(path, CLUSTER_LIST_EXEC)
+        if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
+            return True
+
+    logging.info("Missing required executable {}, ensure it is in PATH.".format(CLUSTER_LIST_EXEC))
+    return False
+
+
+if __name__ == '__main__':
+    # program initialization
+    init_logging()
+    parser = init_argparse()
+    args = parser.parse_args()
+
+    # # validate the dependent program(s) is valid
+    if not check_environment():
+        sys.exit(1)
+
+    # fetch the list of candidate clusters
+    clusters = get_cluster_info(args.cluster_id, args.group, args.subgroup, args.environment)
+
+    upgrade_config = generate_upgradeconfig(args.start_time, args.version, args.channel)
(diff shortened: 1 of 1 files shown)


mrbarge:
Requires review and the managed-upgrade-operator to actually go live, so adding a hold.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+                'force': UC_FORCE,
+            }
+        }
+    }
+    return uc
+
+
+def generate_syncset(upgradeconfig, cluster_id, cluster_name='', environment='production'):
+    """
+    Generates the SelectorSyncSet
```
dofinn, line 135: Nit picky, but this is now `SyncSet`

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    :param upgradeconfig: UpgradeConfig CR as dict
+    :param cluster_id: Cluster ID to target
+    :param cluster_name: Cluster name to target
+    :return: SelectorSyncSet as dict
```
dofinn, line 139: Previous SS comment

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    yaml.preserve_quotes = True
+    yaml.dump(outer, sys.stdout, Dumper=yaml.RoundTripDumper)
+
+
+def init_argparse():
+    """
+    Initialises the command-line argument parser
+    :return: parser instance
+    """
+    parser = argparse.ArgumentParser(description='UpgradeConfig SelectorSyncSet generator',
```
dofinn, line 223: Previous SS comment

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+                cluster_name = get_cluster_name_for_id(cluster_id, environment)
+                cluster_id_to_name[cluster_id] = cluster_name
+            except Exception as err:
+                logging.error('Unable to determine cluster name for cluster ID \'{}\': {}'.format(cluster_id, str(err)))
+                sys.exit(1)
+
+    if groups is not None:
+        for g in groups:
+            try:
+                # pout = subprocess.check_output(['cat', '/home/mbargenq/scratch/' + g + '.txt']).decode('utf-8')
```
dofinn, line 51: nuke this :)

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+
+def get_cluster_name_for_id(cluster_id, environment):
+    """
+    Determines the clusterdeployment name associated with the supplied clusterid
+    :param cluster_id: cluster-id of cluster
+    :return: clusterdeployment name associated with cluster ID
+    """
+    cluster_ns = 'uhc-{}-{}'.format(environment, cluster_id)
+    try:
+        # pout = subprocess.check_output(['cat', '/home/mbargenq/scratch/cluster.txt']).decode(
```
dofinn, line 85: nuke these :)

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+        for cluster_id in cluster_ids:
+            try:
+                cluster_name = get_cluster_name_for_id(cluster_id, environment)
+                cluster_id_to_name[cluster_id] = cluster_name
+            except Exception as err:
+                logging.error('Unable to determine cluster name for cluster ID \'{}\': {}'.format(cluster_id, str(err)))
+                sys.exit(1)
+
+    if groups is not None:
+        for g in groups:
```
2uasimojo, line 49: similar here
```suggestion
    for g in groups or ():
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
+#!/bin/env python
```
2uasimojo, line 1: I didn't check where/how this is going to be running, but do we want to be explicit about using python3? (Hint: We *really* [shouldn't be using python2](https://www.python.org/doc/sunset-python-2/).)

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def valid_version(v):
+    """
+    Verify the supplied version adheres to formatting conventions
+    :param v: version to verify
+    :return: the unmodified version if it parses successfully
+    :raises ArgumentError if version is invalid
+    """
+    p = re.compile('^4\\.[0-9]+\\.[0-9]+')
```
2uasimojo, line 252: Recommend using raw strings here. It allows you to get rid of the backslash doubling, but it's also best practice for regexes anyway.
```suggestion
    p = re.compile(r'^4\.[0-9]+\.[0-9]+')
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+            except OSError as err:
+                logging.info('Command \'{}\' failed with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
+            except subprocess.CalledProcessError as err:
+                logging.info('Command \'{}\' failed with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
+            except ValueError as err:
+                logging.info(
+                    'Output of command \'{}\' failed to parse with error: {}'.format(CLUSTER_LIST_EXEC, str(err)))
+                sys.exit(1)
```
2uasimojo, line 72: Ditto above re double quotes.

Also, since the reactions for the first two exceptions are identical, you could consolidate them. (You could collapse all three if you decided "to parse" wasn't an important distinction, given you're printing the exception anyway.)

```suggestion
            except (OSError, subprocess.CalledProcessError) as err:
                logging.info("Command '{}' failed with error: {}".format(CLUSTER_LIST_EXEC, str(err)))
                sys.exit(1)
            except ValueError as err:
                logging.info(
                    "Output of command '{}' failed to parse with error: {}".format(CLUSTER_LIST_EXEC, str(err)))
                sys.exit(1)
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+UC_API_VERSION = 'upgrade.managed.openshift.io/v1alpha1'
+UC_PDB_DRAIN_TIMEOUT = 60
+UC_FORCE = False
+
+def get_cluster_info(cluster_ids, groups, subgroup, environment):
+    """
+    Returns a list of clusters matching the supplied group and subgroup
+    :param group: Cluster group
+    :param subgroup: Cluster subgroup
+    :return: list of tuple pairs of cluster ID and cluster name
```
2uasimojo, line 34: This is returning a dict, not a list of 2-tuples.

```suggestion
    Returns a mapping of clusters matching the supplied group and subgroup
    :param group: Cluster group
    :param subgroup: Cluster subgroup
    :return: dict, keyed by cluster ID, of cluster names
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+            raise Exception('fetching clusterdeployment returned invalid response')
+
+        cd = output_as_dict['items']
+        if len(cd) == 0:
+            raise Exception('clusterdeployment not found')
+
+        return cd[0]['metadata']['name']
+
+    except Exception as err:
+        raise err
```
2uasimojo, line 100: This is redundant. Just un-try/except the whole thing.

(Actually, to be pedantic, it's a teeny bit worse than redundant: by saying `raise err` rather than just `raise`, the exception that percolates up will have the stack trace from L100 rather than from wherever the actual exception occurred.)

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+    cluster_ns = 'uhc-{}-{}'.format(environment, cluster_id)
+    try:
+        # pout = subprocess.check_output(['cat', '/home/mbargenq/scratch/cluster.txt']).decode(
+        #     'utf-8')
+        pout = subprocess.check_output(['oc', 'get', 'clusterdeployment', '-n', cluster_ns, '-o', 'json']).decode(
+            'utf-8')
+        output_as_dict = json.loads(pout)
+        if 'items' not in output_as_dict:
+            raise Exception('fetching clusterdeployment returned invalid response')
```
2uasimojo, line 91: (Here and L95) Probably overkill for a script like this, but raising the base `Exception` class is discouraged; you're supposed to either use a stock exception or create your own, e.g.

```python
class ClusterDeploymentNotFoundException(Exception):
    pass
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+import argparse
+import datetime
+import json
+import logging
+import os
+import re
+import subprocess
+import sys
+import ruamel.yaml as yaml
+from ruamel.yaml.scalarstring import DoubleQuotedScalarString as dq
```
2uasimojo, line 12: Not used

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+
+import argparse
+import datetime
+import json
+import logging
+import os
+import re
+import subprocess
+import sys
+import ruamel.yaml as yaml
```
2uasimojo, line 11: I noticed you don't seem to be making any actual use of any of the ruamel-isms in this script. Not that it's really a problem to use it anyway, but I'm curious why you chose this rather than the stock `yaml` lib?

...unless it matters to you that the keys remain in the order you set them? In which case `oyaml` would work too.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+                'name': dq(cluster_name),
+            }
+        ],
+        'resourceApplyMode': 'Upsert',
+        'resources': [
+            upgradeconfig
+        ],
+    }
+
+    yaml.preserve_quotes = True
```
2uasimojo, line 170: This is setting a global variable in the ruamel.yaml import, which a) isn't a great idea, and b) isn't doing anything, I'm pretty sure. The documentation isn't great (which would be a reason to use the stock `yaml` lib if you don't need the ruamel-isms) but I believe `preserve_quotes` is a kwarg to the "new API" `YAML()` class initializer and some of its methods.

Does the output change if you remove this line? (You would have to remove [L214](https://github.com/openshift/managed-cluster-config/pull/409/files#diff-b2d12c252edac3e97ef460c9e9c1e4eaR214) as well, because global :P )

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+            'resourceApplyMode': 'Upsert',
+            'resources': [
+                uc
+            ],
+        }
+        items.append(data)
+
+    outer['items'] = items
+
+    yaml.preserve_quotes = True
```
2uasimojo, line 214: See [above](https://github.com/openshift/managed-cluster-config/pull/409/files#diff-b2d12c252edac3e97ef460c9e9c1e4eaR170)

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    yaml.dump(outer, sys.stdout, Dumper=yaml.RoundTripDumper)
+
+
+def init_argparse():
+    """
+    Initialises the command-line argument parser
+    :return: parser instance
+    """
+    parser = argparse.ArgumentParser(description='UpgradeConfig SelectorSyncSet generator',
+                                     usage='%(prog)s [options]')
```
2uasimojo, line 224: Why are you overriding `usage`? Without it, argparse generates a pretty useful usage statement:

```
usage: generate_upgrade_syncset.py [-h] [--subgroup {all,prod,nonprod}]
                                   [--start-time START_TIME] --version VERSION
                                   --channel CHANNEL [--force FORCE]
                                   [--environment {production,staging,integration}]
                                   [--cluster-id CLUSTER_ID]
                                   [--group {sd,sre,cicd,internal,snowflake,external}]
```

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+
+def check_environment():
+    """
+    Checks the system environment for the presence of the cluster-list fetcher
+    :return: bool indicating success
+    """
+    for path in os.environ["PATH"].split(os.pathsep):
+        exe_file = os.path.join(path, CLUSTER_LIST_EXEC)
+        if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
+            return True
```
2uasimojo, line 281: If you're using py3 (and you should be, see [above](https://github.com/openshift/managed-cluster-config/pull/409/files#diff-b2d12c252edac3e97ef460c9e9c1e4eaR1)) you can use [shutil.which()](https://docs.python.org/dev/library/shutil.html#shutil.which) for this.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    return False
+
+
+if __name__ == '__main__':
+    # program initialization
+    init_logging()
+    parser = init_argparse()
+    args = parser.parse_args()
+
+    # # validate the dependent program(s) is valid
```
2uasimojo, line 293: Femtonit: extra `# `.

scripts/generate_upgrade_syncset.py
lisa: 👍  Python 2 is EOL.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+# Script to generate the cluster list
+CLUSTER_LIST_EXEC = 'get-cluster-list.sh'
+# SyncSet CR constants
+SS_API_VERSION = 'hive.openshift.io/v1'
+SS_NAME = 'osd-upgrade-config'
+# UpgradeConfig CR constants
+UC_NAME = 'osd-upgrade-config'
+UC_NAMESPACE = 'openshift-managed-upgrade-operator'
+UC_API_VERSION = 'upgrade.managed.openshift.io/v1alpha1'
+UC_PDB_DRAIN_TIMEOUT = 60
```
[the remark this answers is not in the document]
lisa: I don't think Python debugger, but I'm also not sure what PDB stands for here. :)

Can we have a comment explaining what PDB is and what the timeout is for?

scripts/generate_upgrade_syncset.py
lisa: I have another nit here:

The name of this method is `generate_syncset` and the comment says `Generates the SelectorSyncSet`. I'm glad there's a comment, but based on the method name I can see what this method is doing. A comment to explain what the method is used for would be a welcome improvement. Perhaps: `Generates the SyncSet for ..... to/so that....`

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    :param cluster_id: Cluster ID to target
+    :param cluster_name: Cluster name to target
+    :return: SelectorSyncSet as dict
+    """
+
+    # generate the namespace for the SyncSet in hive format
+    # uhc-<environment>-<clusterID>
+    ss_namespace = 'uhc-{}-{}'.format(environment, cluster_id)
+
+    data = CM()
```
lisa, line 146: Is there a native Python `SyncSet` object that can be used instead of crafting a YAML document? The hope is that the native object will wrap the construction of the object and provide guardrails ensuring that required attributes are present and of the correct type.

scripts/generate_upgrade_syncset.py
lisa, line 170: What is this line intending to do? (A comment please?)

scripts/generate_upgrade_syncset.py
lisa: @2uasimojo It is. See lines 119, 124, 125, 161, 202. I'm curious why it's used and what it does.

scripts/generate_upgrade_syncset.py
lisa: Be careful with a firm regex. Will this ever need to match rc builds, or nightlies? I don't think they adhere to the `x.y.z` where x, y, z are each integers.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+        raise err
+
+
+def generate_upgradeconfig(start_time, version, channel):
+    """
+    Generates the UpgradeConfig CR in python dict form
+    :param version: version to upgrade to
+    :param channel: channel for upgrade
+    :param force: force upgrade
+    :return: dict representation of upgradeconfig CR
```
lisa, line 109: These comments don't mention what `start_time` is doing? And, if I read it right, this method has no `force` parameter?

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    Verify the supplied timestamp adheres to formatting conventions
+    :param d: timestamp to verify
+    :return: the unmodified timestamp if it parses successfully
+    :raises ArgumentError if timestamp is invalid
+    """
+    try:
+        datetime.datetime.strptime(d, '%Y-%m-%dT%H:%M:%SZ')
+    except ValueError:
+        raise argparse.ArgumentError('Time must be in format YYYY-MM-DDTHH:MM:SSZ, eg. 2020-01-25T00:00:00Z')
+    return d
```
lisa, line 270: Why is this needed?

scripts/generate_upgrade_syncset.py
2uasimojo: Linked to [`dq` above](https://github.com/openshift/managed-cluster-config/pull/409/files/f434cede5b9be15182902c779eb78de9860c58fd#diff-b2d12c252edac3e97ef460c9e9c1e4eaR12), it's intended to make the dumped YAML include any redundant quotes around strings, if they were present on input/creation. One reason you might want to do that ([example](https://gitlab.cee.redhat.com/service/app-interface/-/blob/1428b7c41f71d2811c4b6a20349b471f5e290116/hack/new_osd_operator/new_operator.py#L12)) is to minimize diffs when you're changing existing YAML documents that were (usually) written by hand. That doesn't apply here AFAICT.

scripts/generate_upgrade_syncset.py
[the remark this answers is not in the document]
lisa: Oooh of course. Too many TLAs in our industry! 😆

scripts/generate_upgrade_syncset.py
mrbarge: Thanks for noticing this - it was a holdover from a previous revision of the code which dealt with SSS. I've removed all evidence of that previous iteration from the script and updated comments.

scripts/generate_upgrade_syncset.py
mrbarge: Addressed in 0acedad

scripts/generate_upgrade_syncset.py
mrbarge: Addressed in 0acedad

scripts/generate_upgrade_syncset.py
mrbarge: :facepalm: How embarassing! Fixed in 0acedad

scripts/generate_upgrade_syncset.py
mrbarge: Addressed in 0acedad

scripts/generate_upgrade_syncset.py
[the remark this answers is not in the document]
mrbarge: A fair point - I've clarified these with a rename of constants and added comments.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+
+    cluster_id_to_name = dict()
+
+    if cluster_ids is not None:
+        for cluster_id in cluster_ids:
+            try:
+                cluster_name = get_cluster_name_for_id(cluster_id, environment)
+                cluster_id_to_name[cluster_id] = cluster_name
+            except Exception as err:
+                logging.error('Unable to determine cluster name for cluster ID \'{}\': {}'.format(cluster_id, str(err)))
```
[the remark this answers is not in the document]
mrbarge: Thanks, good catch.

scripts/generate_upgrade_syncset.py
```diff
@@ -0,0 +1,301 @@
...
+    Returns a list of clusters matching the supplied group and subgroup
+    :param group: Cluster group
+    :param subgroup: Cluster subgroup
+    :return: list of tuple pairs of cluster ID and cluster name
+    """
+
+    cluster_id_to_name = dict()
+
+    if cluster_ids is not None:
+        for cluster_id in cluster_ids:
```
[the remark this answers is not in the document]
mrbarge: This is awesome, I love this! Have switched to it.

scripts/generate_upgrade_syncset.py
mrbarge: As above :+1:

scripts/generate_upgrade_syncset.py
mrbarge: This will need to run on a bastion and the OSD3 Hive - sadly the latter only has python2. I've tested it in both py2 and py3; the /bin/env is picking up the correct python in both locations (that is to say, bastions will run it via python 3, OSD3 Hive will just call python 2). As this script has a limited life of 1-2 months I'm hoping it'll be a workable compromise.

scripts/generate_upgrade_syncset.py
mrbarge: Migrated to PyYAML for commit e1449f7 and removed/tidied all of this up.

scripts/generate_upgrade_syncset.py
mrbarge: Thanks, agree - I've just left it as-is to use the default in commit e1449f7

scripts/generate_upgrade_syncset.py
mrbarge: Fixed.

scripts/generate_upgrade_syncset.py
mrbarge: I had a look around but was unable to find any helper packages unfortunately. I'm hoping to eventually be able to use the cluster_mgmt API instead of this clunky via-Hive method (which would eliminate the need to be hand-rolling a SyncSet), but that's still TBD pending discussions with the SDA. Doing it this way will be the contingency.

scripts/generate_upgrade_syncset.py
mrbarge: I've ported to PyYAML and removed the need for all of this in e1449f7

scripts/generate_upgrade_syncset.py
mrbarge: Re versions: for the scripts lifetime (1-2 months) we won't need to use it for anything other than stable or fast channel versions. Your point about the regex is sound though. I'll think on the necessity of this validation.
Re start_time: thanks! I've updated the comments to be accurate to the function's signature.

scripts/generate_upgrade_syncset.py
mrbarge: The date is going straight into the UpgradeConfig CR as the scheduled upgrade time in UTC, and needs to be in ISO-8601 format when it hits the CR. So this is just forcing the user of the script to put the correct format in initially to avoid the SyncSet pushing something out with a bad timestamp.

mrbarge:
@lisa @2uasimojo @dofinn Thanks so much for the thorough review, that's made my day! :smile: 

I've hopefully addressed the bulk of your notes in the latest commit.

On the topic of yaml, I've ported it over to PyYAML now and tidied up a lot of the code around that. I was originally using `ruamel.yaml` because I was generating a `SelectorSyncSet` and was getting fancy with commented list items (something `ruamel.yaml` let me do). But that was many revisions ago and no longer relevant to the script's current purpose. And you're both correct, all the double-quoting of values was unnecessary.

scripts/generate_upgrade_syncset.py
2uasimojo: I had assumed you would be running it in a container in any case, is that not true?

Even then, I get that it doesn't seem worth the effort to change over. Unfortunately, taken individually, every other case of this is thinking the same way. So when we do finally decide to move, it'll be a death of a thousand cuts.

Not blocking on this, but I really want to put it in people's heads.

2uasimojo:
Updates look good, thanks. I'm nobody in this repo so I'll refrain from doing any official stampy things.

PR #410, opened 2020-07-28
Add "openshift-sre-sshd" namespace

mbarnes:
The SSHD pod controller ([OSD-3599](https://issues.redhat.com/browse/OSD-3599)) and SSH key ConfigMaps ([OSD-3601](https://issues.redhat.com/browse/OSD-3601)) will exist here. 
Requires a service load balancer quota exemption.

### DIFF
--- a/deploy/osd-ingress/00-openshift-sre-sshd.Namespace.yaml
+++ b/deploy/osd-ingress/00-openshift-sre-sshd.Namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-sre-sshd
--- a/deploy/resource-quotas/10-patch.namespace.openshift-sre-sshd.yaml
+++ b/deploy/resource-quotas/10-patch.namespace.openshift-sre-sshd.yaml
@@ -1,0 +1,7 @@
+kind: Namespace
+apiVersion: v1
+name: openshift-sre-sshd
+applyMode: AlwaysApply
+patch: |-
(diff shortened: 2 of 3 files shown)


lisa:
How will the cloud-ingress-operator be able to watch resources in that namespace (cf https://github.com/openshift/cloud-ingress-operator/pull/76)?

mbarnes:
@lisa Can I ask, where is it documented that cross-namespace watches are not supported?

I understand cross-namespace _ownership references_ are forbidden, but my test cluster observations suggest watches are cluster-wide.  With the controller running in `openshift-cloud-ingress-operator`, ...

- If I create an SSHD resource in the `openshift-sre-sshd` namespace, the controller sees it immediately.
- If I delete the controller-created Deployment or Service in the `openshift-sre-sshd` namespace, the controller also reacts immediately.

If I _remove_ those watches on Deployments and Services from the controller code, then the controller takes significantly longer to react when I delete one of those controller-created objects.

I'm just trying to square my observations with what you and @cblecker are saying.

mbarnes:
@cblecker pointed me to [this bit of the operator configuration](https://github.com/openshift/cloud-ingress-operator/blob/e21b81b0d3dd3a400b2761b6ad212617d772e379/deploy/50_cloud-ingress-operator.Deployment.yaml#L39-L41), which sets `WATCH_NAMESPACE` to `""` to enable watching all namespaces.  So that explains my observations above and hopefully answers @lisa's question.

lisa:
> @lisa Can I ask, where is it documented that cross-namespace watches are not supported?

I think this was covered and clarified in the other PR? If not, happy to reiterate here.

PR #419, opened 2020-07-30
Multiwindowburn

RiRa12621:
Templated out multiwindow multiburn recording and alerting
from here on any other route just needs to add 

```
      record: route:error_slo:percent
      labels:
        route: console
      expr: 0.5
```
with their corresponding expr according to the SLO to get burn alerts

### DIFF
--- a/deploy/sre-prometheus/100-multiwindow-multiburn.yaml
+++ b/deploy/sre-prometheus/100-multiwindow-multiburn.yaml
@@ -1,0 +1,143 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-multiwindow-multiburn
+    role: alert-rules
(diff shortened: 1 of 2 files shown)


RiRa12621:
/hold
needs metrics only available starting from 4.5 to work
cc @jeremyeder

RiRa12621:
in the face of our upstream efforts: to get this upstream, we'd need the commitment frmo the console team or later on any other team for a component, which is added, that they will make this SLO

brancz:
very nice :+1:

metalmatze:
With the slo-libsonnet project I started using the `burnrate5m` suffix. I think that's a bit more readable.

`route:haproxy_route_backend_http_errors:ratio_rate1h`
to
`route:haproxy_route_backend_http_errors:burnrate1h`.

Otherwise, this is pretty much similar. :+1:

PR #421, opened 2020-07-30
Add option to install local pre-push hook to run PR check

jewzaam:
My last 4 PRs I not run `make`.  It annoyed me so I made a simple pre-push hook for myself.
This PR shares it with others with a way to install and docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,3 +54,7 @@ clean:
 git-commit:
 	git add ${SELECTOR_SYNC_SET_DESTINATION}
 	git commit -m "Updated selectorsynceset template added"
+
+.PHONY: install-pre-commit-hook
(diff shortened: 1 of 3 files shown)


jewzaam:
Fun side note, I had done the pre-push hook in a branch that would fail pushing.  Upon attempting to push to mcc it failed locally!  And then I fixed it all back to where it should be.. it surprised me to see it run at first because it's verbose but it did its job.

RiRa12621:
i'd like to see .git/hooks/ added to gitignore so no one has issues constantly checking that in

jewzaam:
@RiRa12621 `.git` is ignored by default, it's where git adds all its things..

PR #429, opened 2020-08-06
Disable curated operators for cluster-admin enabled clusters

jewzaam:
https://issues.redhat.com/browse/OSD-4754


jewzaam:
/hold

Need someone to sanity check that we _should_ do this before we merge it.  It seems reasonable to me but..

jewzaam:
/hold cancel

Removed hold given update to patch subscriptions and internal discussions that basically ended in +1 from folks.

PR #438, opened 2020-08-19
changed severity level of SLAUptimeSRE alert

sulmer-redhat:
OSD-4517

/hold @jewzaam commented:
 "Holding until we have https://issues.redhat.com/browse/OSD-4414 addressed (SOP on this alert). The big question I have is "what actions does SRE take when this alert fires?" and it's not yet answered so we cannot hit on-call with this as a critical alert yet."

### DIFF
--- a/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-api-errors.PrometheusRule.yaml
@@ -17,5 +17,5 @@ spec:
         (sum(rate(apiserver_request_total{job="apiserver", code=~"(400|5..)"}[5m])) / sum(rate(apiserver_request_total{job="apiserver", code=~"([1-5]..)"}[5m])) >= 1)
       for: 5m
       labels:
-        severity: warning
+        severity: critical
         namespace: openshift-monitoring
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -10421,7 +10421,7 @@ objects:
               '
             for: 5m
             labels:
-              severity: warning
+              severity: critical
               namespace: openshift-monitoring
     - apiVersion: monitoring.coreos.com/v1
       kind: PrometheusRule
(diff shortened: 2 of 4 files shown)


jewzaam:
/approve

Just pending the SOP as noted.

NautiluX:
SOP in place now. This PR will need rebase, corresponding ticket in OSD unblocked.

bdematte:
It would be good to add the link to the SOP now it exists to ease investigations.

NautiluX:
the SOP is at the default place for the alert, so it should work.

dofinn:
https://github.com/openshift/ops-sop/blob/master/v4/alerts/SLAUptimeSRE.md 
+1 from me. Will let others way in.

NautiluX:
@superfunhappytime this is the PR to raise the severity of the alert. I understand the original purpose was only for a dashboard creation. However, if we breach the SLA, that will be a heavy symptom so I would assume an alert is appropriate. We created an SOP for getting started debugging API server failures, and most probably this alert won't come without any context at warning or info level, so I would really think it is quite actionable.

dofinn:
FYI, there is also this https://github.com/openshift/configure-alertmanager-operator/blob/master/pkg/controller/secret/secret_controller.go#L125

NautiluX:
well that's an interesting find @dofinn

RiRa12621:
https://github.com/openshift/configure-alertmanager-operator/pull/136
to remove the silence

jharrington22:
@NautiluX @dofinn @RiRa12621 @bdematte 

I'm in favor of changing this to critical but before that can happen the SOP needs to have more actionable content. As an SRE once I run the `crictl ps` and `crilogs` commands, verify etc and networking whats next? The AWS status page almost never shows actual problems. 

Should the result of this be an OHSS ticket?

RiRa12621:
had conversation with @jharrington22 offline about this: we need to have a proper policy for this SLO, that can be linked for to the SOP so it's crystal clear
I'll set something up next week to define one

RiRa12621:
let's move ahead with this and at the same time kick off the conversation again on a potential policy

NautiluX:
just realized this still has a hold. @ramonbutter @sulmer-redhat can you check and `\ unhold` when you're ready to merge?

PR #458, opened 2020-09-02
Update osde2e-noalerts cronjob to ignore moa clusters in non-prod

jewzaam:
https://issues.redhat.com/browse/OSD-5032


jharrington22 (approved):
/lgtm

PR #459, opened 2020-09-03
Change prometheus retention to 7 days

dofinn:
Context: https://issues.redhat.com/browse/OSD-4875


RiRa12621:
I'd honestly rather see us using bigger volumes instead of reducing the retention further

dofinn:
IMO we should increase volume/retention by requirement. This PR is a quick win to remove the alert we can do nothing about. Do we have cases where people want longer retention?

RiRa12621:
You can barely see anything with 7 days of metrics 
Also we currently don't have means to increase volumes on a case by case basis

georgettica:
woot!

PR #464, opened 2020-09-03
OSD-3304 Disable curation for clusters with extended-dedicated-admin …

rogbas:
…label != false

Clusters with such label will be blocked from upgrading to v4.5


rogbas:
PR on hold, it should be merged and promoted to prod on Sep 8th NASA shift

PR #465, opened 2020-09-09
enable devaccess get on clusterversion

dofinn:


jewzaam:
/lgtm
/hold

Consider if we take https://github.com/openshift/managed-cluster-config/pull/466 which goes further in what it grants to both dedicated-readers and osd-devaccess.

jewzaam:
Done with https://github.com/openshift/managed-cluster-config/pull/466 (and fixed with https://github.com/openshift/managed-cluster-config/pull/467).  Closing.

PR #472, opened 2020-09-10
Delete replacing CSVs from openshift-monitoring

cblecker:
This should delete any replacing CSVs in openshift-monitoring. Once complete, we can delete this cronjob.


cblecker:
overriding images:

cblecker:
/override images
overriding images as it's not critical and there is a CI issue: https://coreos.slack.com/archives/CFUGK0K9R/p1599775151015700

PR #478, opened 2020-09-11
Revert crd pruner changes

mrbarge:


jewzaam:
From @cblecker:  "basically, was using MCC to create a job to do cleanup of bad CSVs a couple nights ago because of a bad c-am-o rollout"
This is to remove that temporary job.

PR #480, opened 2020-09-14
OSD-5137 Add OCM_BASE_URL into MUO configmap

mrbarge:
This PR adds a new configmap entry to `managed-upgrade-operator`'s configmap, with a value read from `OCM_BASE_URL`.

OCM_BASE_URL is set per OSD environment in app-interface via MR 8256. [0] 

This is needed in order to supply the `managed-upgrade-operator` with the correct OCM API to query for upgrade policies and service log announcements.

Refs: OSD-5137

[0] https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/8256/

### DIFF
--- a/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
+++ b/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-managed-upgrade-operator
 data:
   config.yaml: |
+    ocmBaseUrl: "${{OCM_BASE_URL}}"
     maintenance:
       controlPlaneTime: 90
       workerNodeTime: 8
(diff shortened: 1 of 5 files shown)


deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-managed-upgrade-operator
 data:
   config.yaml: |
+    ocmBaseUrl: "${{OCM_BASE_URL}}"
```
jewzaam, line 8: `OCM_BASE_URL` needs to be defined in the template parameters.
https://github.com/openshift/managed-cluster-config/blob/b3837d88798a0c20715d057ed1efccceb3e3484e/scripts/templates/template.yaml#L3

jewzaam (changes requested):

PR #497, opened 2020-09-28
make the verification step checked/ignored namespace configurable

bmeng:


deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
```diff
@@ -31,3 +31,10 @@ data:
       - MetricsClientSendFailingSRE
       - UpgradeNodeScalingFailedSRE
       - UpgradeClusterCheckFailedSRE
+    verification:
+      ignoredNamespaces:
+      - openshift-logging
+      namespacePrefixesToCheck:
+      - openshift
+      - kube
+      - default
```
dofinn, line 40: have we reviewed any redhat-* requirements?

bmeng: not yet, just copied the values from here. https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/osd_cluster_upgrader/upgrader.go#L363 
Haven't seen any project with name redhat, but maybe can create spike for that.

dofinn: +1 we need to ascertain if any layered product alerts should block an osd upgrade.

bmeng: https://issues.redhat.com/browse/OSD-5365

PR #499, opened 2020-09-28
Velero alert on 3 failed backups

dofinn:
We currently have no SLO for storing 100% of hourly backups. This alert
appears too frequently to be paged on. Dropping this to 3 sequential
failed backups before alerting.

Evidence is available in pager duty.

### DIFF
--- a/deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-velero-operator.PrometheusRule.yaml
@@ -15,11 +15,12 @@ spec:
     # the deployment's creation time if there's been no backup yet), plus a
     # small static buffer to account for some variation.
     - alert: VeleroHourlyObjectBackupMissed
-      # Hourly backup should complete about 3600 seconds (1 hour) after the
+      # Hourly backup can fail twice before alerting. 
+      # should complete about 10800 seconds (3 hours) after the
       # last successful backup, with a 600 second (10 minute) buffer.
       expr: |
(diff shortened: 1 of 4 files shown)


dofinn:
cc @jewzaam @RiRa12621 @cblecker

mrbarge:
+1 from me - been seeing this somewhat regularly and almost always seems to have sorted itself out in subsequent jobs. This would be in keeping with similar "only alert if continuous jobs fail" work being done in cards like OSD-4540 and OSD-5354.

cblecker:
/hold
for discussion

RiRa12621:
Probably should be renaming the alert as well to not be misleading, since it's now not just one hourly backup that's being missed.
+1 to the idea

jewzaam:
/lgtm

Seems like generally accepted.  I'll leave the hold in case we want to rename the alert.  I think it's fine the way it is given it is an hourly backup that was missed, but maybe it could be `Velero3ConsecutiveHourlyBackupsMissed`?

cblecker:
Can we modify the "message" that is passed on L28?

Otherwise, yeah, I think this is probably the best for now, although suboptimal

RiRa12621:
Should probably still have a Jira card to fix the underlying root cause in addition @cblecker

dofinn:
> Can we modify the "message" that is passed on L28?
> 
> Otherwise, yeah, I think this is probably the best for now, although suboptimal

resolved in https://github.com/openshift/managed-cluster-config/pull/499/commits/32096cc5f4721b3604b27445806d3c834036b1ca +  alertName

PR #503, opened 2020-10-01
Add BZ1883589 workaround

cblecker:
Adds the workaround from https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c8 to all 4.5 clusters that aren't 4.5.99 (a fixed version that we'll have to replace later)

### DIFF
--- a/deploy/bz1883589-kubeapiserver-revert/00-cluster.kubeapiserver.patch.yaml
+++ b/deploy/bz1883589-kubeapiserver-revert/00-cluster.kubeapiserver.patch.yaml
@@ -1,0 +1,7 @@
+apiVersion: operator.openshift.io/v1
+kind: KubeAPIServer
+name: cluster
+applyMode: AlwaysApply
+# https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c8
+patch: '[{ "op": "remove", "path": "/spec/unsupportedConfigOverrides" }]'
+patchType: json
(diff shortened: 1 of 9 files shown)


deploy/bz1883589-kubeapiserver-revert/00-cluster.kubeapiserver.patch.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: operator.openshift.io/v1
+kind: KubeAPIServer
+name: cluster
+applyMode: AlwaysApply
+# https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c8
+patch: '[{ "op": "remove", "path": "/spec/unsupportedConfigOverrides" }]'
+patchType: json
```
jewzaam, line 7: If we have any other overrides set this would remove all of them. Can we remove just `/spec/unsupportedConfigOverrides/apiServerArguments` ?

jewzaam (changes requested):

deploy/bz1883589-kubeapiserver-revert/00-cluster.kubeapiserver.patch.yaml
cblecker: @jewzaam any other config overrides would block upgrades. I believe we have to remove the entire section if we want 4.6 upgrades to be unblocked

jewzaam: If we create additional overrides they may not be fixed in the same z as this one.  It's a risk, small and not likely.

jewzaam:
Decided we'll merge and refine as needed.  This issue isn't going to be forgotten.

/lgtm
/approve

PR #506, opened 2020-10-02
Fix flowschema apiversion and revert prior override in non-prod

jewzaam:
https://bugzilla.redhat.com/show_bug.cgi?id=1883589


rogbas:
It's missing running make generate-synset

PR #518, opened 2020-10-19
Prune replacing CSVs from bad CAMO deploy

cblecker:


deploy/sre-pruning/110-pruning-csvs.CronJob.yaml
```diff
@@ -0,0 +1,26 @@
...
+          restartPolicy: Never
+          containers:
+          - name: csv-pruner
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            args:
+            - /bin/bash
+            - -c
+            - >-
+              oc -n openshift-monitoring get csv | grep Replacing | awk '{print $1}' | xargs oc -n openshift-monitoring delete csv --wait=0 --ignore-not-found
```
jewzaam, line 26: We don't want to leave this around long, very general.  Could we search for specific version text that is being replaced?

jewzaam (changes requested):

PR #527, opened 2020-10-22
Prepare rbac-permissions-operator-config for upsert

fahlmant:
PR Part 2 for https://issues.redhat.com/browse/OSD-5566. This PR applies `Upsert` to the rbac-permissions-operator-config SSS so that the next PR can be successfully applied


jewzaam (changes requested):
Need to delete `deploy/osd-project-request-template/config.yaml` to get that back to using `Sync`

PR #528, opened 2020-10-22
Clean up SSS for rbac-permissions-operator and osd-project-request-template.

fahlmant:
PR part 3 for https://issues.redhat.com/browse/OSD-5566

This PR removes the template and namespace from being deployed via the rbac-permissions-operator SSS, and removes RBAC from the osd-project-request-template SSS that will be moved to the aggregated dedicated-admins permissions directly (#524)

Depends on #527

PR #529, opened 2020-10-22
Add temporary Template to fix clusters in non-prod

jewzaam:
https://issues.redhat.com/browse/OSD-5566

We didn't account for bug https://issues.redhat.com/browse/CO-1188 in the promotion to stage, project creation is broken for clusters that existed before merging https://github.com/openshift/managed-cluster-config/pull/528


jewzaam:
CLosing, we didn't promote #527 before merging #528, so #528 needs reverted

PR #533, opened 2020-10-23
Fix AlertmanagerSilencesActiveSRE to see when there are 2+ silences

cblecker:
If there are 2+ silences, the current metric doesn't match.


RiRa12621:
/lgtm
Less alerts Always better.
We should generally talk about this again.
With metrics and alerts being available in the console there's theoretically no need anymore for customers to have direct access to alertmanager. That would remove the possibility of them silencing alerts and we could get rid of this alert.

PR #536, opened 2020-10-26
Revert "Revert "Revert "Clean up SSS for rbac-permissions-operator and osd-project-request-template."""

jewzaam:
Reverts openshift/managed-cluster-config#532

https://issues.redhat.com/browse/CO-1238

tl;dr: the ClusterRoleBinding didn't delete because we have .metadata.namespace on the resource in the SSS.  we have to revert, fix the NS, then we can delete it.


fahlmant (approved):
/lgtm

PR #538, opened 2020-10-27
fix(monitoting): bump storage up

georgettica:
according to internal consulting, we want the storage to hold for 15d without filling up

### DIFF
--- a/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -12,14 +12,14 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/infra
           operator: Exists
-      retention: 7d
+      retention: 15d
       volumeClaimTemplate:
         metadata:
           name: prometheus-data
         spec:
           resources:
             requests:
-              storage: 50Gi
+              storage: 100Gi
     alertmanagerMain:
       nodeSelector:
         node-role.kubernetes.io/infra: ""
--- a/deploy/osd-increase-prometheus-storage/10-increase-prometheus-storage.ServiceAccount.yaml
+++ b/deploy/osd-increase-prometheus-storage/10-increase-prometheus-storage.ServiceAccount.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: increase-prometheus-storage
+  namespace: openshift-monitoring
+---
--- a/deploy/osd-increase-prometheus-storage/10-increase-prometheus-storage.ServiceAccount.yaml
+++ b/deploy/osd-increase-prometheus-storage/10-increase-prometheus-storage.ServiceAccount.yaml
@@ -3,4 +3,3 @@ kind: ServiceAccount
 metadata:
   name: increase-prometheus-storage
   namespace: openshift-monitoring
----
--- a/deploy/osd-increase-prometheus-storage/40-increase-prometheus-storage-RoleBinding.yaml
+++ b/deploy/osd-increase-prometheus-storage/40-increase-prometheus-storage-RoleBinding.yaml
@@ -11,4 +11,3 @@ subjects:
 - kind: ServiceAccount
   name: increase-prometheus-storage
   namespace: openshift-monitoring
----
--- a/deploy/osd-increase-prometheus-storage/40-increase-prometheus-storage-RoleBinding.yaml
+++ b/deploy/osd-increase-prometheus-storage/40-increase-prometheus-storage-RoleBinding.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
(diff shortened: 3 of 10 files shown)


deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
```diff
@@ -12,14 +12,14 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/infra
           operator: Exists
-      retention: 7d
+      retention: 15d
```
jewzaam, line 15: With the bump in both storage and retention we are not gaining any headroom on consuming all storage for prometheus, which is the goal with OSD-4875.  Recommend we do not make this change in retention and leave the PR to only the storage change.

```suggestion
      retention: 7d
```

jewzaam (changes requested):

deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
NautiluX: I think @jewzaam suggests to change both, just not in the same PR/promotion

jewzaam: I don't see why we need the retention.  What is the driver to increase retention?

NautiluX: retention has been reduced from 10 to 7 days as a short-term solution to get rid of the alerts. Now we want to increase the retention instead as 7 days of metrics is a short period. With OSD-5221 we did a spike to figure out the right way to go forward with this, got approval for enough storage to hold 15 days, and that's what @georgettica is implementing here. Current config has enough headroom for 7 days, so the assumption is the higher amount should be sufficient for 15 days.

jewzaam: Thought we were sitll hitting storage problems with 7d retention and 50Gi storage, so I disagree that this bump to 14d retention and 100Gi storage fixes the storage issue.  If we want to bump to 10d I am OK with that.

georgettica: What size would be okay on your end? You can see the calculations in the spike we did and devise from there.
I would like to work on both avenues (bigger storage for 14 days or current suggestion for 10) in case one gets blocked (by BU or internally)

jewzaam: Without a use case driving 14d over 10d I'm +1 to 100Gi w/ 10d.  Having over a week is nice and I have felt 10d was a reasonable number after we dropped it from 14d originally.

RiRa12621: I think I commented on the ticket: 2 weeks is the minimum that allows you to do WoW for any day in the last seven days.
For example if you need to compare the behavior of a cluster between two weeks.
As long as we don't federate all metrics off cluster, I think we shouldn't lose that ability on cluster, hence the 14d

jewzaam: I read back through OSD-5221 reasoning.  And considering a low volume of KubePersistentVolumeFullInFourDays alerts showing in Tableau, no hits in telemeter in the recent past, inability to get a result from telemeter for longer time slices (just get a timeout for queries), we're probably OK with the 7d and we can adjust retention down again if necessarily I'll approve.

georgettica:
/hold
a valid concen was given by @jewzaam , need to test before merging

RiRa12621:
Afaik the operator cannot adjust that and we need to manually do it.
/cc s-urbaniak @lilic 
Can you confirm that?

s-urbaniak:
we have no automatic data migration in place if the PVC size changes. If the underlying provider does not support volume expansion internally data needs to be moved manually.

georgettica:
@jewzaam I found out how to do the operation on one cluster.. do you have an idea on how to make this run on all of the fleet?

jewzaam:
@georgettica you can create a cronjob to deploy across the fleet.  There's an SOP on pruning resources but it's the same basic idea: https://github.com/openshift/ops-sop/blob/master/v4/howto/prune-resources.md

georgettica:
@jewzaam the job has been found to not crash and burn, and run the patches.. now I will try to:
- stop syncset
- change prom size and retention
- run the job

and see if it works

georgettica:
@s-urbaniak @jewzaam @RiRa12621 I added a command to manually increase the storage and verified all of the pvc had an option to be scaled up

georgettica:
seems like there is a bug in https://github.com/openshift/managed-cluster-config/blob/796398a7937ecab8496951e11c1c1eaf0138cb7c/scripts/generate_template.py#L79

seems that the `oc patch` is being recognized as a a problem

georgettica:
that wasn't the issue, solved in https://github.com/openshift/managed-cluster-config/pull/559

georgettica:
@jewzaam, it seems that it's fine now.. so I guess it can get a +1?

PR #544, opened 2020-10-27
Change rbac-permissions-operator SSS from Upsert -> Sync

fahlmant:
Restores the mode for the rbac-permissions-operator SSS


mwoodson (approved):
/lgtm

PR #545, opened 2020-10-29
lock down cssre to just rhmi and rhoam namespaces

sclarkso:
@karthikperu7 Here are the changes we discussed.  Please review.

### DIFF
--- a/deploy/backplane/20-cssre.SubjectPermission.yml
+++ b/deploy/backplane/20-cssre.SubjectPermission.yml
@@ -10,7 +10,6 @@ spec:
   permissions:
   - allowFirst: true
     clusterRoleName: backplane-cssre-admins-project
-    namespacesAllowedRegex: .*
-    namespacesDeniedRegex: "(^kube$|^kube-.*|^openshift$|^openshift-.*|^default$)"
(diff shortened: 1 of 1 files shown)


karthikperu7:
Thanks @sclarkso 

/lgtm

deploy/backplane/20-cssre.SubjectPermission.yml
```diff
@@ -10,7 +10,6 @@ spec:
   permissions:
   - allowFirst: true
     clusterRoleName: backplane-cssre-admins-project
-    namespacesAllowedRegex: .*
-    namespacesDeniedRegex: "(^kube$|^kube-.*|^openshift$|^openshift-.*|^default$)"
+    namespacesAllowedRegex: "(^redhat-rhmi-.*|^redhat-rhoam-.*)"
```
sclarkso, line 13: @alanmoran We had a meeting with sre-p to discuss backplane permissions, and these changes were merged for backplane.  I think we'll need to revisit these since you implied we need to move back to the previous 'deny' model?

alanmoran: So the issue is the walkthroughs and fuse afaik. When an RHMI user interacts with them they generate random namespaces based on the user feedback.  I'm not sure we actually need access to these?(do you know?)

eg
image
and
image

PR #551, opened 2020-11-10
SDE-931 + OSD-5587: Enable HSTS on default OSD routes in integration

gsleeman:
Add HSTS annotation to the core OCP routes, initially in int environment for further testing


jaybeeunix:
conceptually looks good to me, but i've asked MCC peeps to verify

PR #553, opened 2020-11-12
Cleanup old credentialsrequest for cloud-ingress-operator

mbarnes:
**IMPORTANT: This must not merge until https://github.com/openshift/cloud-ingress-operator/pull/124 is promoted to production.**

See [OSD-5792](https://issues.redhat.com/browse/OSD-5792)


mbarnes:
This isn't needed because the **selectorsyncset/cloud-ingress-operator** uses `resourceApplyMode: Sync`, which will delete the old **credentialsrequest** and **secret**.

PR #559, opened 2020-11-16
fix(generate_template): ignore empty yamls

georgettica:
to allow people to put multiple '---' inside their yamls, this removal of 'None' was introduced

found when a '---' was added to the end of a yaml and caused CI to complain


RiRa12621 (approved):
/lgtm

PR #560, opened 2020-11-16
feat(rbac-permissions): add secret manipulation

georgettica:
currently WIP as it was not tested on a cluster, will be tested later and will remove WIP
---

as we need to supply to the dedicated admin a way to modify the secrets in the 'openshift-logging' namespace

relevant role/rolebindings were created :)


georgettica:
tested

NautiluX (commented):
/lgtm

NautiluX:
added hold for additional review, remove it once done.

jewzaam:
This is re-inventing rbac for logging, only need to add secret to https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-logging/05-role.yaml 

/approve cancel
/lgtm cancel

georgettica:
didn't see it, thank you for pointing it out!

PR #563, opened 2020-11-23
OSD-3985 configure cluster image-pruner to SRE preferences

mrbarge:
This PR patches the `cluster` ImagePruner in OSD clusters >= `4.6` to have execution preferences matching SRE's existing custom image pruner cronjob. Changes from the `cluster` defaults are:
 
- cronjob will run hourly rather than daily
- cronjob will ignore invalid image refs (default is to not ignore)
- cronjob will keep images younger than a day (default is 1 hour)

This job is being performed by a new SelectorSyncSet that will only apply to clusters that are 4.6+.

The existing `sre-image-pruner` cronjob has also been modified.

- Moved into its own SelectorSyncSet that will only apply to 4.3/4.4/4.5 clusters.
- Changed to run on the `30` minute mark, to avoid colliding with the `cluster` image pruner, something which is presently a source of many PagerDuty alerts during APAC region on 4.6 clusters when the two jobs run simultaneously ([BZ1890828](https://bugzilla.redhat.com/show_bug.cgi?id=1890828)).

Refs: [OSD-3985](https://issues.redhat.com/browse/OSD-3985)

### DIFF
--- a/deploy/sre-pruning/images-pre-4.6/config.yaml
+++ b/deploy/sre-pruning/images-pre-4.6/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: In
+    values: ["4.0","4.1","4.2","4.3","4.4","4.5"]
--- a/deploy/sre-pruning/images/config.yaml
+++ b/deploy/sre-pruning/images/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: NotIn
+    values: ["4.0","4.1","4.2","4.3","4.4","4.5"]
--- a/deploy/sre-pruning/105-pruning.rbac.yaml
+++ b/deploy/sre-pruning/105-pruning.rbac.yaml
@@ -5,19 +5,6 @@ metadata:
   namespace: openshift-sre-pruning
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: sre-pruner-images
-roleRef:
-  kind: ClusterRole
-  name: system:image-pruner
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-- kind: ServiceAccount
-  name: sre-pruner-sa
-  namespace: openshift-sre-pruning
----
-apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: sre-pruner-buildsdeploys-cr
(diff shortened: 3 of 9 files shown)


mrbarge:
@cblecker @jewzaam as OWNERS of the pruner sub-repo, could I please get either of your :eyes: on this?

Ideally I would like to have the existing `sre-image-pruner` cronjob not deployed on 4.6+ clusters at all, but I can't think of a nice way to do that in MCC without moving it into it's own separate SelectorSyncSet with a `clusterDeploymentSelector` matching everything < `4.6`. Which I can do, but figured I'd check with you both first.

cblecker:
I'm okay with moving it to it's own SSS. It can be in a subfolder of the image pruner too (like https://github.com/openshift/managed-cluster-config/tree/master/deploy/sre-prometheus/aws)

deploy/sre-pruning/images-pre-4.6/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: In
+    values: ["4.3","4.4","4.5"]
```
cblecker, line 6: For completeness:
```suggestion
    values: ["4.0","4.1","4.2","4.3","4.4","4.5"]
```

deploy/sre-pruning/images/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: NotIn
+    values: ["4.3","4.4","4.5"]
```
cblecker, line 6: For completeness:
```suggestion
    values: ["4.0","4.1","4.2","4.3","4.4","4.5"]
```

deploy/sre-pruning/images-pre-4.6/00-pruning-images.CronJob.yaml
```diff
@@ -1,4 +1,3 @@
----
```
cblecker, line 1: Should the permissions for the SA be split out too?

deploy/sre-pruning/images-pre-4.6/config.yaml
mrbarge: Thank you! This has been added.

deploy/sre-pruning/images/config.yaml
mrbarge: As above, now added.

deploy/sre-pruning/images-pre-4.6/00-pruning-images.CronJob.yaml
mrbarge: Great catch, I didn't think of that. I've broken out the ClusterRoleBinding to sit within the sub-dir now.

cblecker:
/lgtm
/approve
/hold

@mrbarge Please feel free to `/hold cancel` when you're ready for this to merge (and you're ready to validate that it works on stage on <4.5 and >=4.6

mrbarge:
I'm satisfied that this change is working as expected. More details documented in OSD-3985.
For anyone reviewing this PR retrospectively to see if it's okay for promotion: +1 from me.

PR #564, opened 2020-11-23
Do not allow backplane SREP users to access token/secret sensitive data in the cluster-admin namespace

karthikperu7:
https://issues.redhat.com/browse/OSD-5896

Srep should not be able to access sa tokens (from `openshift-backplane-cluster-admin`) which could potentially allow them to perform elevated actions without auditing/approval/paper-trail. This PR is to block that.


karthikperu7:
cc: @wanghaoran1988 @gsleeman

PR #565, opened 2020-11-23
Patch cluster Ingress.config to enable HTTP/2

fahlmant:
Enable HTTP/2 

Reference [OSD-5521](https://issues.redhat.com/browse/OSD-5521) https://issues.redhat.com/browse/OSD-5521

### DIFF
--- a/deploy/cluster-ingress/00-ingress.config.yaml
+++ b/deploy/cluster-ingress/00-ingress.config.yaml
@@ -1,0 +1,6 @@
+apiVersion: config.openshift.io/v1
+kind: Ingress
+name: cluster
+patch: |-
+  { "metadata": {"annotations": {"ingress.operator.openshift.io/default-enable-http2": "true"} } }
+patchType: merge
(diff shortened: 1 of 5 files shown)


cblecker:
/hold
We can only do this in 4.5+, right?

fahlmant:
@cblecker Yep, but this should hit stage first right?

cblecker:
We can add in restrictions to only deploy to certain versions programatically:

```
selectorSyncSet:
  matchExpressions:
    operator: NotIn
    values: ["4.3","4.4","4.5"]
```

jharrington22:
@cblecker good idea but we want to include 4.5 and make sure we exclude `["4.0","4.1","4.2","4.3","4.4"]` ? 

There is a small list of clusters not at 4.5 which I'll add to the ticket, we should understand why they aren't upgraded.

cblecker:
/lgtm
/approve

Seems sane to me. @fahlmant feel free to `/hold cancel` when you're ready for this to merge to stage

PR #575, opened 2020-11-30
RHOAM RBAC Fixes

alanmoran:
1. Applying correct matchLabel to SSS - `"api.openshift.com/addon-managed-api-service": "true"`
2. Updated layered cluster-admin monitoring check to call out RHMI product.
3. Added new RHOAM cluster-admin monitoring


jewzaam (changes requested):
Very inconsistent name usage.  There's still old "layered" prefix group, some things are "mangaed-api", some things "rhmi" and also "rhoam".

What should this be?  I mean, functionally looks OK but as SREP I am going to have a hard time knowing who to contact if one of these hits on-call.

brizrobbo:
@jewzaam how should it be resolved to be optimal? The reason I am asking as this is building knowledge on how to do this for all layered products that will be coming in the future, and it would be awesome to set the scene with this one.

^^^ That being said, as you say this PR looks "functional", if we can expedite the merging before the end of the week that would be magnificent. Reason: were due to go live with RHOAM on Monday, 7th Dec and if this change is not pushed by then we have a workaround of contacting SREP manually to get the label added to any new clusters being created.

alanmoran:
> Very inconsistent name usage. There's still old "layered" prefix group, some things are "mangaed-api", some things "rhmi" and also "rhoam".

Not sure how much control we have over the inconsistent naming here. Add-on labels and quota SKU's/name of "mangaed-api" were created before the product was named Red Hat Openshift API Management. The only places "mangaed-api" should appear is the two I've mentioned below. 


> What should this be? I mean, functionally looks OK but as SREP I am going to have a hard time knowing who to contact if one of these hits on-call.

Agreed. I've updated the alerts to remove "layered" and replaced with RHMI.

dofinn:
lgtm, will let @jewzaam have final sign off.

PR #576, opened 2020-11-30
Hide kube:admin IDP

cblecker:
Came across https://www.openshift.com/blog/enhancing-the-openshift-web-console-login-experience, and it seems simple and straight forward.

This could also be used to hide the OpenShift_SRE IDP if we have sufficient tooling (or once backplane arrives).

fixes [OSD-5768](https://issues.redhat.com/browse/OSD-5768)


cblecker:
/assign @jewzaam 
/hold
for review

source/html/providers.html
```diff
@@ -27,16 +27,18 @@
...
               <ul>
                 {{ range $provider := .Providers }}
-                  <li class="idp">
-                    <a href="{{$provider.URL}}" class="pf-c-button pf-m-secondary pf-m-block" title="Log in with {{$provider.Name}}">{{$provider.Name}}</a>
-                  </li>
+                  {{ if ne $provider.Name "kube:admin" }}
+                    <li class="idp">
+                      <a href="{{$provider.URL}}" class="pf-c-button pf-m-secondary pf-m-block" title="Log in with {{$provider.Name}}">{{$provider.Name}}</a>
+                    </li>
+                  {{ end }}
```
jewzaam, line 34: Source for this in the [README](https://github.com/openshift/managed-cluster-config/tree/master/source/html) is https://openshift.github.io/oauth-templates/od/providers.html which doesn't include this.  Recommend we get that updated upstream.

jewzaam (changes requested):

source/html/providers.html
cblecker: Added follow up item to upstream this fix

PR #578, opened 2020-12-01
feat(osd-devaccess): allow project creation

georgettica:
this will allow project creation for must gathers purposes


dofinn:
@georgettica do we need to add delete see dev-access can clean up after themselves? They should only be able to delete namespaces they created.

cblecker:
@dofinn I believe privs are automatically granted to namespaces you own to delete. I don't think we need to add extra.

georgettica:
I agree with @cblecker , when a project is created.by a user he receives an Admin role to it, so he can delete it when he is done

georgettica:
closing as a solution was devised in #584

PR #586, opened 2020-12-08
[WIP] OSD-5666

robotmaxtron:
Enabling user-workload monitoring on 4.6 clusters


deploy/cluster-monitoring-config/openshift-user-workload-monitoring.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: user-workload-monitoring-config
+  namespace: openshift-user-workload-monitoring
```
jewzaam, line 5: Why are we configuring this instead of providing RBAC for dedicated-admins to set this up?  Not all customers will want this running on their cluster.  I think the thing we need to do is simply set enableUserWorkload: true

jewzaam (changes requested):

deploy/cluster-monitoring-config/openshift-user-workload-monitoring.yaml
fahlmant: `enableUserWorkload: true` will deploy this either way. This configmap definition is just to ensure all the components run on worker nodes only. If we don't want to automatically deploy UWM for all 4.6+ clusters, we'll need to fully close the PR and gate it behind an ocm flag.

jewzaam: Talked on slack.  Reason for this CM was to put the prom operator pod on worker (by default is deployed to masters).  But also brought up that if we manage the CM via SSS it cannot be modified by customers.  So it needs to be pulled out.

fahlmant:
@robotmaxtron Looks good to me. Mind squashing?

deploy/cluster-monitoring-config/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: hive.openshift.io/version-major-minor
+      operator: In
+      values: ["4.6"]
```
jewzaam, line 6:
```suggestion
      values: ["4.6","4.7,"4.8"]
```

Just so we don't have to come back to this later.  At 4.8+ policy will be we don't support versions of OCP on OSD that don't support UWM and this can condition be removed.

jewzaam (changes requested):

robotmaxtron:
see - https://github.com/openshift/managed-cluster-config/pull/587

PR #593, opened 2020-12-10
Adding in uwm for stg

robotmaxtron:
Re-enabling UWM as part of [OSD-5666](https://issues.redhat.com/browse/OSD-5666), set to WIP in case this is not something we want to enable right away.


fahlmant (approved):
/lgtm

PR #596, opened 2020-12-11
MUO external dep config

dofinn:
Related to https://github.com/openshift/managed-upgrade-operator/pull/183
to support configuration of external dependency of which MUO now does
availability checks on prior to upgrading.


deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
```diff
@@ -33,6 +33,12 @@ data:
       - MetricsClientSendFailingSRE
       - UpgradeNodeScalingFailedSRE
       - UpgradeClusterCheckFailedSRE
+    extDependencyAvailabilityChecks:
+      http:
+        timeout: 10
+        urls:
+          - https://quay.io/health
+          - https://api.openshift.com
```
mrbarge, line 41: This could better be represented as 
```suggestion
          - ${OCM_BASE_URL}
```

Staging MUOs should care about the staging API health, not Prod (etc..)

PR #600, opened 2020-12-14
OSD-5751 Suppress HAProxyReload OCP alert for SREP alert

dofinn:
HAProxyReload fails is currently a self resolving NO-OP. This PR creates
an SREP implemenation of the alert that we can tune easily
if required while we await backporting of the fix.

The current alert will be sent to nullReceiver as per:
https://github.com/openshift/configure-alertmanager-operator/pull/125


karthikperu7:
/lgtm 
for the prom rule change

There is also a change to the OWNERS file under sre-prometheus that needs to be given +1 from someone already in it.

PR #603, opened 2020-12-16
Quick fixes for broken CSSRE rbac and a missing NS for SA cronjob

jewzaam:
The clusterrole for CSSRE was renamed with backplane work.  You cannot change roleRef in an clusterrolebinding, so this broken existing cluster RBAC.  The fix is simply rename the clusterrolebinding which will delete the old one and create the new one.  It will use a unique name.

The namespace used by SA cronjob was removed.  Added it back to a SSS w/ Upsert mode so it can safely be deleted in the future.


mwoodson (approved):
/lgtm

alanmoran:
lgtm

PR #607, opened 2020-12-21
Fix sre-build-test toleration of infra taints

mrbarge:
Staging clusters now have infra nodes with a `NoSchedule` taint instead of `PreferNoSchedule`. [0]

Since this change was implemented, the `sre-build-test` job has not been running on staging clusters as it can't find a node to schedule to, so it stays permanently `Pending`. Its toleration settings for infra nodes do not look correct. This PR addresses the `toleration` of the `sre-build-test` cronjob so that it correctly tolerates the Infra node taint.

I've successfully tested this on a staging cluster and verified scheduling to infra nodes.

refs [OSD-2927](https://issues.redhat.com/browse/OSD-2927); cc @jewzaam 

[0] https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2008

### DIFF
--- a/deploy/sre-build-test/50-build-test.CronJob.yaml
+++ b/deploy/sre-build-test/50-build-test.CronJob.yaml
@@ -18,14 +18,11 @@ spec:
               requiredDuringSchedulingIgnoredDuringExecution:
                 nodeSelectorTerms:
                 - matchExpressions:
-                  - key: "node-role.kubernetes.io"
-                    operator: In
-                    values:
-                      - infra
+                  - key: "node-role.kubernetes.io/infra"
+                    operator: Exists
           tolerations:
-          - operator: Equal
-            key: node-role.kubernetes.io
-            value: infra
+          - operator: Exists
+            key: node-role.kubernetes.io/infra
             effect: NoSchedule
           serviceAccountName: sre-build-test
           restartPolicy: Never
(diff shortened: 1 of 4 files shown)


hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -5242,7 +5242,7 @@ objects:
                 restartPolicy: Never
                 containers:
                 - name: osd-logging-freeze
-                  image: quay.io/openshift/origin-cli:4.7.0
```
cblecker, line 5245: Is this a regeneration error? I don't see a matching change for this

cblecker: Oh, looks like 1d7331cddb7881a8a8bd592467cf1edece9350dc

mrbarge: That's puzzled me too - I'm not sure why it's changing that image URL. It happens when I run `make` straight off master as it currently exists, too.

deploy/sre-build-test/50-build-test.CronJob.yaml
```diff
@@ -23,9 +23,8 @@ spec:
                     values:
                       - infra
```
cblecker, line 24: This needs to change too

mrbarge: Thanks for spotting - now fixed.

cblecker:
As https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2008 merged, this test is broken on stage. Merging this so we can fix this on stage.

PR #608, opened 2020-12-21
Fall back to docker hub image

cblecker:
As the `quay.io/app-sre/python` image is private, this provides for the ability to fall back to the docker hub image if it's unable to pull the app-sre image.

### DIFF
--- a/.github/workflows/makecommit.yml
+++ b/.github/workflows/makecommit.yml
@@ -1,21 +1,21 @@
 name: Make Generate
 on: push
 
-jobs: 
+jobs:
   run:
     name: Make Generate Template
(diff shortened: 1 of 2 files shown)


lisa:
@cblecker Are the images bit-for-bit the same?

cblecker:
@lisa Yes, as I believe app-SRE mirrors the image via app-interface.

lisa:
@cblecker "LGTM" but let's wait til January and things thaw out unless there's some more urgent issue forcing our hand.

cblecker:
GitHub actions are working now: https://github.com/cblecker/managed-cluster-config/actions/runs/464629787

PR #609, opened 2021-01-04
Don't back up unnecessary objects

cblecker:
Events are unnecessary for backup and create a large number of API calls. This will both shrink and speed up backups.

List based on https://github.com/konveyor/mig-operator/blob/master/docs/usage/ExcludeResources.md


mbarnes (approved):
Taking your word for it on some of the list items not mentioned in [ExcludeResources.md](https://github.com/konveyor/mig-operator/blob/master/docs/usage/ExcludeResources.md), but otherwise :+1:.

PR #614, opened 2021-01-13
remove quick fix to add backplane-srep ns

karthikperu7:
Removing the backplane-sre ns now that MCC with backplane prod RBAC files has been promoted.
As per : https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-serviceaccounts/README.md


karthikperu7:
/hold 

for @jewzaam 's review

PR #615, opened 2021-01-13
OSD-5555: Change maxUnhealthy to 2 so MHC works in 2-worker clusters

feichashao:
Currently, in MHC maxUnhealthy=40%, this will lead to two problems:
- MHC won't work for 2-infra/2-worker clusters, as floor(2*40%)=0.
- 40% is too large for MHC to short-circuit. Eg. for a 3-AZ cluster, if one AZ (1/3 nodes) has network connectivity issue, MHC will run into a hot loop which keep re-creating instances in that AZ.

With change to maxUnhealthy=2,
- MHC can work for 2-infra/2-worker clusters.
- MHC can short-circuit early to prevent hot loop.
- Con: In case there's an AZ issue that brings all instances down and instances remain in stopped state after the AZ issue (rare), MHC won't reconcile and need human to bring them up.

### DIFF
--- a/deploy/osd-machine-api/010-machine-api.srep-infra-healthcheck.MachineHealthCheck.yaml
+++ b/deploy/osd-machine-api/010-machine-api.srep-infra-healthcheck.MachineHealthCheck.yaml
@@ -14,4 +14,4 @@ spec:
   - type:    "Ready"
     timeout: "480s"
     status: "Unknown"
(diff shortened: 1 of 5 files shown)


feichashao:
@yithian  @cblecker @luis-falcon could you help to review this?

cblecker (approved):

PR #618, opened 2021-01-14
Jira OSD-6173: deploy/osd-logging: prepare for logging 4.5+

RiRa12621:
Logging has changed it's structure. The elasticsearch operator is expected to live in `openshift-operators-redhat`. We don't want to allow dedicated admins to create `openshift-*` namespaces, so we have to create it as well as the rules accordingly, so they can add the subscription and an `elasticsearch` instance.
The elasticsearch cluster itself however will stay in `openshift-logging` by design

### DIFF
--- a/deploy/osd-logging/05-role.yaml
+++ b/deploy/osd-logging/05-role.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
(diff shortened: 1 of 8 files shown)


jewzaam (changes requested):

RiRa12621:
/hold
need to figure out single vs multinamespace

deploy/osd-logging/05-role.yaml
```diff
@@ -66,3 +67,50 @@ rules:
...
+  - clusterserviceversions
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - events
+  - namespaces
+  - persistentvolumeclaims
+  - persistentvolumes
```
georgettica, line 110: why do they need PV permissions?
does this acutally do something as a Role?

georgettica (commented):
love that you are OWNER to this folder :)

deploy/osd-logging/05-role.yaml
RiRa12621: it needs to be able to list the PVs if i understood correctly and those can be in this namespace or in the other

georgettica: Iirc pv are namespace agnostic, so putting it here might be confusing (or at least it was for me)

RiRa12621:
/retitle Jira OSD-6173: deploy/osd-logging: prepare for logging 4.5+

deploy/osd-logging/05-role.yaml
RiRa12621: PVs are in fact namespace agnostig, PVCs are not.
additionally that role needs the privileges anyways so it doesn't matter if they are namespace scoped.

georgettica: I don't think it makes sense to put them in the role.. what about putting them in the clusterrole?

RiRa12621: Which one?

georgettica: The pv

RiRa12621: No, in which clusterrole

georgettica: I didn't see one.. maybe create one for it in this folder?

It's separates the roles but makes.it clean it's a cluster wide resource

RiRa12621:
/close
Obsolete

PR #619, opened 2021-01-14
Add components for observatorium token refresher

arjunrn:
As part of the epic "Observatorium Tenant for OSD" we need to deploy a [token injecting proxy](https://github.com/observatorium/token-refresher) into all clusters. This is a simply proxy which simply adds credentials into the HTTP request which is send to the observatorium tenant. The credentials for the proxy are already present on the hive clusters. So the following components need to be deployed through the managed configuration:
1. `SelectorSyncSet` which mirrors the contents of the secret into all clusters. 
2. A `Deployment` and corresponding `Service` for the proxy which mounts the secret and injects them into the requests.

In order to do a gradual rollout the `SelectorSyncSet`s are targetted only at non-production clusters. Once the metric ingestion is confirmed and working correctly the deployment will be rolled out to production clusters as well.

cc @dofinn

### DIFF
--- a/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
+++ b/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
@@ -1,0 +1,52 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: authentication-proxy
+    app.kubernetes.io/name: token-refresher
+    app.kubernetes.io/version: master-2020-12-04-5504078
+  name: token-refresher
+  namespace: openshift-monitoring
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: authentication-proxy
+      app.kubernetes.io/name: token-refresher
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: authentication-proxy
+        app.kubernetes.io/name: token-refresher
+        app.kubernetes.io/version: master-2020-12-04-5504078
+    spec:
+      containers:
+      - args:
+        - --oidc.audience=observatorium-telemeter
+        - --oidc.client-id string=\$\(CLIENT_ID\)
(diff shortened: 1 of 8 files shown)


2uasimojo:
> As part of the epic "Observatorium Tenant for OSD"

Can you link the epic and/or specific card for this?

You'll at least need to do the regenerate thing (I think you just run `make`) and check in the resulting hack/osd-managed-cluster-config-*.yaml.tmpl files.

I'll take a closer look here in a minute.

2uasimojo (commented):
Will this need RBAC?

deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: authentication-proxy
+        app.kubernetes.io/name: token-refresher
+        app.kubernetes.io/version: master-2020-12-04-5504078
+    spec:
+      containers:
+      - args:
+        - --oidc.audience=observatorium-telemeter
```
dofinn, line 25: Is this the service account? in which case it should be observatorium-osd?

dofinn:
@2uasimojo the related card is https://issues.redhat.com/browse/OSD-5128  

@arjunrn please run make so the `hack/` directories are generated. 

Also I thought we were deployed https://github.com/openshift/kube-rbac-proxy within the same pod as the token refresher for RBAC?

deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
arjunrn: No this is would be the [audience field](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) for the requested OIDC token.

arjunrn:
@dofinn So it looks like `NetworkPolicy` will do the same job here as `kube-rbac-proxy`. There could be installations of kubernetes/openshift where network policies are not available and in such a case the proxy could be used as a side car.

PR #625, opened 2021-01-19
Disable h2 on hive

wanghaoran1988:


jewzaam (changes requested):
This won't do the intended work.  This will stop patching ingress but it won't change what is already patched.  Add a new deploy/cluster-ingress-hive that does the inverse, patching http2 to be false.

PR #630, opened 2021-01-20
Fix token-refresher deployment with correct escaping and flag name

arjunrn:
Followup to #619 . Fixed the incorrect escaping of interpolated environment variables in the token refresher deployment.


lisa (approved):
/lgtm
/approve

PR #632, opened 2021-01-21
Add SSS to patch candidate and fast channels in ClusterVersion

jewzaam:
https://issues.redhat.com/browse/OSD-6205

### DIFF
--- a/deploy/osd-channel-patch/README.md
+++ b/deploy/osd-channel-patch/README.md
@@ -1,0 +1,13 @@
+https://issues.redhat.com/browse/OSD-6205
+
+When OCP installs a cluster it sets the ClusterVerison channel to the stable for that major.minor.  So if you install a cluster using an image available only in a candidate channel the in-cluster channel is still on stable.  No upgrades will be listed available until that cluster's version is in the stable channel OR the in-cluster channel is updated.
+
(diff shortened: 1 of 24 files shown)


jewzaam:
/hold

we have some stuff being promoted, just holding to make sure I don't muddy the promotion with this

dofinn:
I thought only candidate channels were impacted by this ?

mrbarge:
> I thought only candidate channels were impacted by this ?

It happens on clusters installing a `fast` version too, unfortunately. It's the OCP installer doing it, it just defaults clusters to be on `stable` post-install.

The change looks fine to me. I'm a little worried that we have both an SSS and MUO trying to update that field using a different part of CS as their source of truth (MUO uses `upgrade_channel_group`, SSS uses the `clusterdeployment` label).. although I've been wracking my brain to think of potential ways it could break and can't think of any scenarios. MUO only sets the field at upgrade time, and at that point in time the clusterdeployment's label should already have a matching value to that in CS's `upgrade_channel_group`.

jewzaam:
@mrbarge same re: two sources to set it.  I tried to lay out my thoughts in the readme.  I'll update with specific scenarios since I still do fear there is something overlooked :)

NautiluX (commented):
/lgtm

jewzaam:
/hold cancel

Removed hold since I plan to test shortly after it lands.  And there isn't any pending promotion this could hold up.

PR #633, opened 2021-01-22
Tuned thanosQuerier cpu limit so avoid throttle

superfunhappytime:
We found that the new limit for cpu on thanosQuerier was tripping up the CPUHighThrottle alert (approx 30-35% throttling) - Tuned up the limit to allow a little room


cblecker (approved):

superfunhappytime:
Sorry, I missed UWM thanosQuerier limit

PR #636, opened 2021-01-25
[deploy|hack] OSD-5906 Enabling samesite cookie (strict) on default haproxy route… 

treymer:
…s in integration

According to [Route-specific annotations](https://docs.openshift.com/container-platform/4.6/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration) `haproxy.router.openshift.io/cookie-same-site` as `Strict` is similar in behavior to `None`, but None allows for cross-origin requests.

Three options are as follows:
`Lax`: cookies are transferred between the visited site and third-party sites.
`Strict`: cookies are restricted to the visited site.
`None`: cookies are restricted to the visited site.
See also: [SameSite cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict)

Followed the [PR](https://github.com/openshift/managed-cluster-config/pull/551/files) as a boiler plate.


treymer:
Still a WIP, do not merge.

dofinn:
@treymer do we need to include oauth route ? I see mention of it in JIRA but no definitive outcome?

treymer:
@dofinn Since the oauth route is a passthrough route we cannot set cookies on it. In detail, "Cookies cannot be set on passthrough routes, because the HTTP traffic cannot be seen. Instead, a number is calculated based on the source IP address, which determines the back-end".

treymer:
Per this bugzilla you must annotate the router.openshift.io/cookie-same-site value. This is what the router is looking for as well https://github.com/openshift/router/blob/master/pkg/router/template/configmanager/haproxy/manager.go#L1108

PR #637, opened 2021-01-25
Add workaround for BZ1915081

cblecker:
Add workaround to ensure the node-ca daemonset is patched with the correct toleration. This will be a noop on most clusters, and matches the intended config from the image-registry-operator.

part of [OSD-2927](https://issues.redhat.com/browse/OSD-2927)


cblecker:
Confirmed working in stage

PR #639, opened 2021-01-27
OHSS-2561 Extend hive integration token timeout

mrbarge:
This PR extends the authentication token timeout for Hive integration clusters to a week.

The change has been successfully tested (in `SyncSet` form) on a test cluster to verify correct syntax of the patch.

Refs: [OHSS-2561](https://issues.redhat.com/browse/OHSS-2561) where this has been discussed.

### DIFF
--- a/deploy/osd-oauth-templates/ohss-2561/config.yaml
+++ b/deploy/osd-oauth-templates/ohss-2561/config.yaml
@@ -1,3 +1,9 @@
-deploymentMode: "Direct"
-direct:
-  environments: ["integration"]
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: environment
+    operator: In
+    values: ["integration"]
+  - key: ext-managed.openshift.io/hive-shard
+    operator: In
+    values: ["true"]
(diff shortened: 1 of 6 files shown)


mrbarge:
Want some :eyes: on this to make sure I've done it right, so just putting a hold on until then.
/hold

deploy/osd-oauth-templates/ohss-2561/config.yaml
```diff
@@ -0,0 +1,3 @@
+deploymentMode: "Direct"
+direct:
+  environments: ["integration"]
```
jewzaam, line 3: This is an interesting case I didn't consider when setting up Direct.  The yaml for the "patch" doesn't apply if you use `oc apply -f` but you can apply the patch via the hive managing integration:

```suggestion
deploymentMode: "SelectorSyncSet"
selectorSyncSet:
  matchExpressions:
  - key: environment
    operator: In
    values: ["integration"]
  - key: ext-managed.openshift.io/hive-shard
    operator: In
    values: ["true"]
```

jewzaam (changes requested):

jewzaam:
/approve

Once this is sorted out, I'm good with the change.

PR #645, opened 2021-01-29
Apply UWM to clusters without a uwm-disabled label

fahlmant:
This will roll out UWM to all 4.6+ clusters, excluding any that have a `uwm-diabled` label applied to the ClusterDeployment.

NOTE: This is a spike, not meant to go to prod, as an option for UWM roll out

### DIFF
--- a/deploy/cluster-monitoring-config-uwm/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config-uwm/cluster-monitoring-config.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-monitoring
 data:
   config.yaml: |
+    enableUserWorkload: true 
     prometheusK8s:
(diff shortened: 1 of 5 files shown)


lisa:
@fahlmant Is there a Jira card to which this relates? If so, can you link it in the initial comment?

fahlmant:
There is not. This is for a spike based off a conversation in the UWM rollout doc

NautiluX:
/approve
As said on chat, I think it would be cleaner to add a second label to ensure the change doesn't enable UWM on all clusters when promoted for whatever reason, but I get you plan to revert it before it gets promoted.

jharrington22:
/lgtm

In the future could we have used the `environments` key in the config to restrict this sync set to only ClusterDeployments labeled "stage"?

PR #648, opened 2021-02-01
OHSS-2561 revise targeting of hive integration clusters

mrbarge:
This PR updates the targeting of Hive integration clusters for [OHSS-2561](https://issues.redhat.com/browse/OHSS-2561) to be based on a list of cluster IDs.

PR #639 was intended to target any Hive integration clusters through `clusterdeployment` labelling; it did not succeed because Hive integration clusters are production clusters, and so their `environment` label is set to `production`. There does not appear to be any way in a `clusterdeployment` to target any/all Hive integration clusters, so the fallback option is a list of cluster IDs.

As the number of Hive integration clusters is hopefully low, the maintenance overhead on this should be low until the change can eventually be replaced by an OCM-configurable way of managing this configuration.

### DIFF
--- a/deploy/osd-oauth-templates/ohss-2561/config.yaml
+++ b/deploy/osd-oauth-templates/ohss-2561/config.yaml
@@ -1,9 +1,9 @@
 deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   matchExpressions:
-  - key: environment
+  - key: api.openshift.com/id
     operator: In
-    values: ["integration"]
+    values: ["1g268u7pp694gj152nj16me4sv615lpv"]  # any Hive integration cluster IDs
   - key: ext-managed.openshift.io/hive-shard
     operator: In
     values: ["true"]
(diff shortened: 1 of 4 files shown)


mrbarge:
@jewzaam could I please request your :eyes: on this? Want to make sure I'm not violating any MCC tenets by targeting a clusterID.

jewzaam:
@mrbarge we use ext- labels to add `ext-managed.openshift.io/hive-shard`, same could be done to add a new label `ext-managed.openshift.io/hive-environment` instead of pinning to an ID.  Our SOP around this is weak.  I see it in https://github.com/openshift/ops-sop/blob/master/v4/setup/velero.md#hive-specific but it should be in `v4/knowledge_base/shard-guide.md` I think.  NOTE App SRE setup of hive shards includes this labeling afaik, so the shard bring-up process would need updated.

jewzaam:
Sorry, that being said, short term I'm OK to put this in as long as we track and link here that we're doing something to generalize.

/approve
/hold

PR #650, opened 2021-02-01
Enable deployment of operators in the openshift-operators-redhat NS

jewzaam:
https://issues.redhat.com/browse/OSD-6184

PR #649, opened 2021-02-01
Disable in-cluster logging on ROSA and OSD 4.8+

jewzaam:
https://issues.redhat.com/browse/OSD-6322
https://issues.redhat.com/browse/OSD-6325

PR #652, opened 2021-02-01
Revert osd logging stuff

jewzaam:
We hit API limits for downloading the templates.  Reverting until we can sort that out.

Caused by: https://issues.redhat.com/browse/APPSRE-2959


lisa:
Linking to PRs this reverts: #651, #650, #649.

PR #655, opened 2021-02-02
Delete backplane serviceaccounts after expiry time

sedroche:
Jira: https://issues.redhat.com/browse/OSD-5914

Backplane leaves serviceaccounts indefinitely atm but they should expire. This cronjob removes them after an expiry time.

### DIFF
--- a/deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
+++ b/deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
@@ -1,0 +1,47 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: openshift-backplane
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-delete-serviceaccounts
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-get-namespace
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+    verbs:
+      - get
+      - list
+---
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+    - osd-get-namespace
+  permissions:
+    - allowFirst: true
+      clusterRoleName: osd-delete-serviceaccounts
+      namespacesAllowedRegex: "(^openshift-backplane-.*)"
+  subjectKind: User
+  subjectName: system:serviceaccount:openshift-backplane:osd-delete-backplane-serviceaccounts
--- a/deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
+++ b/deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
@@ -7,7 +7,7 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: osd-delete-serviceaccounts
+  name: osd-delete-backplane-serviceaccounts
 rules:
   - apiGroups:
       - ""
@@ -41,7 +41,7 @@ spec:
     - osd-get-namespace
   permissions:
     - allowFirst: true
-      clusterRoleName: osd-delete-serviceaccounts
+      clusterRoleName: osd-delete-backplane-serviceaccounts
       namespacesAllowedRegex: "(^openshift-backplane-.*)"
   subjectKind: User
   subjectName: system:serviceaccount:openshift-backplane:osd-delete-backplane-serviceaccounts
--- a/deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml
+++ b/deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml
@@ -1,0 +1,34 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: openshift-backplane
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          serviceAccountName: osd-delete-backplane-serviceaccounts
+          containers:
+            - name: osd-delete-backplane-serviceaccounts
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  for ns in $(oc get ns --no-headers | grep -e ^openshift-backplane-.* | awk '{print $1}');do
+                      for sa in $(oc get sa -n $ns -l managed.openshift.io/backplane=true --no-headers | awk '{print $1}');do
+                        creation=$(date -d $(oc get sa $sa -n $ns -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                        now=$(date +%s);
+                        expiryMins=$(oc get sa $sa -n $ns -o jsonpath='{.metadata.annotations.managed\.openshift\.io/backplane-expiry-duration}');
+                        if [[ $(expr $((now-creation)) \> ${expiryMins:=720} \* 60) == 1 ]]; then
+                          oc delete sa $sa -n $ns
+                        fi
+                      done;
+                  done;
--- a/deploy/osd-delete-backplane-serviceaccounts/00-delete-backplane-serviceaccounts.namespace.yml
+++ b/deploy/osd-delete-backplane-serviceaccounts/00-delete-backplane-serviceaccounts.namespace.yml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-backplane
--- a/deploy/osd-delete-backplane-serviceaccounts/config.yaml
+++ b/deploy/osd-delete-backplane-serviceaccounts/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: api.openshift.com/environment
+      operator: NotIn
+      values: ["production"]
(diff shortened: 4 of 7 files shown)


sedroche:
@wanghaoran1988 @karthikperu7

wanghaoran1988:
This looks good to me, @jewzaam Could you take another look?

deploy/osd-delete-backplane-serviceaccounts/00-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: osd-delete-serviceaccounts
+subjects:
+  - kind: ServiceAccount
+    name: osd-delete-backplane-serviceaccounts
+    namespace: backplane
```
jewzaam, line 32: Should limit to backplane namespaces only.  Can use SubjectPremissions to manage this so it is based on regex.

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
```diff
@@ -0,0 +1,34 @@
...
+          serviceAccountName: osd-delete-backplane-serviceaccounts
+          containers:
+            - name: osd-delete-backplane-serviceaccounts
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  for sa in $(oc get sa -A -l managed.openshift.io/backplane=true --no-headers | awk '{print $1 ":" $2}');do
```
jewzaam, line 25: When RBAC is limited to backplane namespaces I expect this needs to use the same regex to look for SA only in those same namespaces.

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
```diff
@@ -0,0 +1,34 @@
...
+                    name=$(echo $sa | cut -d':' -f2);
+                    namespace=$(echo $sa | cut -d':' -f1);
+                    creation=$(date -d $(oc get sa $name -n $namespace -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                    now=$(date +%s);
+                    expiryMins=$(oc get sa $name -n $namespace -o jsonpath='{.metadata.annotations.managed\.openshift\.io/backplane-expiry-duration}');
+                    if [[ $(expr $((now-creation)) \> ${expiryMins:=120} \* 60) == 1 ]]; then
+                      oc delete sa $name -n $namespace
+                    fi
```
jewzaam, line 33: What if I'm using that SA still?  2 hours isn't that long if there's an incident.  I expect I have to log in again.  And if backplane cli were to automate that any active port forwards like Prometheus need restarted.  Backplane is proxying all requests.  Could backplane manage deletion based on the timestamp for last time the SA was used?

jewzaam (changes requested):
/hold

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
wanghaoran1988: backplane is not caching the cluster metadata until it's being used, instead of 2 hours, maybe we can set to 12h, or 24h.

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
sedroche: SubjectPermission and filtering to openshift-backplane ns added

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
jewzaam: I like that, 12h may be a good number.  That way it should expire while I'm offline assuming a reasonable work day, even if I created the SA at the beginning or end of the work day.

jewzaam: Maybe a INFO output on ocm backplane login about when the SA expires?  it could be simple math based on the creation timestamp but requires both the cli and the job here to be on the same schedule.  small risk of drive, account for in good comments.

sedroche:
Have updated this to 12 hours. Missed the perms for listing namespaces after the change to subjectpermissions so I have added that as well

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.CronJob.yaml
sedroche: I'll create a card for this request

deploy/osd-delete-backplane-serviceaccounts/00-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,58 @@
...
+metadata:
+  name: osd-backplane-namespaces
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: osd-get-namespace
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: system:serviceaccount:backplane:osd-delete-backplane-serviceaccounts
```
jewzaam, line 58: Not going to hold up the PR for it, but FYI you can manage global perms with SubjectPermission as well, see `.spec.clusterPermissions`.  But it does what this is doing so :+1: to what you have.

jewzaam (approved):

sedroche:
@jewzaam I updated to use subjectpermssions for the clusterrolebinding. Needs one more `lgtm` to get this in

deploy/osd-delete-backplane-serviceaccounts/00-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: backplane
```
wanghaoran1988, line 5: This namespaces does not exists on all OSD clusters, and we should create one first, per the name, we should have a "openshift-" prefix.

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: openshift-backplane
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-delete-serviceaccounts
```
jewzaam, line 10:
```suggestion
  name: osd-delete-backplane-serviceaccounts
```

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,47 @@
...
+kind: SubjectPermission
+metadata:
+  name: osd-delete-backplane-serviceaccounts
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+    - osd-get-namespace
+  permissions:
+    - allowFirst: true
+      clusterRoleName: osd-delete-serviceaccounts
```
jewzaam, line 44:
```suggestion
      clusterRoleName: osd-delete-backplane-serviceaccounts
```

deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml
```diff
@@ -0,0 +1,34 @@
...
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  for ns in $(oc get ns --no-headers | grep -e ^openshift-backplane-.* | awk '{print $1}');do
+                      for sa in $(oc get sa -n $ns -l managed.openshift.io/backplane=true --no-headers | awk '{print $1}');do
+                        creation=$(date -d $(oc get sa $sa -n $ns -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                        now=$(date +%s);
+                        expiryMins=$(oc get sa $sa -n $ns -o jsonpath='{.metadata.annotations.managed\.openshift\.io/backplane-expiry-duration}');
```
jewzaam, line 29: Nice touch, gives us flexibility to manage in one place :+1:

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
```diff
@@ -0,0 +1,47 @@
...
+      namespacesAllowedRegex: "(^openshift-backplane-.*)"
+  subjectKind: User
```
jewzaam, line 46:
```suggestion
  subjectKind: ServiceAccount
```

jewzaam (changes requested):

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
sedroche: rbac operator doesn't currently handle type of `ServiceAccount`
https://github.com/openshift/rbac-permissions-operator/blob/f0cfa43269218d756876792834ecd6c1209d8b01/pkg/controller/utils/controllerutil.go#L105-L109

The rolebinding definition for a sa is:
```yaml
subjects:
- kind: ServiceAccount
  name: osd-delete-backplane-serviceaccounts
  namespace: openshift-backplane
```

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
sedroche: Have update name here and in subjectpermisison

deploy/osd-delete-backplane-serviceaccounts/10-delete-backplane-serviceaccounts.rbac.yaml
sedroche: @jewzaam This is the last issue on this PR. Seem ok?

jewzaam: Yea, if User works then OK.  If not we'll know in stage.

sedroche:
@christoph Can I get a `lgtm` for this. All feedback has been applied now

dofinn:
Lets keep an on eye on `User` in subjectPermission CRs

PR #657, opened 2021-02-02
Initial comment to change logging support.

jewzaam:
Make sure we don't delete resources when moving things out of SSS osd-logging

https://issues.redhat.com/browse/OSD-6324


jharrington22 (approved):
/lgtm

PR #659, opened 2021-02-02
Add Infra Node Imbalance SRE Alert

dustman9000:
This adds an SRE alert to detect when infra nodes become imbalanced (due to the timing of node readiness during install time). The promQL is to count the pods based on the name (`prometheus-k8s-*`) per node. If the node count does not match the infra node count, then this will trigger an alert.

Related SOP PR for link: https://github.com/openshift/ops-sop/pull/1185

### DIFF
--- a/deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-infra-node-imbalance
+    role: alert-rules
+  name: sre-infra-node-imbalance
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-infra-node-imbalance
+    rules:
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
```diff
@@ -0,0 +1,21 @@
...
+    prometheus: sre-infra-node-imbalance
+    role: alert-rules
+  name: sre-infra-node-imbalance
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-infra-node-imbalance
+    rules:
+    - alert: KubeInfraNodeImbalanceSRE
+      expr: count(count by (node) (kube_pod_info{namespace="openshift-monitoring",pod=~"prometheus-k8s-.*"})) != count(cluster:infra_nodes)
```
jharrington22, line 14: So 2 pods on 1 node would = 1?

dustman9000: Yes

jharrington22 (approved):
/lgtm

RiRa12621:
Where's the upstream feedback for this?

PR #654, opened 2021-02-02
Update RBAC for osd-cluster-ready

2uasimojo:
Real health checks have been [added to osd-cluster-ready-job](https://github.com/iamkirkbater/osd-cluster-ready-job/pull/1). This commit updates the RBAC to make that work in cluster.

It also changes the Job's `restartPolicy` to keep trying on failures.

PR #663, opened 2021-02-02
osd-cluster-ready: restartPolicy: OnFailure

2uasimojo:
Change the `restartPolicy` for osd-cluster-ready to `OnFailure` so that it will restart if errors occur. Remove the overrides for `backoffLimit` and `activeDeadlineSeconds` to let the defaults for those take over.

(Note that this is not really related to the maximum cluster age observed by the job. The job only fails if an error occurs, not if health checks are unsuccessful. They tie together only in the sense that, if the job restarts after one or more failures, it still uses the inception of the cluster (not the job) to determine the maximum age.)

### DIFF
--- a/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
+++ b/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
@@ -4,8 +4,6 @@ metadata:
     name: osd-cluster-ready
     namespace: openshift-monitoring
 spec:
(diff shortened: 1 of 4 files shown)


2uasimojo:
> Rebase after #654 lands.

Done.

But continue to hold pending [this discussion](https://github.com/iamkirkbater/osd-cluster-ready-job/pull/2#issuecomment-772614083).

PR #665, opened 2021-02-02
Update Infra Node Imbalance Alert For Multi-Az

dustman9000:
After further testing, it was observed that the alert would fire on multi-az clusters. Updating the query to check for a value of 2 or less before firing.

### DIFF
--- a/deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
@@ -11,7 +11,7 @@ spec:
   - name: sre-infra-node-imbalance
     rules:
     - alert: KubeInfraNodeImbalanceSRE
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-sre-infra-node-imbalance.PrometheusRule.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   - name: sre-infra-node-imbalance
     rules:
     - alert: KubeInfraNodeImbalanceSRE
-      expr: count(count by (node) (kube_pod_info{namespace="openshift-monitoring",pod=~"prometheus-k8s-.*"})) != count(cluster:infra_nodes)
+      expr: count(count by (node) (kube_pod_info{namespace="openshift-monitoring",pod=~"prometheus-k8s-.*"})) < 2
```
cblecker, line 14: Wouldn't this be better as something like this:
```suggestion
      expr: count by (node) (kube_pod_info{namespace="openshift-monitoring",pod=~"prometheus-k8s-.*"}) > 1
```

If the count per node is greater than 1, then there is an imbalance

cblecker (commented):
Line 21 is redundant and can be removed

PR #666, opened 2021-02-03
Adding updated flowschema path for 4.5 and 4.6 clusters

ravitri:
Based on [OSD-6246](https://issues.redhat.com/browse/OSD-6246), the existing workaround that was implemented as per [OSD-5467](https://issues.redhat.com/browse/OSD-5467) was replaced by updated workaround for 4.5 and 4.6 clusters.
* 4.5 BZ with workaround - https://bugzilla.redhat.com/show_bug.cgi?id=1902018#c17  (This is for clusters >4.5.16 where workaround in https://access.redhat.com/solutions/5448851 didn't work) 
* 4.6 BZ the workaround is the same as mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1896240#c19 

This PR removes the old patch and adds new one till the BZ releases fix. Once the fix is released, this patch should be reverted.

cc: @dofinn @cblecker @jewzaam @gsleeman


ravitri:
Based on discussion with API master team, the fix is available in recent versions 4.5.30 and 4.6.13 so no need to apply the workaround patch for now. The overall direction would be to recommend cluster upgrade to 4.6.13 to have the issue fixed.

deploy/OSD-6246-flowcontrol-fix/config.yaml
```diff
@@ -3,7 +3,7 @@ selectorSyncSet:
   matchExpressions:
   - key: hive.openshift.io/version-major-minor
     operator: In
-    values: ["4.5"]
+    values: ["4.5","4.6"]
   - key: hive.openshift.io/version-major-minor-patch
     operator: NotIn
-    values: ["4.5.99"] # this should be replaced with the fixed version
+    values: ["4.6.13"]
```
jewzaam, line 9: Need to enumerate all the versions that have the fix.  For example, with 4.6.14+ this SSS will match.

deploy/bz1883589-kubeapiserver-revert/00-service-accounts.flowschema.patch.yaml
```diff
@@ -1,7 +0,0 @@
-apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
-kind: FlowSchema
-name: service-accounts
-applyMode: AlwaysApply
-# https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c18
-patch: '{"spec":{"priorityLevelConfiguration":{"name":"workload-low"}}}'
```
jewzaam, line 6: Does this need to be applied as a part of this to set the value back to the shipped default?

deploy/OSD-6246-flowcontrol-fix/00-exempt-service-accounts-flowschema.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: FlowSchema
+metadata:
+  name: exempt-service-accounts
```
jewzaam, line 4: Recommend naming to be osd / sre specific and have less possible collision with what may ship with OCP.  Meaning, we need to ensure this does not result in overwriting and then deleting a FlowSchema the core platform provides.

```suggestion
  name: osd-exempt-service-accounts
```

jewzaam (changes requested):

deploy/OSD-6246-flowcontrol-fix/config.yaml
ravitri: Added 4.6.15 and 4.6.16 based on https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/ for now.

deploy/bz1883589-kubeapiserver-revert/00-service-accounts.flowschema.patch.yaml
ravitri: My intent here was to remove and previous P&F patches applied on the clusters of which the one from BZ https://bugzilla.redhat.com/show_bug.cgi?id=1883589 was amongst them. This BZ was used to track the issue initially and that patch was proposed. However, that didn't seem to work and it further got fixed in 4.5.30 and 4.6.13 as per https://bugzilla.redhat.com/show_bug.cgi?id=1908383#c19. This [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1908383) mentions of the same issue even after upgrading to 4.5.16 so better to remove that patch.

I did not want to mix two patches and wanted to consider the common patch for 4.5 and 4.6 which was mentioned more recently as a workaround till the customer upgrades the cluster. Let me know if this answers your query.

jewzaam: By removing the patch from hive it simply stops applying the patch.  If by "remove" you mean "revert" then you have to patch in the old value as was prepared in `bz1883589-kubeapiserver-revert`.  As it stands any cluster that has the flowschema patch from `bz1883589-kubeapiserver-fix` will continue to have that patch.  That leaves pre-existing clusters inconsistent with new clusters.

deploy/OSD-6246-flowcontrol-fix/config.yaml
```diff
@@ -6,4 +6,4 @@ selectorSyncSet:
     values: ["4.5","4.6"]
   - key: hive.openshift.io/version-major-minor-patch
     operator: NotIn
-    values: ["4.6.13"]
+    values: ["4.6.13","4.6.15","4.6.16"]
```
jewzaam, line 9: So we don't have to keep updating this, let's just knock out a reasonably large swath of possible patch numbers (13 to 99):

```suggestion
    values: ["4.6.13","4.6.14","4.6.15","4.6.16","4.6.17","4.6.18","4.6.19","4.6.20","4.6.21","4.6.22","4.6.23","4.6.24","4.6.25","4.6.26","4.6.27","4.6.28","4.6.29","4.6.30","4.6.31","4.6.32","4.6.33","4.6.34","4.6.35","4.6.36","4.6.37","4.6.38","4.6.39","4.6.40","4.6.41","4.6.42","4.6.43","4.6.44","4.6.45","4.6.46","4.6.47","4.6.48","4.6.49","4.6.50","4.6.51","4.6.52","4.6.53","4.6.54","4.6.55","4.6.56","4.6.57","4.6.58","4.6.59","4.6.60","4.6.61","4.6.62","4.6.63","4.6.64","4.6.65","4.6.66","4.6.67","4.6.68","4.6.69","4.6.70","4.6.71","4.6.72","4.6.73","4.6.74","4.6.75","4.6.76","4.6.77","4.6.78","4.6.79","4.6.80","4.6.81","4.6.82","4.6.83","4.6.84","4.6.85","4.6.86","4.6.87","4.6.88","4.6.89","4.6.90","4.6.91","4.6.92","4.6.93","4.6.94","4.6.95","4.6.96","4.6.97","4.6.98","4.6.99"]
```

jewzaam (changes requested):
Reverting the old patch is needed to ensure consistency between pre-existing and new cluster.

cblecker:
/hold
based on requested changes from Naveen

ravitri:
/unhold
@cblecker , @jewzaam - I have done the following:
* Remove fix in from dir `bz1883589-kubeapiserver-fix`
* Apply revert patch from `bz1883589-kubeapiserver-revert` for all 4.5 clusters (since originally the fix was applied to only 4.5 as well)
* Add newer workaround patch of service account flowschema which is valid for 4.5 and 4.6 versions below 4.6.13

Let me know if this seems fine.

jewzaam:
Thanks @ravitri , this looks good. 
/approve
/lgtm

Remove hold when you're ready for it to merge.
/hold

jewzaam (approved):

ravitri:
@jewzaam - Thanks a lot for the review 👍🏼 Just one query though. When all OSD clusters upgrade to 4.6.13+ version, am I right to understand that I can simply remove the `bz1883589-kubeapiserver-revert` and the new workaround patch added in this PR? I think your point to maintain existing revert patch would apply only for existing resources which are modified but not to any resource which we don't want at all. Is that correct?

jewzaam:
> @jewzaam - Thanks a lot for the review 👍🏼 Just one query though. When all OSD clusters upgrade to 4.6.13+ version, am I right to understand that I can simply remove the `bz1883589-kubeapiserver-revert` and the new workaround patch added in this PR? I think your point to maintain existing revert patch would apply only for existing resources which are modified but not to any resource which we don't want at all. Is that correct?

yea, that should be all that's needed.  hive will remove the FlowSchema.  the revert sss would be removed too, which simply stops applying the patch

ravitri:
@jewzaam - Thanks a lot to confirm
Am ok to merge the PR
/unhold

PR #667, opened 2021-02-03
Adds patch for global fluentd toleration

iamkirkbater:
Attempts to add a patch for fluentd tolerations to allow fluentd to be on infra nodes with the new NoSchedule taint.

Helps remove blockers for [OSD-2927](https://issues.redhat.com/browse/OSD-2927)

### DIFF
--- a/deploy/osd-logging/infra-node-taint/00-patch.yaml
+++ b/deploy/osd-logging/infra-node-taint/00-patch.yaml
@@ -1,0 +1,8 @@
+apiVersion: logging.openshift.io/v1
+applyMode: AlwaysApply
+kind: ClusterLogging
+name: instance
+namespace: openshift-logging
+patch: '{"spec":{"collection":{"logs":{"fluentd":{"tolerations":[{"operator": "Exists"}]}}}}}'
+patchType: merge
+
--- a/deploy/osd-logging/infra-node-taint/config.yaml
+++ b/deploy/osd-logging/infra-node-taint/config.yaml
@@ -1,0 +1,7 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: api.openshift.com/environment
+    operator: NotIn
+    values: ["production"]
+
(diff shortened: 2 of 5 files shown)


deploy/osd-logging/infra-node-taint/00-patch.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: logging.openshift.io/v1
+applyMode: AlwaysApply
+kind: ClusterLogging
+name: instance
+namespace: openshift-logging
+patch: '{"spec":{"collection":{"logs":{"fluentd":{"tolerations":[{"operator": "Exists"}]}}}}}'
```
cblecker, line 6: I think this will fail if this doesn't exist.

iamkirkbater: I guess that's the rub here, do we just fail for clusters that don't have logging installed (is there an implication to this other than just failing messages in SSS logs?) or should we do a one-time patch as part of the update process script and then add this to the documentation and _hope_ the customer adds it?

Otherwise if the customer doesn't add this as part of their install if they're not following the official docs we wouldn't get the fluentd pods on the infra nodes and would have be flying blind to those nodes.

jewzaam: We can't do this, customers have put ClusterLogging in CICD, i.e. app-sre... this will get overwritten.  We may need to simply make this work dependent on moving to log forwarding.  Also consider adding this to OSD docs, but that is going away soon with the log forwarding work so isn't worth the short term update.

cblecker: My gut says let's add it to docs, and then as a part of the migration for this, delete all the fluentd pods so that we don't get misscheduled alerts.

The consequence is that fluentd isn't running on infra nodes. If cluster wants those logs, we point them at the docs.

cblecker (changes requested):
/hold

iamkirkbater:
Ok, that's fair to me.  

Actions: I'll update the migration to remove the fluentd pods from the infra nodes after we update the node taints.  I'll close this PR in lieu of that.

PR #669, opened 2021-02-04
enable mgmt of jobs OSD-6023

dofinn:


wanghaoran1988:
You need update https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/srep/10-srep-admins-project.ClusterRole.yml instead of this to allow us manage jobs in the core namespaces instead of all NS including customer's.

cblecker:
/hold

JIRA card doesn't provide clear details of request. Why is SRE hand rolling jobs?

dofinn:
> /hold
> 
> JIRA card doesn't provide clear details of request. Why is SRE hand rolling jobs?

@cblecker https://github.com/openshift/ops-sop/blob/master/v4/alerts/KubeJobFailed.md some of our SOPs require it.

jewzaam:
In that SOP I see we only need create job for the osd-ready-job.  There doesn't appear to be a way to restart jobs.  Given this though we only need to add `"create"`, not `"*"`.  SRE should not be directly editing or patching jobs.

PR #670, opened 2021-02-04
none should upgrade OSD like this OSD-6284

dofinn:


cblecker:
/lgtm
/hold

One question.. is there anything else an SRE might need to do to the clusterversion CR besides upgrading?

If no, LGTM and you are free to `/hold cancel`.

dofinn:
> /lgtm
> /hold
> 
> One question.. is there anything else an SRE might need to do to the clusterversion CR besides upgrading?
> 
> If no, LGTM and you are free to `/hold cancel`.
@cblecker IMO, no. If we need to, we should be elevating privileges anyway. This should not be apart of everyday access.

PR #672, opened 2021-02-04
OSD-5906 Adding samesite cookie strict to prod

treymer:


treymer:
Don't want to merge this going into the weekend. Also, going to look into what it would take to get e2e testing going on route changes like this.

treymer:
e2e test commited to osde2e to verify samesite cookie settings on routes here https://github.com/openshift/osde2e/blob/main/pkg/e2e/verify/samesite_cookie.go

treymer:
This can be merged Monday to avoid any weekend fallout.

PR #673, opened 2021-02-04
Increase Node Infra Imbalance Alert Sample Duration to 4 hours

dustman9000:
This should hopefully suppress alerts showing up during scale test and/or installation time.


fahlmant (approved):
/lgtm

PR #674, opened 2021-02-05
OSD-3075 adding node afinity rules for 4.5+ cluster image registry

luis-falcon:
[OSD-3075](https://issues.redhat.com/browse/OSD-3075) adding node afinity rules for 4.5+ cluster image registry configuration.

Created 2 directories. one for clusters before 4.5 (name: registry) and one for clusters to be patch 4.5+ (name : registry-affinity).

inside there is the config files and the rules. The rule file on the registry-affinity dir has the new nodeAffinity and the config to apply to clusters 4.5+.


deploy/osd-registry/registry-affinity/cluster.Config.patch.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: imageregistry.operator.openshift.io/v1
+applyMode: AlwaysApply
+kind: Config
+name: cluster
+# patch registry to deploy on infras: https://docs.openshift.com/container-platform/4.2/machine_management/creating-infrastructure-machinesets.html#infrastructure-moving-registry_creating-infrastructure-machinesets
+# patch registry to expose default route: https://docs.openshift.com/container-platform/4.1/registry/configuring-registry-operator.html#registry-operator-default-crd_configuring-registry-operator
+patch: '{"spec":{"defaultRoute":true,"nodeSelector":{"node-role.kubernetes.io/infra":""},"affinity":{"nodeAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"preference":{"matchExpressions":[{"key": "node-role.kubernetes.io/infra","operator":"Exists"}]},"weight":1 }]}}, "tolerations":[{"effect":"NoSchedule","key":"node-role.kubernetes.io/infra","operator":"Exists"}] }}'
```
jewzaam, line 7: I've had a tough time with spaces in patch in the cli, do we know this works?  Suggest we not try to test it and just:

```suggestion
patch: '{"spec":{"defaultRoute":true,"nodeSelector":{"node-role.kubernetes.io/infra":""},"affinity":{"nodeAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"preference":{"matchExpressions":[{"key":"node-role.kubernetes.io/infra","operator":"Exists"}]},"weight":1}]}},"tolerations":[{"effect":"NoSchedule","key":"node-role.kubernetes.io/infra","operator":"Exists"}]}}'
```

jewzaam (changes requested):
/approve

jewzaam (changes requested):
Just noticed the original patch isn't moved:
https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-registry/cluster.Config.patch.yaml

Alternative to adding the `registry/` sub-dir is to put `osd-registry/registry/config.yaml` as `osd-registry/config.yaml` and be done.  

As we're going to need to tweak this anyway, here's what I recommend instead...

1. Create `osd-registry/legacy/` directory.
1. Move `osd-registry/cluster.Config.patch.yaml` to `osd-registry/legacy/`.
1. Move `osd-registry/registry/config.yaml` to `osd-registry/legacy/config.yaml`.
1. Remove `osd-registry/registry/`.
1. Move `osd-registry/registry-affinity/*` to `osd-registry/`.

This will have the `osd-registry` SSS be the new normal (4.5+).
And a new SSS created for older clusters, `osd-registry-legacy`.
And in the future we can simply delete `osd-registry/legacy/` when we have no more 4.4 clusters.

jewzaam (approved):

jewzaam:
Hold is for a bit more testing, talking w/ Luis on slack.  Question on having both nodeSelector and affinity.  It works but need to verify things when infra are down, confirm registry is schedule to workers.

dustman9000:
@luis-falcon - It looks like the `/hold` came too late before this was merged. Do you mind testing on stage before we promote to prod?

PR #675, opened 2021-02-05
Fix PR check

jewzaam:
Variable names and name of environments were wrong.
Looks like I broke it on 2020-08-19.  Oops.
The validation to ensure `make` was run has not been running.


fahlmant (approved):
/lgtm

PR #676, opened 2021-02-05
Osd 3075 image registry afinity settings

luis-falcon:
Added a patch to remove the nodeSelector from the image-registry. this will allow for the nodeAffinity rule to function correctly.


cblecker:
This will likely need to be reverted as-is. The two patches will conflict. Doesn't match the design Luis and I discussed.

@jewzaam @luis-falcon

jewzaam:
@cblecker the `nodeSelector` in the other patch needs removed, was missed in this PR. thanks!

PR #678, opened 2021-02-05
Remove KubeInfraNodeImbalanceSRE Alert

dustman9000:
Removing alert due the sheer amount of alerts generated during ROSA scale test. Will come back with a more automated approach.


jharrington22 (approved):
/lgtm

PR #681, opened 2021-02-08
OSD-6387: osd-cluster-ready: Bump backoffLimit

2uasimojo:
During the 1000-cluster scale test, we still saw some instances of the osd-cluster-ready job exceeding the default six retries. With this commit, we're bumping it up to a number that should comfortably exceed the configured `MAX_CLUSTER_AGE_MINUTES`, thus giving that value full control over timing out the job.

[OSD-6387](https://issues.redhat.com/browse/OSD-6387)


jharrington22 (approved):
/lgtm

PR #682, opened 2021-02-08
OSD-6348: Add CuratorJobFailedSRE alert

2uasimojo:
Peel out a special case of KubeJobFailed for logging curator jobs, which alert when a Job fails.

The curator job runs on a 1h interval. If one fails, but the next succeeds, that constitutes a "recovery" which should be considered a happy path in a k8s context. So this alert only fires if we see two consecutive failures (over two hours, to account for the cronjob
interval).

CF:
- Disable KubeJobFailed for curator: https://github.com/openshift/configure-alertmanager-operator/pull/140
- SOP: https://github.com/openshift/ops-sop/pull/1202

[OSD-6348](https://issues.redhat.com/browse/OSD-6348)

### DIFF
--- a/deploy/sre-prometheus/100-curator-jobs.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-curator-jobs.PrometheusRule.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-curator-jobs
+    role: alert-rules
+  name: sre-curator-jobs
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-curator-jobs
+    rules:
+    # https://issues.redhat.com/browse/OSD-6348
+    - alert: CuratorJobFailedSRE
+      expr: kube_job_failed{job="kube-state-metrics",namespace="openshift-logging",job_name=~"^curator-.*"} > 1
+      # The cronjob runs this once an hour; and (anecdotally) the job
+      # takes close to an hour to run.
+      for: 120m
+      labels:
+        severity: warning
+        namespace: '{{ $labels.namespace }}'
+      annotations:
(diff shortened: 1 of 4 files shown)


2uasimojo:
/hold

I want to get the configure-alertmanager-operator and ops-sop PRs in place before we land this.

deploy/sre-prometheus/100-curator-jobs.PrometheusRule.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  groups:
+  - name: sre-curator-jobs
+    rules:
+    # https://issues.redhat.com/browse/OSD-6348
+    - alert: CuratorJobFailedSRE
+      expr: kube_job_failed{job="kube-state-metrics",namespace="openshift-logging",job_name=~"^curator-.*"} > 1
+      # The cronjob runs this once an hour; and (anecdotally) the job
+      # takes close to an hour to run.
+      for: 120m
```
dofinn, line 18: Should we bump this to 125m to allow the next job to run? When successful its duration is short. 

```oc -n openshift-logging get jobs                                                                       ✔    default/mta-dev-test/dofinn.openshift ⎈  !2/24  
NAME                                        COMPLETIONS   DURATION   AGE
curator-1612822500                          1/1           6s         46m
```

2uasimojo: Whoah, you ran one in 6s? The ones I've seen take ~50m.

The 2h takes into account that the cronjob runs them every 1h -- i.e. I'm giving the job a full hour to complete as well.

2uasimojo:
> I want to get the configure-alertmanager-operator and ops-sop PRs in place before we land this.

These are up.

But I want to continue to hold for discussion. In particular, see [this part of the SOP writeup](https://github.com/openshift/ops-sop/pull/1202/files#diff-76f5be694d45beb9b55b447e615b37f8f35d0e973409d186953bfff0ebe7b347R31-R34).

PR #692, opened 2021-02-11
Use podman as a priority, fall back to docker

jharrington22:
Use podman as priority, fall back to docker.

Tested to generate files for #691

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ ifndef GEN_TEMPLATE
 $(error GEN_TEMPLATE is not set; check project.mk file)
 endif
 
-CONTAINER_ENGINE?=docker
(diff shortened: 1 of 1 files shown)


dofinn:
are we standardizing on this across other makefiles in OSD?

2uasimojo:
> are we standardizing on this across other makefiles in OSD?

[Yes](https://github.com/openshift/boilerplate/blob/459c3cdf9d1bf451f99a08fbc21833c817d540e1/boilerplate/openshift/golang-osd-operator/standard.mk#L18-L19), but mcc is a snowflake and not likely to use boilerplate any time soon. (It's certainly not suited to the golang-osd-operator convention, which is the only one we have right now.)

cblecker:
Based on this is what's in boilerplate, I'm fine with it.

/lgtm

jharrington22:
Thanks

PR #693, opened 2021-02-11
Update owners files

cblecker:
Swap in new FLs/TLs. Manually add @jewzaam to approvers so that he can still approve changes.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -2,14 +2,15 @@
 
 aliases:
   srep-functional-leads:
-    - cblecker
     - dofinn
+    - dustman9000
+    - fahlmant
     - jaybeeunix
-    - lisa
+    - jharrington22
     - NautiluX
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -2,14 +2,15 @@
 
 aliases:
   srep-functional-leads:
-    - cblecker
     - dofinn
+    - dustman9000
+    - fahlmant
     - jaybeeunix
-    - lisa
+    - jharrington22
```
fahlmant, line 9: I don't think James is needed here since he's below on level 15

cblecker: @fahlmant He's listed twice because he's technically the FL of Omega too

fahlmant: Ah got it :+1:

fahlmant (approved):
/lgtm

PR #694, opened 2021-02-12
Update cluster-monitoring-config.yaml to add killswitch note

RiRa12621:
We want the killswitch instruction to be up to date.
if the location of this section changes to a different file, the SOP **has to** be updated.

Depends on https://github.com/openshift/ops-sop/pull/1207/files


RiRa12621:
Thanks @cblecker

PR #695, opened 2021-02-16
Added temporary metrics for observatorium for scale tests

arjunrn:
Added metrics `kube_node_info` and `kube_node_created` and `kube_node_role` which consists of a timeseries for
every node in the cluster.


jharrington22:
@arjunrn can we remove the WIP?

/lgtm

PR #699, opened 2021-02-17
Enable production to send metrics to new observatorium tenant

RiRa12621:
The new observatorium tenant is ready and we want to start sending metrics.

This PR enabled remote_write in production and removes the limitation on the token-refresher deployment


jharrington22 (approved):
/lgtm

PR #706, opened 2021-02-19
Adding storage class pd-ssd to GCP clusters

aliceh:
For test purposes made pd-ssd the default storage class.

### DIFF
--- a/deploy/osd-storage/cluster-storage-gcp-ssd.yaml
+++ b/deploy/osd-storage/cluster-storage-gcp-ssd.yaml
@@ -1,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: ssd
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/gce-ssd
+parameters:
+  type: pd-ssd
+  replication-type: none
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
+reclaimPolicy: Delete
--- a/deploy/osd-storage/cluster-storage-gcp-standard.yaml
+++ b/deploy/osd-storage/cluster-storage-gcp-standard.yaml
@@ -1,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "false"
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-standard
+  replication-type: none
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
+reclaimPolicy: Delete
--- a/deploy/osd-storage/config.yaml
+++ b/deploy/osd-storage/config.yaml
@@ -1,0 +1,9 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet: 
+  matchExpressions: 
+    - key: hive.openshift.io/cluster-platform
+      operator: In
+      values: [gcp]
+    - key: api.openshift.com/environment
+      operator: In
+      values: [stage]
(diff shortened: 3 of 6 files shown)


deploy/osd-storage/cluster-storage-gcp-standard.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "false"
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-standard
```
dustman9000, line 9:
```suggestion
  type: pd-ssd
```
We already have pd-standard, we want to add pd-ssd right?

deploy/osd-storage/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet: 
+  matchExpressions: 
+    - key: hive.openshift.io/cluster-platform
+      operator: In
+      values: [gcp]
```
fahlmant, line 6: I think these need to be wrapped in double quotes

deploy/osd-storage/config.yaml
```diff
@@ -0,0 +1,9 @@
...
+    - key: api.openshift.com/environment
+      operator: In
+      values: [stage]
```
dustman9000, line 9: Why are we restricting the environment to stage?

deploy/osd-storage/cluster-storage-gcp-standard.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard
```
dustman9000, line 4: this should be called `ssd`

dustman9000, line 4:
```suggestion
  name: ssd
```

deploy/osd-storage/cluster-storage-gcp-standard.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
```
dustman9000, line 1: you can remove this file as we already have this storage class for gcp clusters

deploy/osd-storage/cluster-storage-gcp-ssd.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: ssd
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
```
dustman9000, line 6: We want this to be false 
```suggestion
    storageclass.kubernetes.io/is-default-class: "false"
```

deploy/osd-storage/config.yaml
dustman9000, line 6:
```suggestion
      values: ["gcp"]
```

deploy/osd-storage/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: SelectorSyncSet
```
dustman9000, line 1: We also want to add CCS to the constraints:
```
  - key: api.openshift.com/ccs
    operator: In
    values: ["true"]
```

deploy/osd-storage/config.yaml
```diff
@@ -0,0 +1,9 @@
...
+selectorSyncSet: 
+  matchExpressions:
```
dustman9000, line 3: We want to only create and not continuously update: https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/syncset_types.go#L37
```suggestion
  applyBehavior: "CreateOnly"
  matchExpressions: 
```

PR #707, opened 2021-02-19
Added annotation to force recreation of token-refresher pods

arjunrn:
Since the `token-refresher` pods use values from a secret which is pushed through app-interface the pods themselves do not get recreated when the secret values are updated. The only way to force the pods to use the new values is to recreate the pods and adding this annotation will force that.

### DIFF
--- a/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
+++ b/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
@@ -15,6 +15,8 @@ spec:
       app.kubernetes.io/name: token-refresher
   template:
     metadata:
+      annotations:
+        last-updated: "2021-19-02"
       labels:
         app.kubernetes.io/component: authentication-proxy
         app.kubernetes.io/name: token-refresher
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -7455,6 +7455,8 @@ objects:
             app.kubernetes.io/name: token-refresher
         template:
           metadata:
+            annotations:
+              last-updated: '2021-19-02'
             labels:
               app.kubernetes.io/component: authentication-proxy
               app.kubernetes.io/name: token-refresher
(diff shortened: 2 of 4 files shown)


deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
```diff
@@ -15,6 +15,8 @@ spec:
       app.kubernetes.io/name: token-refresher
   template:
     metadata:
+      annotations:
+        last-updated: "2021-19-02"
```
cblecker, line 19: Should this be a variable itself that is updated by app-interface when the secret is?

jharrington22: That sounds like a APPSRE feature to me :slightly_smiling_face:

jharrington22: cc/ @maorfr thoughts? Discussion could go else where.

maorfr: This Deployment is a part of a SelectorSyncSet, which means it lands on a customer cluster. What secret in app-interface is related to that?

maorfr: From the app-interface side, if you add `qontract.recycle: "true"` annotation to the Secret itself, all using pods will be recycled.

maorfr: You can totally do that. Use the openshift templates parameter and add that parameter to the mcc saas file.

jharrington22: This would be awesome, thanks @cblecker @maorfr I'm going to card this up as there will be other secrets that want the same treatment. For this PR however we'll merge as is and I'll not it in the card to be reverted once the above solution is in place.

2uasimojo: I'm apparently the only one who cares that this date is scrambled.

PR #711, opened 2021-02-22
Enable remote write in old configuration

arjunrn:
This PR should be merged only when the UWM changes have to be reverted.

/assign @jharrington22


fahlmant:
This needs to merge before the UWM changes get merged

PR #712, opened 2021-02-22
Adding pd-ssd storage class

aliceh:
In this commit storage calss pd-ssd will be added to all gcp clusters. [OSD-6468](https://issues.redhat.com/browse/OSD-6468)


deploy/osd-storage/config.yaml
```diff
@@ -0,0 +1,11 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet: 
+  applyBehavior: "CreateOnly"
```
dustman9000, line 3: This is so a dedicated-admin user can setup their own private encryption keys in the future.

aliceh: Thanks for explaining!

jharrington22 (approved):
/lgtm

PR #715, opened 2021-02-23
Enable remote writes for non-UWM clusters with correct queue config

arjunrn:
The `maxSamplesPerSend` configuration item was not set correctly. So it's value was not applied. Also for cluster versions <= 4.5 the default value was 100 and this was not high enough. Newer versions of prometheus have a default value of 500 and this is now the value we will use.


jharrington22:
Tested on a 4.5 cluster, after the `cluster-monitoring-config` configmap was applied prometheus reported the following:

```
  queue_config:
    capacity: 500
    max_shards: 1000
    min_shards: 1
    max_samples_per_send: 500
    batch_send_deadline: 5s
    min_backoff: 30ms
    max_backoff: 100ms
```

PR #716, opened 2021-02-23
Reinstall RMO if too many CSVs are on cluster

NautiluX:
* This is to mitigate a bug in OLM:
  * old CSVs are stuck in `Replace` state
* This CronJob reinstalls RMO to clean up stuck CSVs
* The CronJob will clean itself up after running

### DIFF
--- a/deploy/sre-pruning/115-reinstall-rmo.CronJob.yaml
+++ b/deploy/sre-pruning/115-reinstall-rmo.CronJob.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sre-rmo-cleanup-sa
+  namespace: openshift-route-monitor-operator
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: sre-rmo-cleanup-role
+  namespace: openshift-route-monitor-operator
+rules:
+- apiGroups:
+  - "operators.coreos.com"
+  resources:
+  - clusterserviceversions
+  - subscriptions
+  - installplans
+  verbs:
+  - list
+  - get
+  - delete
+  - create
+- apiGroups:
+  - "batch"
+  resources:
+  - cronjobs
+  verbs:
+  - list
+  - get
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - list
+  - get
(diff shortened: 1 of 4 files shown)


deploy/sre-pruning/115-reinstall-rmo.CronJob.yaml
```diff
@@ -0,0 +1,107 @@
...
+            imagePullPolicy: Always
+            command:
+            - sh
+            - -c
+            - |
+              #!/bin/bash
+
+              set -euxo pipefail
+
+              COUNT_CSV=$(oc get csv | grep route-monitor-operator | wc -l) || true
```
georgettica, line 94: on a different solution, you can all the namespace here
```suggestion
              COUNT_CSV=$(oc -n openshift-route-monitor-operator get csv  | grep route-monitor-operator | wc -l) || true
```

deploy/sre-pruning/115-reinstall-rmo.CronJob.yaml
```diff
@@ -0,0 +1,107 @@
...
+          restartPolicy: Never
+          containers:
+          - name: rmo-reinstaller
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - sh
+            - -c
+            - |
+              #!/bin/bash
```
georgettica, line 90: does that actually run as a bash script?

georgettica (commented):
overall I added some nits that are good in the future and may make this readable, but I won't block it because of that

deploy/sre-pruning/115-reinstall-rmo.CronJob.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+              set -euxo pipefail
+
+              COUNT_CSV=$(oc get csv | grep route-monitor-operator | wc -l) || true
+              if [[ $COUNT_CSV -gt 2 ]]; then
+                oc -n openshift-route-monitor-operator get csv | grep -v NAME | awk '{print $1}'  | grep route-monitor-operator | xargs oc delete csv
```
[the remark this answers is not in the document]
NautiluX: thanks for pointing out other options, since this is not meant to stay on clusters very long, I think it's not worth the effort to change and test again ;)

mwoodson (approved):
/lgtm

PR #717, opened 2021-02-23
Revert "Reinstall RMO if too many CSVs are on cluster"

jharrington22:
Reverts openshift/managed-cluster-config#716


fahlmant (approved):
/lgtm

PR #721, opened 2021-02-24
remove clusterlogging operatorgroup

RiRa12621:
We are allowing dedicated admins to install the operator as a whole and OLM has enough privileges to lay down the operatorgroup.
If we leave it like that, we run into conflicts with the operatorgroup coming via OLM as it has one more API


RiRa12621:
/hold
need to test

cblecker:
/lgtm
/approve
/hold

Is there a JIRA/bug for this Rick?

RiRa12621:
It's part of the whole "unblock" logging upgrade

So no explicit ticket for just this.
Probably makes sense to have one

PR #723, opened 2021-02-25
OSD-6583 enable observatorium in nonprod

dofinn:
Due to ongoing scaling issues in production/observatorium. This PR
enables observatorium in staging only.

Caveats:
This is double handling. This now has a prod and non-prod configuration
that may need to stay in sync until the scaling issues are resolved.


arjunrn:
@dofinn I'm ok with this change. But I would recommended changing the `queueConfig` to be in-line with the values [here](https://gist.github.com/spaparaju/f2be03879f4a96831812beca1d898d19).

dofinn:
@arjunrn Thanks. Please see 7bc926b

jharrington22:
@dofinn +1 to this however one small nit. 

Can we change the directory names on the existing to have the suffix `-prod`

This change has `non-uwm-non-prod` I feel that might get missed, I'd rather be specific about whats in production.

deploy/cluster-monitoring-config-non-prod/cluster-monitoring-config.yaml
```diff
@@ -0,0 +1,101 @@
...
+            action: keep
+            regex: '(cluster_admin_enabled|identity_provider|kube_node_info|kube_node_created|kube_node_role)'
+          queueConfig:
+            capacity: 2500
+            maxShards: 1000
+            minShards: 1
+            maxSamplesPerSend: 500
+            batchSendDeadline: 60s
+            minBackoff: 30ms
+            maxBackoff: 100ms
```
jharrington22, line 27: @dofinn @arjunrn if we are going to set these parameters for stage in all versions we should do the same in production

dofinn: sure, lets tune stage first tho? try hit staging observatorium with some metrics to see if its client or server at under saturation that causes issues?

dofinn:
> @dofinn +1 to this however one small nit.
> 
> Can we change the directory names on the existing to have the suffix `-production`
> 
> This change has `non-uwm-non-prod` I feel that might get missed, I'd rather be specific about whats in production.

@jharrington22 done: fd41ad64dd5fe04e6f54a246899a5fa370cb4735

deploy/cluster-monitoring-config-non-uwm-staging-integration/cluster-monitoring-config.yaml
```diff
@@ -0,0 +1,100 @@
...
+  name: cluster-monitoring-config
+  namespace: openshift-monitoring
+data:
+  config.yaml: |
+    prometheusK8s:
+    # If the location of this config changes,
+    # make sure to adjust https://github.com/openshift/ops-sop/blob/master/v4/howto/o11m_mst_killswitch.md
+    # so that killswitch instructions point to the right file!
+      remoteWrite:
+        - url: http://token-refresher.openshift-monitoring.svc.cluster.local
```
arjunrn, line 13: I just remembered that we don't want to rollout remote writes to 4.5 clusters which this file targets.

arjunrn: I believe the `non-uwm` configuration should be the same for both production and non-production clusters. So they can probably be merged.

PR #726, opened 2021-03-01
Remove last-updated annotation from token-refresher

arjunrn:
Removed the `last-updated` annotation which was added in #707 to force the pods to be recreated and update the secret value. This is not required anymore and can be removed. It's also causing [failures in osde2e](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/osde2e-stage-aws-addon-ocs-converged/1366176564076613632) CI tests.


jharrington22 (approved):
/lgtm

PR #727, opened 2021-03-01
Scc that will allow dedicated-admins the rights to capture pcaps

luis-falcon:


deploy/osd-pcap-collector/pcap-collector-scc.yaml
```diff
@@ -0,0 +1,20 @@
+kind: SecurityContextConstraints
+apiVersion: v1
+metadata:
+  name: pcap
+allowPrivilegedContainer: true 
+allowHostNetwork: true
+allowedCapabilities:
+- 'NET_ADMIN'
+-  'NET_RAW'
```
dustman9000, line 9: nit
```suggestion
- 'NET_RAW'
```

deploy/osd-pcap-collector/config.yaml
```diff
@@ -0,0 +1,3 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  applyBehavior: "CreateOnly"
```
dustman9000, line 3: We want it to continuously update, so remove this line and use the default apply behavior. 
```suggestion
```

deploy/osd-pcap-collector/config.yaml
```diff
@@ -0,0 +1,3 @@
+deploymentMode: "SelectorSyncSet"
```
dustman9000, line 1: We can remove this file altogether to such that it gets applied everywhere.

deploy/osd-pcap-collector/pcap-collector-scc.yaml
```diff
@@ -0,0 +1,20 @@
...
+allowHostNetwork: true
+allowedCapabilities:
+- 'NET_ADMIN'
+- 'NET_RAW'
+runAsUser:
+  type: MustRunAsNonRoot
+seLinuxContext:
+  type: RunAsAny
+users:
+- system:serviceaccount:sbr-hosted-pcap:default
```
dustman9000, line 15: Note: This is a hard-coded namespace for CEE to run node-level pcaps from...

deploy/osd-pcap-collector/pcap-collector-scc.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: pcap
+allowPrivilegedContainer: true 
+allowHostNetwork: true
+allowedCapabilities:
+- 'NET_ADMIN'
+- 'NET_RAW'
+runAsUser:
+  type: MustRunAsNonRoot
+seLinuxContext:
+  type: RunAsAny
```
dustman9000, line 13: We need to make sure this is also non-root.
```suggestion
  type: MustRunAsNonRoot
```

luis-falcon: Ok. will test this in a cluster.

deploy/osd-pcap-collector/pcap-collector-scc.yaml
```diff
@@ -0,0 +1,20 @@
...
+users:
+- system:serviceaccount:sbr-hosted-pcap:default
+groups:
```
cblecker, line 16: Why are we assigning this to groups in this way? This will make this the default SCC for these groups.

deploy/osd-pcap-collector/pcap-collector-scc.yaml
```diff
@@ -0,0 +1,20 @@
...
+allowPrivilegedContainer: true 
+allowHostNetwork: true
+allowedCapabilities:
+- 'NET_ADMIN'
+- 'NET_RAW'
+runAsUser:
+  type: MustRunAsNonRoot
+seLinuxContext:
+  type: RunAsAny
+users:
```
cblecker, line 14: We should grant SCC access via RBAC, rather than in this way

deploy/osd-pcap-collector/pcap-collector-scc.yaml
jaybeeunix: There isn't a MustRunAsNonRoot in the SELinux context. The only other option is to specify MustRunAs and manually specify the security context. Luis is trying out a couple of suggestions I have there.

PR #733, opened 2021-03-03
Enable oc adm policy who-can

wanghaoran1988:


wanghaoran1988:
For reference: https://github.com/openshift/openshift-apiserver/blob/9aa1ff543fee47e6dbf4019f35395be97b93412c/pkg/bootstrappolicy/policy.go#L228
we need both authz api group and legacy group.

PR #736, opened 2021-03-04
chore(route-monitor-operator): add console monitor [OSD-5499]

georgettica:
I have 2 questions for this PR:
- who should be the owners
- what should the `.spec.targetAvailabilityPercent` be?


RiRa12621:
It doesn't need a separate owner.
You can't merge without the gods' approval to prod anyway
Availability seems fine, needs revisit after a while to see if we can raise it

georgettica:
/hold
I need to add a SOP for it

deploy/osd-route-monitor-operator/100-rmo.RouteMonitor.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: monitoring.openshift.io/v1alpha1
+kind: RouteMonitor
+metadata:
+  name: routemonitor-sample
```
RiRa12621, line 4:
```suggestion
name: routemonitor-sample
```

cblecker:
> It doesn't need a separate owner.

There should be an owners file with a `reviewers` section detailing who should be point of contact for reviews. You can look at the other folders under `deploy/` for examples

> what should the `.spec.targetAvailabilityPercent` be?

Do we have a process for selecting these? Will alerts be raised based off this, or just a metric?

cc @dofinn

RiRa12621:
Who needs that if it can't go in without "approved" anyway?

Yes alerts is what route monitor operator does

cblecker:
> Who needs that if it can't go in without "approved" anyway?

Because change to MCC should be reviewed by subject matter experts in addition to leads.

RiRa12621:
So you need me to tell you it's ok, so you can approve? Not very efficient imho
Let SMEs approve if they're (rightfully) supposed to evaluate and review changes.

deploy/osd-route-monitor-operator/100-rmo.RouteMonitor.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: monitoring.openshift.io/v1alpha1
+kind: RouteMonitor
+metadata:
+  name: console
+spec:
+  route:
+    name: console
+    namespace: openshift-console
+  slo:
+    targetAvailabilityPercent: "99.5"
```
dofinn, line 10: This needs to be a string?

dofinn:
We have nothing in stone that helps us define initial values for SLOs. 
Lets make sure we have some dedicated actions to follow this up to analyze if we can increase this SLO based on current data.

deploy/osd-route-monitor-operator/100-rmo.RouteMonitor.yaml
georgettica: There are no floating points support in the apiserver :(

dofinn: The types defined in the CRD? 
https://github.com/openshift/route-monitor-operator/blob/master/config/crd/bases/monitoring.openshift.io_routemonitors.yaml#L55  

We could update this to be a `number` as per: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#type

georgettica: thanks @dofinn ! I'll see how we do that in json stuffz

georgettica: @dofinn seems this is not 100% supported by operator-sdk / kubebuilder as of now
see https://github.com/kubernetes-sigs/controller-tools/issues/245 as people work on shying away from floats

also in an internal discussion I tried to understand how to implement doubles in the api resource, but didn't revieve any ❤️  on it (aside of @dastergon)

georgettica:
/unhold
changes that @NautiluX requested are in and SOP is also in

@RiRa12621 can you give a final /lgtm?
not sure if @dofinn has additional notes on this PR aside of what was raised

deploy/osd-route-monitor-operator/100-rmo.RouteMonitor.yaml
dofinn: ooo thats interesting. nice find ! @georgettica

georgettica:
as @NautiluX suggeste, changed NS and need a lgtm becuase of it

georgettica:
tested on stage (I can replace my routemonitor with this one, but outcome should be similar)

georgettica:
changing to the current CR still works and is scraped, only future concern is the message naming, which should be solved in the RMO code itself

(the value of the breach is 1 and not a good indication to how long the alert is firing)

deploy/osd-route-monitor-operator/100-openshift-route-monitor-operator.console.RouteMonitor.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     name: console
     namespace: openshift-console
   slo:
-    targetAvailabilityPercent: "99.5"
+    targetAvailabilityPercent: "99.95"
```
NautiluX, line 11: Why now 99.95? According to the [SLO doc](https://docs.google.com/document/d/1MIh_4GNTFyMaGnPBrq_cxKIrWoKw7CMShq47qp5Hj48/edit#) it should be 99.5? Probably I miss some discussion here...

NautiluX: turns out this was confused with the API server SLA. We'll change it back to 99.5.

wgordon17:
Will this automatically merge once #771 is resolved?

georgettica:
@wgordon17 If I get a +1 for this then yes

wgordon17:
@georgettica Consider this a ➕ 1 from PM if it helps :) Not sure if you need plus one's from others ¯\_(ツ)_/¯

cblecker:
@georgettica LGTM, but as previously commented, needs an `OWNERS` file that looks something like:
```
reviewers:
- georgettica
- RiRa12621
```

PR #738, opened 2021-03-05
OSD-6387: osd-cluster-ready: Actually bump backoffLimit

2uasimojo:
In PR #681 we mistakenly put the `backoffLimit` in the Pod spec instead of the Job spec, so it had no effect. Fix.

[OSD-6387](https://issues.redhat.com/browse/OSD-6387)


jharrington22:
This is awkward. Nice catch!

/lgtm

jharrington22 (approved):

PR #745, opened 2021-03-11
Add a notification banner for clusters in Limited Support

cblecker:
This adds an in-cluster banner for clusters that are in Limited Support prompting them to check OCM


dofinn:
@cblecker nice work!, however we should plan to lifecycle this with releases of CIS. Any thoughts there? otherwise i feel it will always be an after thought.

PR #751, opened 2021-03-16
Added separate monitoring configuration for scale test clusters

arjunrn:
Also added label selector to restrict configuration for only scale test clusters. When the scale test is over this PR can simply be reverted.

### DIFF
--- a/deploy/cluster-monitoring-config-scale-test/OWNERS
+++ b/deploy/cluster-monitoring-config-scale-test/OWNERS
@@ -1,0 +1,3 @@
+reviewers:
+- boranx
+- cblecker
--- a/deploy/cluster-monitoring-config-scale-test/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config-scale-test/cluster-monitoring-config.yaml
@@ -1,0 +1,101 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-monitoring-config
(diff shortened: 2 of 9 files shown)


jharrington22:
@arjunrn looks good to me. 

@fahlmant @cblecker would both of you mind putting an eye on this before I promote to stage. I'd like to put this into prod today in preparation for the scale tests.

deploy/cluster-monitoring-config-non-uwm/config.yaml
```diff
@@ -8,3 +8,6 @@ selectorSyncSet:
     - key: ext-managed.openshift.io/uwm-disabled
       operator: In
       values: ["true"]
+    - key: api.openshift.com/legal-entity-id
+      operator: NotIn
+      values: ${{SCALE_TEST_LEGALENTITY_IDS}}
```
cblecker, line 13: ORing on this means that his config will go out to all non-scale test clusters in every environment

cblecker (changes requested):
/hold

This won't work as-is.

deploy/cluster-monitoring-config-non-uwm/config.yaml
jharrington22: right, I asked about that. We can remove this OR but what does that mean for the non-uwm config since this is in prod today?

Do we need another split?

cblecker:
/close
Closing in favour of #753

PR #753, opened 2021-03-16
Enable Observatorium for only certain legal entity ids in all envs

jharrington22:
Remove configuration differences between staging/integration and production and only enable Observatorium for certain legal entity id's. 

Legal entity id's are stored in app-interface [MR 16378](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/16378)

This the result of this PR is less metrics being shipped to the staging Observatorium but control for the production scale test to be only scale test clusters. We can continue to ramp up metrics sending with legal entity id's until we're ready to enable it across the fleet.

This PR closes #751 

cc/ @arjunrn @cblecker @dofinn @jeremyeder


cblecker:
/lgtm
/hold

Needs the app-interface MR with the vars to merge first

PR #757, opened 2021-03-17
Fix apiversion of the scc. Removed dedicated admins from rolebinding.…

luis-falcon:
Fix apiversion of the scc. Removed dedicated admins from rolebinding Dedicated admins should not be able to perform node level pcaps


deploy/osd-pcap-collector/03-pcap-collector-ClusterRoleBind.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: pcap-collector 
 subjects:
 - kind: Group
-  name: dedicated-admins
+  name: osd-sre-admins
```
dustman9000, line 7: remove whitespace

PR #760, opened 2021-03-17
OSD-6646: Bump osd-cluster-ready (no longer needs to exec) 

2uasimojo:
https://github.com/openshift/osd-cluster-ready/pull/7 removes the logic that talks to prometheus and alertmanager, so `pods/exec`
permissions are no longer needed.

[OSD-6646](https://issues.redhat.com/browse/OSD-6646)


deploy/rosa-console-branding-configmap/rosa-brand-logo.yaml
```diff
@@ -1,6 +1,48 @@
...
+  rosa-brand-logo.svg: "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<!-- Generator:
```
2uasimojo, line 3: So hey, no idea why this delta is here, but it was generated when I ran `make` so...

jharrington22: @cblecker any idea?

jharrington22: Delta was tested on two clusters.

jharrington22:
Delta was tested on two clusters.

/lgtm
/approve

deploy/rosa-console-branding-configmap/rosa-brand-logo.yaml
2uasimojo: I expect the newlines are irrelevant, since they're embedded in XML.

I thought CI would have complained about this if it was a problem... but that job is [only checking the generated deployment yaml](https://github.com/openshift/managed-cluster-config/blob/4d4bb85e9dd4113a2c301633eb7fdb0ad56663ac/hack/app_sre_pr_check.sh#L27-L40).

I'll spin up a new stage cluster once this lands to make sure it's copacetic.

PR #764, opened 2021-03-19
OSD SME alerts

dofinn:
@jeremyeder 
@cblecker 
As per meet.


RiRa12621:
That's probably not going to work since owners files don't work with GitHub teams.
You need to either put everyone's handles or create and alias 
.
There's and example for an alias file in the root of this repository

PR #768, opened 2021-03-24
Enable Observatorium for OSD

jharrington22:
This PR enables Observatorium for all OSD clusters ROSA/OSD/CCS

To disable this revert this PR.


cblecker:
We need to remove https://github.com/openshift/managed-cluster-config/blob/659fc7d2c6158297b6fb0af91ccce8bbdbbd11c5/deploy/cluster-monitoring-config/config.yaml#L10-L12 and should add the observatorium pieces to https://github.com/openshift/managed-cluster-config/blob/master/deploy/cluster-monitoring-config-non-uwm/cluster-monitoring-config.yaml

PR #769, opened 2021-03-24
permissions for dedicated admins to launch pcaps

luis-falcon:
Added permissions for dedicated admins to launch pcaps at the POD interface in their own namespaces


dustman9000:
Looks good mechanically from a functionality standpoint. I want to hold just so security can give a final OK.
/lgtm
/hold

dustman9000:
As a side conversation with security folks, this looks good...
/hold cancel

PR #772, opened 2021-04-01
schedule openshift-state-metrics on infra nodes

yithian:
for [OSD-6637](https://issues.redhat.com/browse/OSD-6637)


yithian:
tested by applying the change manually to a cluster in stage

jharrington22 (approved):

PR #773, opened 2021-04-07
MTSRE team perms

sedroche:


deploy/backplane/mtsre/config.yaml
```diff
@@ -0,0 +1,4 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchLabels:
+      api.openshift.com/addon-managed-odh: "true"
```
sedroche, line 4: I think this is correct, the addon id is `managed-odh` and the others are prefixed `addon-`

wanghaoran1988:
@sedroche My understanding is "managed-odh" is one addon that managed by mtsre team, so I'd like the perms related to "managed-odh" can be in a separate directory similar with cssre team, see https://github.com/openshift/managed-cluster-config/tree/master/deploy/backplane/cssre/rhoam

wanghaoran1988:
I am not sure whether we should enable them to escalate, but it looks like we should in long term, @cblecker WDYT? if the answer is yes, we should update https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/elevated-sre/10-impersonate-cluster-admin.ClusterRoleBinding.yml

deploy/backplane/mtsre/00-mtsre.namespace.yml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-backplane-mtsre
```
chopraapooja, line 4: Why these files are prefixed with numbering ? is it apply order ?

sedroche:
@cblecker Another question of whether this team will need monitoring perms like cssre. e.g. https://github.com/openshift/managed-cluster-config/blob/master/deploy/backplane/cssre/01-cssre-monitoring-role.yml

cblecker:
This seems fine to me

/approve

PR #774, opened 2021-04-07
chore: containerize the generation of files

boranx:
Fixes: https://issues.redhat.com/browse/OSD-6775

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 tmp_exporters/
 generated/
 .DS_Store
+.vscode/
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,14 @@ endif
 
 CONTAINER_ENGINE=$(shell command -v podman 2>/dev/null || command -v docker 2>/dev/null)
 
+ifeq ($(CONTAINER_ENGINE),)
+# Running already in a container
+OC := oc --dry-run=true --kubeconfig=.kubeconfig
+else
+# Run the oc inside container
+OC := $(CONTAINER_ENGINE) run --rm -v `pwd -P`:`pwd -P`:z -w=`pwd` quay.io/openshift/origin-cli:4.7.0 oc --dry-run=client --kubeconfig=.kubeconfig
+endif
+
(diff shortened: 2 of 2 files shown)


2uasimojo:
This change looks sane to me. Not sure why `$(CONTAINER_ENGINE)` isn't resolving, and I can't seem to log into the CI environment to fiddle around. Maybe the env has already expired? (Though that seems awful quick.)

/retest

boranx:
Maybe the CI doesn't have a `CONTAINER_ENGINE` though it runs on my local

2uasimojo:
> Maybe the CI doesn't have a `CONTAINER_ENGINE` though it runs on my local

Ohhh, I understand now -- the CI env is already running in a container. You can't run docker from within a container.

This means we'll have to resort to some kind of trickery so the `make` targets know whether they're running in a container or not.

2uasimojo:
> This means we'll have to resort to some kind of trickery so the `make` targets know whether they're running in a container or not.

It would probably be adequate to simply detect whether `$(CONTAINER_ENGINE)` is set. Something like:

```make
ifeq ($(CONTAINER_ENGINE),)
	# We're already in a container
	oc ...
else
	# Running locally
	$(CONTAINER_ENGINE) run ...
endif
```

boranx:
@2uasimojo agreed, the fix has just been applied which I think seems more cleaner now

boranx:
@2uasimojo agreed, the fix has just been applied which I think seems more cleaner now

2uasimojo:
Tested this locally and via a [dummy commit](https://github.com/openshift/managed-cluster-config/pull/776). I wasn't expecting the hook to automatically run `make` and commit the changes -- when did that get added? -- but as far as this PR goes, everything works beautifully. Thanks @boranx !

/lgtm

2uasimojo:
> when did that get added?

According to git history, it has been there forever. Why did I always have to do this manually?

boranx:
yeah, that's a nice bot trick :))

PR #779, opened 2021-04-12
Fix deployment config for RMO

cblecker:
The deploymentMode here would have caused this to never be deployed. This changes it to be deployed to all non-production environments. This file can be removed completely to push this out to prod when ready.

FYI @NautiluX @georgettica @RiRa12621


RiRa12621:
That explain why there were no alerts 😁

PR #780, opened 2021-04-12
Remove stale bug workarounds

cblecker:
A number of workarounds and pruners were put in place to deal with bugs. This removes a number of these that no longer apply:

- BZ1915081 workaround only applies to clusters that were originally installed in versions prior to 4.5. It only needs to be applied once. As we don't support 4.4 installs and all existing clusters have this, this can be removed.
- BZ1915661 workaround only applies to 4.7.0 FCs. We don't have/support these versions anymore, so this can be removed.
- BZ1793850 workaround only applies to clusters that were originally installed in versions prior to 4.4. It only needs to be applied once. As we don't support 4.3 installs and all existing clusters have this, this can be removed.
- BZ1843243 workaround only applies to clusters that were originally installed in versions prior to 4.4. It only needs to be applied once. As we don't support 4.3 installs and all existing clusters have this, this can be removed.
- The cronjob pruner was a remanent left around after a bad cronjob was pushed out. It can be safely removed.


jharrington22:
Nice @cblecker ! 

/lgtm

PR #781, opened 2021-04-14
Move console branding CRs over to patches

cblecker:
This moves the console CR over to use a patch, rather than a completely managed object.

fixes [OSD-7012](https://issues.redhat.com/browse/OSD-7012)


cblecker:
/hold cancel

This has been tested on stage, and works as expected.

PR #786, opened 2021-04-16
Fully qualify all velero resource exclusions

cblecker:
This ensures that the correct type of resource is excluded, as well as both API endpoints for the events objects.


gsleeman (approved):
/lgtm

PR #787, opened 2021-04-18
[BLOCKED] fix(RMO): clean config.yaml [OSD-5499]

georgettica:
This is blocked until we fix issues in RMO that make SLO checking stable

according to this comment https://github.com/openshift/managed-cluster-config/pull/779#issue-613984228
it can be removed with no implications as it's in PROD already

### DIFF
--- a/deploy/osd-route-monitor-operator/config.yaml
+++ b/deploy/osd-route-monitor-operator/config.yaml
@@ -1,6 +1,0 @@
-deploymentMode: "SelectorSyncSet"
-selectorSyncSet:
-  matchExpressions:
-  - key: api.openshift.com/environment
-    operator: NotIn
-    values: ["production"]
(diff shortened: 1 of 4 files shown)


dofinn:
> it can be removed with no implications as it's in PROD already

How is it in prod already with this condition still in MCC?

cblecker:
For clarity: this is *not* a noop. This configuration is not in production yet.

Last I heard from @NautiluX and @RiRa12621, there were bugs discovered in the RMO. It should be confirmed these are fixed and promoted before this is pushed to prod.

/hold

georgettica:
Sorry, I looked at MCC and saw the commit was on one of the prod merges
This looked to be a noop.

I am updating the title to reflect it

georgettica:
Solved in a different PR
/close

PR #789, opened 2021-04-20
Reinstall route-monitor-operator

2uasimojo:
CSVs are getting wedged due to a [bad build](https://gitlab.cee.redhat.com/service/saas-route-monitor-operator-bundle/-/blob/staging/route-monitor-operator/0.1.276-8560aa0/route-monitor-operator.clusterserviceversion.yaml#L176).

Some seem to be recovering on their own, but not all.

This CronJob was created following https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/reinstall-operator.md
I tested it on a happy cluster (it correctly shorts out) and a sad one (it resolves the problem).

It *only* needs to go to staging.

### DIFF
--- a/deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
+++ b/deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
@@ -94,7 +94,7 @@ spec:
               COUNT_CSV=$(oc -n "$NAMESPACE" get csv -l operators.coreos.com/route-monitor-operator.openshift-route-monitor-operator= -o name | wc -l) || true
               if [[ $COUNT_CSV -ge 2 ]]; then
                 oc -n "$NAMESPACE" delete csv -l operators.coreos.com/route-monitor-operator.openshift-route-monitor-operator=
-                oc -n "$NAMESPACE" get installplan -o name | xargs oc delete
+                oc -n "$NAMESPACE" delete installplan --all --ignore-not-found
                 touch /tmp/sub.yaml
                 oc -n "$NAMESPACE" get subscription route-monitor-operator -o yaml > /tmp/sub.yaml
                 oc -n "$NAMESPACE" delete -f /tmp/sub.yaml
--- a/deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
+++ b/deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sre-operator-reinstall-sa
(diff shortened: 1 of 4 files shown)


deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
```diff
@@ -0,0 +1,107 @@
...
+                oc -n "$NAMESPACE" get installplan -o name | xargs oc delete
+                touch /tmp/sub.yaml
+                oc -n "$NAMESPACE" get subscription route-monitor-operator -o yaml > /tmp/sub.yaml
+                oc -n "$NAMESPACE" delete -f /tmp/sub.yaml
+                oc -n "$NAMESPACE" apply -f /tmp/sub.yaml
+              fi
+              oc -n "$NAMESPACE" delete cronjob sre-operator-reinstall || true
+              oc -n "$NAMESPACE" delete role sre-operator-reinstall-role || true
+              oc -n "$NAMESPACE" delete rolebinding sre-operator-reinstall-rb || true
+              oc -n "$NAMESPACE" delete serviceaccount sre-operator-reinstall-sa || true
```
cblecker, line 106: Why? These will get removed when the SSS is reverted.

deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
```diff
@@ -0,0 +1,107 @@
...
+            - sh
+            - -c
+            - |
+              #!/bin/bash
+              set -euxo pipefail
+              NAMESPACE=openshift-route-monitor-operator
+              COUNT_CSV=$(oc -n "$NAMESPACE" get csv -l operators.coreos.com/route-monitor-operator.openshift-route-monitor-operator= -o name | wc -l) || true
+              if [[ $COUNT_CSV -ge 2 ]]; then
+                oc -n "$NAMESPACE" delete csv -l operators.coreos.com/route-monitor-operator.openshift-route-monitor-operator=
+                oc -n "$NAMESPACE" get installplan -o name | xargs oc delete
```
cblecker, line 97:
```suggestion
                oc -n "$NAMESPACE" delete installplan --all --ignore-not-found
```

deploy/sre-pruning/115-pruning-RMO-CSVs.Cronjob.yaml
2uasimojo: Just following the SOP :)

I guess it's "in case we forget to revert"?

2uasimojo: tbh, not sure why this needs to be a CronJob as opposed to just a Job -- except "you're not supposed to do that".

PR #790, opened 2021-04-20
Revert "Reinstall route-monitor-operator"

2uasimojo:
Reverts openshift/managed-cluster-config#789

Clusters have recovered.


2uasimojo:
/hold

> Clusters have recovered.

This isn't true yet :)

PR #792, opened 2021-04-21
OSD-7065: Pull osd-cluster-ready by digest

2uasimojo:
NOTE: Until https://issues.redhat.com/browse/OSD-6276 is resolved, we're still building and pushing osd-cluster-ready container images manually, so we should solve the "disappearing digest" problem via "option 2" -- create a tag corresponding to the digest to make sure that digest persists. To go with this PR, I created [quay.io/openshift-sre/osd-cluster-ready:freeze-digest-eb600ec385f76bd1cb746287dafc512ffcdaf09fd7aa7c7d78ac7aed30f665d2](https://quay.io/repository/openshift-sre/osd-cluster-ready?tag=freeze-digest-eb600ec385f76bd1cb746287dafc512ffcdaf09fd7aa7c7d78ac7aed30f665d2&tab=tags)

[OSD-7065](https://issues.redhat.com/browse/OSD-7065)

### DIFF
--- a/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
+++ b/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
@@ -11,14 +11,14 @@ spec:
     template:
         metadata:
             name: osd-cluster-ready
-            annotations:
+            labels:
                 # Keep this in sync with image
-                managed.openshift.io/imageTag: v0.1.45-c9f1f45
+                managed.openshift.io/version: v0.1.45-c9f1f45
         spec:
             containers:
             - name: osd-cluster-ready
-              # Keep the `managed.openshift.io/imageTag` annotation in
-              # sync with this
+              # Keep the `managed.openshift.io/version` label in sync
+              # with this
               image: quay.io/openshift-sre/osd-cluster-ready@sha256:eb600ec385f76bd1cb746287dafc512ffcdaf09fd7aa7c7d78ac7aed30f665d2
               command: ["/root/main"]
             restartPolicy: OnFailure
--- a/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
+++ b/deploy/osd-cluster-ready/60-osd-ready.Job.yaml
@@ -11,10 +11,15 @@ spec:
     template:
         metadata:
             name: osd-cluster-ready
+            annotations:
+                # Keep this in sync with image
+                managed.openshift.io/imageTag: v0.1.45-c9f1f45
         spec:
             containers:
             - name: osd-cluster-ready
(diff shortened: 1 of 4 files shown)


2uasimojo:
Why do I have to `make` explicitly sometimes, but sometimes it happens magically for me??

2uasimojo:
Ah, there it goes.

cblecker:
Didn't we move this job to be administered by the config-alertmanager-operator?

Also, .spec.template is immutable?

2uasimojo:
> Didn't we move this job to be administered by the config-alertmanager-operator?

No. We were gonna, but we ended up not.

> Also, .spec.template is immutable?

What do you mean by immutable? This shouldn't redeploy because [CreateOnly](https://github.com/openshift/managed-cluster-config/blob/ed26922cbd61cea9466f186b0d76346402e102b5/deploy/osd-cluster-ready/config.yaml#L3). Are you saying I'm not allowed to *create* a Job with metadata in it? The [docs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-template) don't say anything about that. Also:

```
[efried@efried managed-cluster-config]$ oc explain job.spec.template
KIND:     Job
VERSION:  batch/v1

RESOURCE: template <Object>

DESCRIPTION:
     Describes the pod that will be created when executing a job. More info:
     https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

     PodTemplateSpec describes the data a pod should have when created from a
     template

FIELDS:
   metadata	<Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata

   spec	<Object>
     Specification of the desired behavior of the pod. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
```

deploy/osd-cluster-ready/60-osd-ready.Job.yaml
```diff
@@ -11,10 +11,15 @@ spec:
     template:
         metadata:
             name: osd-cluster-ready
+            annotations:
+                # Keep this in sync with image
+                managed.openshift.io/imageTag: v0.1.45-c9f1f45
```
2uasimojo, line 16: Let's make this a label, as that seems to be what's done elsewhere.

(Aside: that doesn't make a lot of sense to me, as I can't think of a scenario where you would want/need to be able to search by a version.)

2uasimojo: Done.

jharrington22 (approved):
/lgtm

PR #793, opened 2021-04-21
OSD-7065: Pull token-refresher by digest

2uasimojo:
[OSD-7065](https://issues.redhat.com/browse/OSD-7065)

### DIFF
--- a/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
+++ b/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
@@ -57,7 +57,7 @@ spec:
               key: receiver-url
         - name: ISSUER_URL
           value: "https://sso.redhat.com/auth/realms/redhat-external"
-        image: quay.io/observatorium/token-refresher:master-2021-02-24-1e01b9c
+        image: quay.io/observatorium/token-refresher@sha256:6ce9b80cd1d907cb6c9ed2a18612f386f7503257772d1d88155a4a2e6773fd00
         name: token-refresher
         ports:
         - containerPort: 8080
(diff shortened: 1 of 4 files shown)


2uasimojo:
/hold

We need to figure out who controls the [repository](https://quay.io/repository/observatorium/token-refresher?tag=master-2021-02-24-1e01b9c&tab=tags) for this so we can do whatever is necessary to avoid the "disappearing digest" problem.

jharrington22:
@2uasimojo the digest going missing here would cause the token refresher pod CrashLoopBackoff alert on all OSD prod systems until we update this digest. I think we can move forward with it as long as we document the fix.

Not doing this means we don't get OSD metrics to MST which I guess is not critical but given we know how to fix errors I'd prefer to make this change.

2uasimojo:
> @2uasimojo the digest going missing here would cause the token refresher pod CrashLoopBackoff alert on all OSD prod systems until we update this digest. I think we can move forward with it as long as we document the fix.
> 
> Not doing this means we don't get OSD metrics to MST which I guess is not critical but given we know how to fix errors I'd prefer to make this change.

Ack. Give me a little bit to work with the observatorium team and see how hard it would be to implement "don't rebuild". If it's too hard, I'll document as you say and we'll move on.

2uasimojo:
> work with the observatorium team and see how hard it would be to implement "don't rebuild"

The team is receptive. I've proposed https://github.com/observatorium/token-refresher/pull/6.

2uasimojo:
https://github.com/observatorium/token-refresher/pull/6 just merged. We should update the sha we're using; then this should be ready to go.

2uasimojo:
> We should update the sha

Done

/hold cancel

cblecker:
/lgtm
/approve

Thanks for updating this @2uasimojo !

PR #794, opened 2021-04-22
Add MUO UpgradeConfig CR renaming job

mrbarge:
This is a short-lived job intended to be coupled with [OSD-6918](https://issues.redhat.com/browse/OSD-6918) and the promotion of its [associated MUO PR](https://github.com/openshift/managed-upgrade-operator/pull/224) which involves renaming of the managed-upgrade-operator CR from `osd-upgrade-config` to `managed-upgrade-config`.

This job will recreate the existing `osd-upgrade-config` CR as `managed-upgrade-config` if it exists, and then delete the `osd-upgrade-config` CR. It will only perform this action on clusters where an `osd-upgrade-config` CR already exists.

This will fix the following potential issues which can stem from renaming the managed CR:
- Clusters which are mid-upgrading would not report a completed upgrade to OCM.
- Clusters which are not mid-upgrade but had an existing `osd-upgrade-config` present due to an automatic upgrade policy would have that resource orphaned.

### DIFF
--- a/deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
+++ b/deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
@@ -1,0 +1,66 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-muo-cr-rename-sa
+  namespace: openshift-managed-upgrade-operator
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-muo-cr-rename-role
+  namespace: openshift-managed-upgrade-operator
+rules:
+  - apiGroups:
+      - upgrade.managed.openshift.io
+    resources:
+      - upgradeconfigs
+    verbs:
+      - get
+      - list
+      - delete
+      - create
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: osd-muo-cr-rename-rb
+  namespace: openshift-managed-upgrade-operator
+roleRef:
+  kind: Role
+  name: osd-muo-cr-rename-role
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: osd-muo-cr-rename-sa
+  namespace: openshift-managed-upgrade-operator
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-muo-cr-rename
+  namespace: openshift-managed-upgrade-operator
+spec:
+  failedJobsHistoryLimit: 1
+  ttlSecondsAfterFinished: 100
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          serviceAccountName: osd-muo-cr-rename-sa
+          containers:
+            - name: osd-muo-cr-rename
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -euo pipefail -c
+                - |
+                  for uc in $(oc get upgradeconfig -n openshift-managed-upgrade-operator osd-upgrade-config -o name --ignore-not-found); do
+                     oc delete upgradeconfig -n openshift-managed-upgrade-operator --ignore-not-found=true managed-upgrade-config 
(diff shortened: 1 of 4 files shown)


mrbarge:
Tightly coupled to the promotion of a MUO PR, so putting a hold until we're ready. It's ready for review though.
/hold

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
```diff
@@ -0,0 +1,65 @@
...
+        spec:
+          restartPolicy: Never
+          serviceAccountName: osd-muo-cr-rename-sa
+          containers:
+            - name: osd-muo-cr-rename
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
```
NautiluX, line 60: probably add `-euo pipefail`
I don't see any dangerous statements in the commands, but when chaining oc commands that include delete statements I'd feel more comfortable if the script aborts when one of them fails.

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
```diff
@@ -0,0 +1,65 @@
...
+                - |
+                  for uc in $(oc get upgradeconfig -n openshift-managed-upgrade-operator osd-upgrade-config -o name --ignore-not-found); do
```
NautiluX, line 62: `$uc` doesn't appear to be used, can there be more than one?

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
```diff
@@ -0,0 +1,65 @@
...
+  name: osd-muo-cr-rename-sa
+  namespace: openshift-managed-upgrade-operator
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-muo-cr-rename
+  namespace: openshift-managed-upgrade-operator
+spec:
+  failedJobsHistoryLimit: 5
```
NautiluX, line 44:
```suggestion
  failedJobsHistoryLimit: 1
  ttlSecondsAfterFinished: 100
```
If for some reason the job fails it will trigger an alert when there are too many failed jobs.

NautiluX (changes requested):

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
mrbarge: Thanks, great suggestion - added in commit d901b0f

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
mrbarge: There will be at most one, but there may also be none - the loop was done as a easier way of handling the logic case of handling if there was none at all (in which case it'll just skip over).

NautiluX (commented):
/lgtm

mrbarge:
@NautiluX may I ask if you could re-add your lgtm on this please? No changes since your last review, just a rebase so that I'm not adding the `hack/` files as part of my PR.

deploy/sre-pruning/115-pruning-rename-MUO-CR.CronJob.yaml
```diff
@@ -0,0 +1,66 @@
...
+          containers:
+            - name: osd-muo-cr-rename
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -euo pipefail -c
+                - |
+                  for uc in $(oc get upgradeconfig -n openshift-managed-upgrade-operator osd-upgrade-config -o name --ignore-not-found); do
+                     oc delete upgradeconfig -n openshift-managed-upgrade-operator --ignore-not-found=true managed-upgrade-config
```
bmeng, line 64: why do we try to remove the upgrade config with new name `managed-upgrade-config` here? and recreate it in next step?

mrbarge: This was mainly because I wanted to always treat "osd-upgrade-config" as the source of truth to replace with a "managed-" equivalent, and not whatever "managed-" might be floating around at the same time. This could potentially be important if the cluster happens to be mid-upgrade to preserve what is in its `Status` (although I'd _hope_ that a mid-upgrade cluster doesn't end up with a "managed-" UC). 

Once an existing `managed-*` has been deleted and replaced with the contents of `osd-upgrade-config` and `osd-upgrade-config` the job should not attempt to re-delete on a re-run because `osd-upgrade-config` will no longer exist, immediately short-circuiting the loop.

PR #802, opened 2021-05-02
fix(maxunhealthy): bump count [OSD-6620]

georgettica:
this solves issues with bump upgrade

there would not be a bump from 3 -> 4 as 3 unhealthy nodes mean a bigger problem

### DIFF
--- a/deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
+++ b/deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
@@ -14,4 +14,4 @@ spec:
   - type:    "Ready"
     timeout: "480s"
     status: "Unknown"
-  maxUnhealthy: 2
+  maxUnhealthy: 3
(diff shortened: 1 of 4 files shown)


georgettica:
needs to be tested, after that the WIP will be removed

georgettica:
blocked by code-freeze.

but it's fine because tests were not done yet

georgettica:
stopping the 3 nodes in this case does work.
please review and merge when needed

georgettica:
fixed make.
/retest

georgettica:
no longer in WIP, works

jharrington22:
@georgettica what OCP versions are you targeting?

georgettica:
wasn't specific on versioning

I tested with `4.7.8` (the default version)

@jharrington22 should the version affect this card though?

georgettica:
closing as this is not working.. trying above lastest

jharrington22:
@georgettica the card doesn't mention a specific version. This change will not hit 4.7 clusters which is new the default.

jharrington22:
Closing in favor of #806

PR #804, opened 2021-05-04
Added CSSRE access to managed api service internal

g0ldrush:


NautiluX (changes requested):
need to run make to update syncsets

PR #806, opened 2021-05-05
fix(MaxUnhealthyCompute): bump size[OSD-6620]

georgettica:
change from 2->3. this should freeze operations if 4 nodes restart at the same time. which is a big indication something is wrong


georgettica:
fixes https://github.com/openshift/managed-cluster-config/pull/802

georgettica:
After internal discussion, removing
/unhold

georgettica:
/hold

Maybe not on weekend

jharrington22:
This closes #802 

/hold cancel

PR #807, opened 2021-05-05
Add cronjob to delete leftoever cdo metrics service

dustman9000:
This removes the leftover service for prom metrics from CDO since operator deployments can't delete resources.


jharrington22 (approved):
/lgtm

PR #809, opened 2021-05-06
OSD-6981 All SREP to read some commons resources 

qhua948:


hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -2173,7 +2173,23 @@ objects:
...
+        resources:
+        - pods
+        - services
+        - daemonsets
+        - roles
+        - deployments
+        - replicasets
+        - statefulsets
+        - jobs
+        - cronjobs
```
wanghaoran1988, line 2188: It looks like we already have the perms get/list at cluster scope except roles, could you remove others and update the 'apiGroups" to "rbac.authorization.k8s.io/v1"?

qhua948: @wanghaoran1988 Done

PR #822, opened 2021-05-11
Add clusterrole to provide dedicated-admins with storageclass permissions.

tnierman:
Added clusterrole `osd-gcp-storage-dedicated-admin` to which allows dedicated-admins to modify storageclasses in order to create encrypted PVs for https://issues.redhat.com/browse/OSD-7216.


tnierman:
I believe the `managed.openshift.io/aggregate-to-dedicated-admins: "cluster"` label should be sufficient to apply the RBAC rules to the dedicated-admins group. When testing this, I didn't need to create a clusterrolebinding object.

dustman9000:
> This lgtm in regards to what's needed from the customer perspective to enable encrypted PVs with a CMK.
> Does there need to be a `ClusterRoleBinding` alongside it, or will that come in a separate PR?

Nope! This is taken care of by the rbac-permission-operator: https://github.com/openshift/managed-cluster-config/blob/8e1d9887a8fafd226641ed0aa73fadc0dabe4ca1/deploy/rbac-permissions-operator-config/03-dedicated-admins-cluster.ClusterRole.yaml#L4-L8

wanghaoran1988:
If this meant to be GCP CCS only, should we add a config.yml that includes ? :
```
  api.openshift.com/ccs: "true"
  hive.openshift.io/cluster-platform: gcp
```

dustman9000:
> If this meant to be GCP CCS only, should we add a config.yml that includes ? :
> 
> ```
> 
>   api.openshift.com/ccs: "true"
> 
>   hive.openshift.io/cluster-platform: gcp
> 
> ```

Yes and the config.yaml already has those conditions.

dustman9000:
@tnierman - In deploy/osd-gcp-ssd-storage/config.yaml, can you also remove the line `applyBehavior: "CreateOnly"` since these permissions have the potential of removing the pd-ssd StorageClass.

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -7135,6 +7135,19 @@ objects:
     resourceApplyMode: Sync
     applyBehavior: CreateOnly
```
cblecker, line 7136: Because it's in this folder, it will only be created, not updated. I don't think that's our intention. We probably want to make a subfolder?

cblecker:
> can you also remove the `applyBehavior: "CreateOnly"` line

I think that part is required for the storage class itself, as it's immutable, right?

dustman9000:
> > can you also remove the `applyBehavior: "CreateOnly"` line
> 
> I think that part is required for the storage class itself, as it's immutable, right?

It was intended to allow a dedicated-admin to "modify" the StorageClass, but yeah, since StorageClasses are immutable it makes sense to leave it and create the subdir.

tnierman:
Somehow I screwed up the git rebase badly enough that I think it'll just be easier to close this and recreate a fresh PR. I'll link it here once I've remade

tnierman:
https://github.com/openshift/managed-cluster-config/pull/825

PR #824, opened 2021-05-12
Fixup leftover CIO metrics service

cblecker:
In https://github.com/openshift/cloud-ingress-operator/pull/177, the metrics service for CIO was renamed. This created a duplicate leftover service in some clusters, and alerts that CIO was missing in others. This PR creates a CronJob to cleanup the leftover service, and ensure that the correct service is being observed for the CIO alerts.


fahlmant (approved):
/lgtm

PR #827, opened 2021-05-13
Remove CDO metrics service pruner

dustman9000:
This removes the CDO metrics service pruner as it's already been deployed and ran against all envs.


fahlmant (approved):
/lgtm

PR #828, opened 2021-05-13
Add pruning for duplicate CIO servicemonitor

cblecker:
Follow up to #824 and https://github.com/openshift/cloud-ingress-operator/pull/177 to remove the matching service monitor


fahlmant (approved):
/lgtm

PR #832, opened 2021-05-16
SRE should be able to patch configmaps

sedroche:


sedroche:
cc @wanghaoran1988

cblecker:
@sedroche @wanghaoran1988 Why is this needed? It's not linked to a JIRA card, and typically, an SRE should not be patching config maps.

sedroche:
> @sedroche @wanghaoran1988 Why is this needed? It's not linked to a JIRA card, and typically, an SRE should not be patching config maps.

@cblecker Appologies. Some more details should have been given.
https://issues.redhat.com/browse/OHSS-4208
Over the weekend I had to elevate to add an alert to managed upgrade operators ignore list.
Maybe we can reduce scope to that cm?

PR #834, opened 2021-05-18
[OSD-7254] Added resources for RHODS resources in managed clusters

qhua948:


qhua948:
Replaces #820

cblecker:
/close

I believe the GPU operator is going to be done as an addon

PR #838, opened 2021-05-24
Fix ClusterResourceQuota and ResourceQuota SelectorSyncSets To Be Specific Per StorageClass

dustman9000:
This makes the storage and load-balancer quota maintenance a little less painful. 

This PR achieves the following:
* Use specific storage classes for storage ClusterResourceQuota such that OCS SCs can be used on non-CCS clusters.
* Move static load-balancer and storage resources out to `deploy/resource-quotas/load-balancer` and `deploy/resource-quotas/storage` respectively.
* Generate quotas dynamically upon top-level make
* Put quota counts into project.mk to be consistent with other folders.

JIRA: https://issues.redhat.com/browse/OSD-6674


hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -41,10 +41,1214 @@ parameters:
 metadata:
   name: selectorsyncset-template
 objects:
+- apiVersion: hive.openshift.io/v1
```
dustman9000, line 44: This block just moved up due to how the SSS's are appended now. Nothing changed with the identity SSS.

cblecker:
/hold
for review

I'm not a huge fan of embedding hard coded strings like storage class names in the existing python script for generating things.

Should the generated storage classes bit be pulled out into a separate script/make target that can generate things into the deploy directory? (similar to what we do with generating the oauth templates)

dustman9000:
> /hold
> for review
> 
> I'm not a huge fan of embedding hard coded strings like storage class names in the existing python script for generating things.
> 
> Should the generated storage classes bit be pulled out into a separate script/make target that can generate things into the deploy directory? (similar to what we do with generating the oauth templates)

I've addressed your concerns and moved the resource generation out to `deploy/resource-quotas/load-balancer` and `deploy/resource-quotas/storage` directories. I also pulled out the hard-coded quota counts and put them in a project.mk to be consistent with the other folders.

dustman9000:
Closing to address the issue in a smaller incremental approach.

PR #839, opened 2021-05-24
Allow Dedicated Admins To Annotate NetNamespaces

dustman9000:
This will allow dedicated admins to annotate netnamespaces resources.

Example for enabling multicast:
```
oc annotate netnamespace my-ns netnamespace.network.openshift.io/multicast-enabled=true
```

Depends on https://github.com/openshift/managed-cluster-validating-webhooks/pull/159 so that dedicated-admins cannot modify protected namespaces.

JIRA: https://issues.redhat.com/browse/OSD-7367


dustman9000:
Validating webhooks added and merged to support this.
/hold cancel

lisa (approved):
/lgtm

PR #848, opened 2021-06-07
add two SLO required metrics to MST

RiRa12621:
Those metrics allows us to drill down into cluster availability for the fleet without being dependant on telemeter.

### DIFF
--- a/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -16,7 +16,7 @@ data:
           writeRelabelConfigs:
           - sourceLabels: [__name__]
             action: keep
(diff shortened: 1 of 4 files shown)


deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
```diff
@@ -16,7 +16,7 @@ data:
           writeRelabelConfigs:
           - sourceLabels: [__name__]
             action: keep
-            regex: '(cluster_admin_enabled|identity_provider)'
+            regex: '(cluster_admin_enabled|uster_operator_conditions|dentity_provider|probe_success)'
```
cblecker, line 19: Looks like a missing "c" in this line

cblecker (changes requested):

PR #853, opened 2021-06-09
add configurable ignored namespace for MUO

bmeng:


mrbarge:
/hold
/lgtm
Feel free to unhold when the corresponding MUO PR is merged.

mrbarge:
/hold
as per above.

bmeng:
cc @NautiluX 
Currently, there are multiple RMO related alerts which blocked the whole manged upgrade. We are going to ignore the RMO related alerts to make the upgrade going on as we assume the alerts there should not be blocker for the upgrade.
Please help check what is wrong for the RMO in e2e testing.

mrbarge:
/hold
until 4.5 and 4.6 configmaps are also updated.

PR #860, opened 2021-06-14
Migrate OSD specific CIO configurations to MCC

cblecker:
ref [OSD-7345](https://issues.redhat.com/browse/OSD-7345)

This migrates the APIScheme and SSHD configuration CRs to the MCC repo from the cloud-ingress-operator repo

### DIFF
--- a/deploy/cloud-ingress-operator-configuration/sshd/config.yaml
+++ b/deploy/cloud-ingress-operator-configuration/sshd/config.yaml
@@ -1,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: api.openshift.com/sts
+    operator: NotIn
+    values: ["true"]
+  - key: api.openshift.com/private-link
+    operator: NotIn
+    values: ["true"]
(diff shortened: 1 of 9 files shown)


cblecker:
/hold
Need app-interface changes before this merges

cblecker:
/hold cancel
app-interface MR has merged

deploy/cloud-ingress-operator-configuration/sshd/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
```
jharrington22, line 1: Does this need to be `Upsert` I guess not since we're creating it.

jharrington22: No it doesn't since we're just creating here.

jharrington22: And the sync set in CIO is already [upsert](https://github.com/openshift/cloud-ingress-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L50) in preparation for the removal

PR #866, opened 2021-06-22
[Osd 7542] don't alert on cio for private-link clusters

UlrichSchlueter:
https://issues.redhat.com/browse/OSD-7542

Adding condition to SSS for the CloudIngressOperatorOfflineSRE prometheus rule so it does not get synced to an aws private-link cluster to avoid false alerts like https://redhat.pagerduty.com/incidents/POKJQ99

Match conditions tested on a private hive/workload cluster pair.


deploy/sre-prometheus/aws/sts-privatelink/config.yaml
```diff
@@ -4,5 +4,8 @@ selectorSyncSet:
   - key: api.openshift.com/sts
     operator: NotIn
     values: ["true"]
+  - key: api.openshift.com/private-link
```
fahlmant, line 7: I believe this needs an OR as part of the match

UlrichSchlueter: Will retest the combinations of sts/private-link =true/false/not set. 
If the above needs and OR, do you think these conditions need to change, too : https://github.com/openshift/managed-cluster-config/blob/master/deploy/cloud-ingress-operator-configuration/sshd/config.yaml

fahlmant: Actually after looking at this for now, I think it will work as expected.

UlrichSchlueter: for completeness: the test matrix for the conditions:

image

fahlmant (approved):
/lgtm

PR #871, opened 2021-06-30
Reduce prometheus retention

dofinn:
Currently we deploy OSD with 100GB volumes for prometheueses. This is regardless of the clusters size and workload activity. 

We are now starting to receive alerts for `KubePersistentVolumeFillingUp` where the short term solution is editing this CM to reduce the retention and have prom truncate some data to reclaim space. 

If the prom disks fill up, some awful things can happen -> https://issues.redhat.com/browse/OHSS-4541 

Given that these metrics are utilized (underutilized?) by SRE, We should reduce retention to 11 days until we have a more resilient provisioning policy in place that takes into consideration potential cluster characteristics. https://issues.redhat.com/browse/OSD-7629 

I did not update this in user-workload-monitoring as SRE is not utilizing those metrics. We in future should test what happens when UWM is bombed by metrics to assure that openshift-monitoring stays available and consumable.

### DIFF
--- a/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -31,7 +31,7 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/infra
           operator: Exists
-      retention: 15d
+      retention: 11d
       volumeClaimTemplate:
         metadata:
           name: prometheus-data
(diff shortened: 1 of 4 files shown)


mmazur:
I agree with the idea. Until we have better sizing policy, we should prioritize avoiding incidents.

RiRa12621:
/lgtm
Good enough as temp fix

PR #879, opened 2021-07-08
allow dedicated-admins to delete pods in openshift-customer-monitoring

maorfr:
when growing storage for Prometheus operator, the storage is not getting automatically updated in the PVCs, perhaps due to https://github.com/prometheus-operator/prometheus-operator/issues/4079.

to overcome this, we need to manually edit PVCs (dedicated-admins have that power) and delete pods one by one. we do not have permissions to delete pods. instead, we currently delete the StatefulSet, which is re-created by the operator, thus getting new pods.

instead and to prevent downtime, it would be useful to have the permissions to delete pods in the openshift-customer-monitoring namespace.

thank you!

### DIFF
--- a/deploy/osd-customer-monitoring/05-role.yaml
+++ b/deploy/osd-customer-monitoring/05-role.yaml
@@ -15,6 +15,12 @@ rules:
   - get
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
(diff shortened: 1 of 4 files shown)


wanghaoran1988:
/approve
They are already to able delete a lot CRs, it make sense to add "delete” pods perms.

PR #883, opened 2021-07-19
[OSD-7437] Create alert for image-pruner to trigger if it has failed more than two times

dastergon:
In my PR on [configure-alertmanager-operator](https://github.com/openshift/configure-alertmanager-operator/pull/173), I disabled the upstream KubeJobFailed alert for image-pruner because a single-failed job causes noise and it isn't an indication of an issue with the setup. Consequently, after looking at clusters with such image-pruner jobs, I introduce a custom KubeJobFailed alert for image-pruner that counts the number of failed jobs, and if we have more than 2 failed jobs for 30m we trigger the alert. Note: The image-pruner job runs every hour.

PTAL

See: https://issues.redhat.com/browse/OSD-7437

### DIFF
--- a/deploy/sre-prometheus/100-sre-image-pruner.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-image-pruner.PrometheusRule.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-image-pruning
+    role: alert-rules
+  name: sre-image-pruning
+  namespace: openshift-image-registry
+spec:
+  groups:
+  - name: sre-image-pruning
+    rules:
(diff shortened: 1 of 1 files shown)


RiRa12621:
/hold
can we maybe adjust https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-sre-pruning.PrometheusRule.yamlto fully cover our use case

mmazur:
Agree that we should probably reuse the existing alert.

First thought is to just fold it into PruningCronjobErrorSRE, but that's a critical and this is a warning.

Should it be enough to rename this alert to PruningCronjobErrorSRE WARNING? (Can you have two alerts with the same name and different severities?)

cblecker:
@dastergon What's the status on this?

dastergon:
@cblecker the PR was created during the Atlas trial period, but it didn't progress ever since. I will have a look at this during this sprint.

PR #884, opened 2021-07-20
Adds label selector for OSD Logging support

iamkirkbater:
Satisfies [OSD-7564](https://issues.redhat.com/browse/OSD-7564)

Converts from version selector to label selector for extended logging support, enabling those clusters who do have logging to upgrade and continue to receive support.

This label has already been added to the necessary clusters.

### DIFF
--- a/deploy/osd-logging/supported/config.yaml
+++ b/deploy/osd-logging/supported/config.yaml
@@ -1,14 +1,8 @@
 deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   matchExpressions:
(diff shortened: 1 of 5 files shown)


cblecker:
@iamkirkbater what's the status of this?

iamkirkbater:
Should be ready to go.  (Not sure why I never saw the update on this), but I think it should do what we're expecting it to do.

iamkirkbater:
The last of the CDs that need the label have been patched, so this should be :100: ready to go now.

jharrington22 (approved):
/lgtm

PR #886, opened 2021-07-21
OSD-7084 alert when a certificate is about to expire

yithian:
sets up an alert when a certificate managed by certman-operator is about to expire


yithian:
/hold

blocked on https://github.com/openshift/certman-operator/pull/201

yithian:
this needs to go in app-interface

PR #887, opened 2021-07-26
[OSD-7830] ignore openshift-cnv namespace for healthcheck

a7vicky:
This PR adds the `openshift-cnv` namespace to the managed-upgrade-operator's cluster health check ignore list.

Ref: [OSD-7830](https://issues.redhat.com/browse/OSD-7830)


bmeng:
not sure if old versions need to be covered, but it should be harmless.

/lgtm

PR #891, opened 2021-08-04
add permissions for managed kafka rts

StevenTobin:
Adding permissions for mkrts backplane access, based primarily on this doc https://source.redhat.com/groups/public/openshiftplatformsre/wiki/backplane_cluster_permissions

### DIFF
--- a/deploy/backplane/mkrts/redhat-kas-fleetshard/10-mkrts-fleetshard-admins.SubjectPermission.yml
+++ b/deploy/backplane/mkrts/redhat-kas-fleetshard/10-mkrts-fleetshard-admins.SubjectPermission.yml
@@ -1,0 +1,12 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: backplane-mkrts-admins
+  namespace: openshift-rbac-permissions
+spec:
+  permissions:
+  - allowFirst: true
+    clusterRoleName: admin
+    namespacesAllowedRegex: (^redhat-managed-kafka-.*|^redhat-kas-fleetshard-.*)
+  subjectKind: Group
+  subjectName: system:serviceaccounts:openshift-backplane-rts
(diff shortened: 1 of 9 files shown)


deploy/backplane/mkrts/redhat-kas-fleetshard/10-mkrts-fleetshard-admins.SubjectPermission.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: backplane-mkrts-admins
+  namespace: openshift-rbac-permissions
+spec:
+  permissions:
+  - allowFirst: true
+    clusterRoleName: admin
+    namespacesAllowedRegex: (^redhat-managed-kafka-.*|^redhat-kas-fleetshard-.*)
```
david-martin, line 10:
```suggestion
    namespacesAllowedRegex: (^redhat-managed-kafka-.*|^redhat-kas-fleetshard-.*|^managed-application-services-observability)
```

wanghaoran1988: @david-martin Could you clarify what managed-application-services-observability ns does, why it's not name with prefix "redhat-"?

StevenTobin: The managed-application-services-observability namespace is where the observability stack that monitors the managed-kafka components is deployed.

david-martin: > why it's not name with prefix "redhat-"?

@wanghaoran1988 I can't recall the reason for that name being used. 
Is it a problem do you think?
I know there are some references to `redhat-` in other places, similar to `openshift-` that maybe mean we should change the prefix? (dedicated-admin/reader role i think?)

@pb82 do you recall the reason for the name?

pb82: I can't recall a particular reason. The observabilit stack can be installed in any namespace, but as you said, there might be other namespace filters that expect a `redhat-` prefix.

wanghaoran1988: @david-martin I am asking because the ns used by addon service on osd all follow the convention to start with "redhat-", if this is just an exception, I am ok with that.

PR #896, opened 2021-08-11
Use a merge patch to remove Config.spec.nodeSelector

2uasimojo:
The osd-registry SelectorSyncSet has been failing to apply with a status message like:

`failed to apply patch 1: the server rejected our request due to an error in our request`

We think this is due to the fact that a JSON patch asking to `remove` a field will fail if that field is already absent.

Change to a `merge` patch setting the field to `null` instead, which has the same effect, but idempotently.

[OSD-7926](https://issues.redhat.com/browse/OSD-7926)


cblecker:
/lgtm
/approve

Thanks Eric!

PR #903, opened 2021-08-19
Add RBAC to allow dedicated-admins to patch kube-apiserver-to-kubelet-signer on 4.5/4.6

cblecker:
fixes [OSD-8153](https://issues.redhat.com/browse/OSD-8153)

There is a use case for a customer to be able to self-serve patch the secret object `kube-apiserver-to-kubelet-signer` located in the `openshift-kube-apiserver-operator` namespace.

Allowing this would permit a dedicated-admin to run the following command, and rotate the certificates early:
```
oc annotate -n openshift-kube-apiserver-operator secret/kube-apiserver-to-kubelet-signer auth.openshift.io/certificate-not-after-
```

This is only required on 4.5 and 4.6, as starting in 4.7, the certificate rotation doesn't impact the cluster's nodes ([RFE-1212](https://issues.redhat.com/browse/RFE-1212)).

References:
- https://access.redhat.com/solutions/5319661

### DIFF
--- a/deploy/rbac-rotate-certs/15-dedicated-admins-cert-secret.Role.yaml
+++ b/deploy/rbac-rotate-certs/15-dedicated-admins-cert-secret.Role.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: dedicated-admins-cert-secret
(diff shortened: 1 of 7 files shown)


cblecker:
/hold
Need to update PR description

PR #908, opened 2021-08-23
OSD-7916 cronjob to clean backplane script resources

feichashao:
For [OSD-7916](https://issues.redhat.com/browse/OSD-7916). Place a cronjob in OSD clusters to clean up backplane script related resources.

Assumptions:
- pod/sa/role/rolebinding/clusterrole/clusterrolebinding share the same name, like `openshift-job-czxtr`.
  - Deleting the resource using label-selector needs 2 requests: 1 GET request to get the name and namespace and 1 DELETE request to delete the resource.
  - It will find all the resources using label-seletor in one-go.
  - It will delete the resource using the name directly, then it can save 1 request per resource.
- Some pod/sa/role/rolebinding/.. can be missing.
  - If a node got drain, the pod in that node will be deleted automatically.
  - User might delete some of the resources.

Potential issues:
- The cronjob run every 24hrs. If the bp script finishes shortly before the cronjob, the user might not have enough time to retrieve the corresponding logs.
  - Possible solution is to add an annotation like `delete-not-before:time`.
- The cronjob will clean-up all resources which the corresponding pod is not in Running state. It also delete Pending pods because Pending usually indicate there are some problems. If the cronjob runs right after the user run a backplane script, the Pending pod will be deleted unexpectedly.
  -  Possible solution is to add a threshold of the pod age.

### DIFF
--- a/deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
+++ b/deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
@@ -1,0 +1,65 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-backplane
+  namespace: openshift-backplane-managed-scripts
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-delete-backplane-script-resources
+  namespace: openshift-backplane-managed-scripts
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - serviceaccounts
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: osd-delete-backplane-script-resources
+  namespace: openshift-backplane-managed-scripts
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: osd-delete-backplane-script-resources
+subjects:
+- kind: ServiceAccount
+  name: osd-backplane
+  namespace: openshift-backplane-managed-scripts
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-delete-backplane-script-resources
+rules:
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - roles
+      - rolebindings
+      - clusterroles
+      - clusterrolebindings
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: osd-delete-backplane-script-resources
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: osd-delete-backplane-script-resources
+subjects:
+- kind: ServiceAccount
+  name: osd-backplane
+  namespace: openshift-backplane-managed-scripts
--- a/deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
+++ b/deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
@@ -1,0 +1,125 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-delete-backplane-script-resources
+  namespace: openshift-backplane-managed-scripts
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "42 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - preference:
+                  matchExpressions:
+                  - key: node-role.kubernetes.io/infra
+                    operator: Exists
+                weight: 1
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/infra
+              operator: Exists
+          restartPolicy: Never
+          serviceAccountName: osd-backplane
+          containers:
+            - name:  osd-delete-backplane-script-resources
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  set -euxo pipefail
+
+                  # Namespace which pods run in
+                  NS="openshift-backplane-managed-scripts"
+                  # Label which all backplane script resources should have
+                  LABEL="managed.openshift.io/backplane-job-uuid"
+                  # Only delete resources created before this time, in unix epoch seconds.
+                  DEL_ONLY_BEFORE=$(($(date +%s) - 3600*24))
+
+                  ### Gather data
+                  # Format: pod_name pod_status creation_timestamp
+                  PODS=$(oc get pod -n $NS --selector $LABEL -o custom-columns=NAME:.metadata.name,PHASE:.status.phase,CT:.metadata.creationTimestamp --no-headers)
+
+                  # Format: sa_name
+                  SAS=$(oc get serviceaccount -n $NS --selector $LABEL -o custom-columns=NAME:.metadata.name --no-headers)
+
+                  # Format: role_name role_ns
+                  ROLES=$(oc get role -A --selector $LABEL -o custom-columns=NAME:.metadata.name,NS:.metadata.namespace --no-headers) 
+
+                  # Format: rolebinding_name rolebinding_ns
+                  ROLEBINDINGS=$(oc get rolebinding -A --selector $LABEL -o custom-columns=NAME:.metadata.name,NS:.metadata.namespace --no-headers)
+
+                  # Format: clusterrole_name
+                  CROLES=$(oc get clusterrole --selector $LABEL -o custom-columns=NAME:.metadata.name --no-headers)
+
+                  # Format: clusterrolebinding_name
+                  CROLEBINDINGS=$(oc get clusterrolebinding --selector $LABEL -o custom-columns=NAME:.metadata.name --no-headers)
+
+                  # Pick up those resources we don't want to delete.
+                  # - Current running pods and related resources.
+                  # - pods created within THRESHOLD_PERIOD and related resouces.
+                  PODS_TO_KEEP=$(awk -v before=$DEL_ONLY_BEFORE \
+                    '{
+                    if($2 ~ /Running/){print $1} 
+                    else{cmd = sprintf("date +\"%%s\" --date=%s", $3); cmd|getline unix_sec; close(cmd); if(unix_sec+0 > before+0){print $1}}
+                    }' \
+                    <<< "$PODS")
+
+                  PODS_TO_DEL=$(awk '{print $1}' <<< "$PODS")
+
+                  for pod in $PODS_TO_KEEP
+                  do
+                    PODS_TO_DEL=$({ grep -v "$pod" || test $? = 1; } <<< "$PODS_TO_DEL")
+                    SAS=$({ grep -v "$pod" || test $? = 1; } <<< "$SAS")
+                    ROLES=$({ grep -v "$pod" || test $? = 1; } <<< "$ROLES")
+                    ROLEBINDINGS=$({ grep -v "$pod" || test $? = 1; } <<< "$ROLEBINDINGS")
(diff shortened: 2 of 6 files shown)


feichashao:
/hold

Hi @qhua948  could you help a look and let's have a discussion?

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+metadata:
+  name: osd-delete-backplane-script-resources-rb
+  namespace: openshift-backplane-managed-scripts
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: osd-delete-backplane-script-resources-role
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: User
```
qhua948, line 34: Shouldn't this be ServiceAccount here?

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-delete-backplane-script-resources
+  namespace: openshift-backplane-managed-scripts
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "* * */1 * *"
```
qhua948, line 10: I think this means it runs every minute?

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+                  NS="openshift-backplane-managed-scripts"
+                  LABEL="managed.openshift.io/backplane-job-uuid"
+
+                  ### Gather data
+                  # Format: pod_name pod_status
+                  PODS=$(oc get pod -n $NS --selector $LABEL -o custom-columns=NAME:.metadata.name,PHASE:.status.phase --no-headers)
+
+                  # Format: pod_name(running)
+                  PODS_RUNNING=$(echo "$PODS" | { grep -i Running || test $? = 1; } | awk '{print $1}')
```
qhua948, line 47:
```                   
PODS_RUNNING=$({ grep -i Running || test $? = 1; } <<<"$PODS" | awk '{print $1}')
```

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+                  # Format: pod_name(non-running)
+                  PODS_NOT_RUNNING=$(echo "$PODS" | { grep -v -i Running || test $? = 1; } | awk '{print $1}')
```
qhua948, line 50: Similar here

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  # Format: clusterrole_name
+                  CROLES=$(oc get clusterrole --selector $LABEL -o custom-columns=NAME:.metadata.name --no-headers)
+
+                  # Format: clusterrolebinding_name
+                  CROLEBINDINGS=$(oc get clusterrolebinding --selector $LABEL -o custom-columns=NAME:.metadata.name --no-headers)
+
+                  ## We don't want to delete resources related to a running pod.
+                  for pod in $PODS_RUNNING
+                  do
+                    SAS=$(echo "$SAS" | grep -v "$pod")
```
qhua948, line 70: `$({ grep -v || :; } <<< $"SAS")` To guard against `-e` failures

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+                    ROLES=$(echo "$ROLES" | grep -v "$pod")
+                    ROLEBINDINGS=$(echo "$ROLEBINDINGS" | grep -v "$pod")
+                    CROLES=$(echo "$CROLES" | grep -v "$pod")
+                    CROLEBINDINGS=$(echo "$CROLEBINDINGS" | grep -v "$pod")
+                  done
+
+                  ## Delete non-running pods
```
qhua948, line 77: I think we should implement some grace period here, because in the worse case, a just finished pod can be deleted immediately.

One idea is to check against the creation time of the pod and if it say <= `cron period` or some other cutoff, then delete it

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  for pod in $PODS_NOT_RUNNING
+                  do
+                    oc delete pod -n $NS $pod
+                  done
+
+                  ## Delete Rolebindings
+                  [ -z "$ROLEBINDINGS" ] || echo "$ROLEBINDINGS" | while read line
```
qhua948, line 84: use `[[ ]]` since this is bash

qhua948: `while read line<<< "$ROLEBINDINGS"`

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  do
+                    rb_name=$(echo $line | awk '{print $1}')
```
qhua948, line 86: Similar here

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  done
+
+                  ## Delete ClusterRolebindings
+                  for clb in $CROLEBINDINGS
+                  do
+                    oc delete clusterrolebinding $clb
+                  done
+
+                  ## Delete Roles
+                  [ -z "$ROLES" ] || echo "$ROLES" | while read line
```
qhua948, line 98: as abv

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
qhua948: This could be hard to do in bash :/

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
feichashao: indeed, should make it `0 0 */1 * *`.   Or better pick a random time of the day.

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
feichashao: Copied from other files and it worked... Seems we can specify it as a user then use a long name `system:serviceaccount:<ns>:<name>`.  But according to https://kubernetes.io/docs/reference/access-authn-authz/rbac/  we should specify it as `ServiceAccount` as the official documented way.   I will change that.

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
feichashao: use `test $? = 1` here in case `grep` exits with 2, which could be other errors than no match.

feichashao:
@wanghaoran1988  PTAL when you have time, thanks!

deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml
```diff
@@ -0,0 +1,125 @@
...
+                  matchExpressions:
+                  - key: node-role.kubernetes.io/infra
+                    operator: Exists
+                weight: 1
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/infra
+              operator: Exists
+          restartPolicy: Never
+          serviceAccountName: osd-delete-backplane-script-resources-sa
```
wanghaoran1988, line 28: The name can be shorter, how about osd-backplane?

wanghaoran1988:
Here is how we clean up the SAs https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml#L41

feichashao:
> Here is how we clean up the SAs https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml#L41

Thank you!  The current implementation use the creationtimestamp + 24hr to determine whether to delete a resource.   In the future if we need more flexibility in per resource level, we can let backplane-api to put a label, and the cronjob delete a resource base on the label.

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+  name: osd-delete-backplane-script-resources-role
+subjects:
+- kind: ServiceAccount
+  name: osd-backplane
+  namespace: openshift-backplane-managed-scripts
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-delete-backplane-script-resources-crole
```
wanghaoran1988, line 40: s/crole/role

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-backplane
+  namespace: openshift-backplane-managed-scripts
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-delete-backplane-script-resources-role
```
wanghaoran1988, line 10: remove "-role"

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - serviceaccounts
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: osd-delete-backplane-script-resources-rb
```
wanghaoran1988, line 26: remove "-rb"

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+  namespace: openshift-backplane-managed-scripts
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: osd-delete-backplane-script-resources-role
```
wanghaoran1988, line 31: remove "-role"

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - clusterrolebindings
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: osd-delete-backplane-script-resources-crb
```
wanghaoran1988, line 57: remove "-crb"

deploy/osd-delete-backplane-script-resources/10-delete-backplane-script-resources.rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: osd-delete-backplane-script-resources-crole
```
wanghaoran1988, line 61: remove "-crole"

PR #910, opened 2021-08-25
Add permissions to policy

sam-nguyen7:
OSD-7718 Add private link permissions to required permissions

`"ec2:CreateVpcEndpointServiceConfiguration",
"ec2:DeleteVpcEndpointServiceConfigurations",
"ec2:DescribeVpcEndpointServiceConfigurations",
"ec2:DescribeVpcEndpointServices",
"ec2:ModifyVpcEndpointServicePermissions"`


fahlmant (changes requested):
Given that PrivateLink is optional, I think another file should be made, maybe another directory even, with the specific role for PrivateLInk, instead of adding it all to one role

PR #913, opened 2021-08-28
Fix versions of cluster-monitoring-operator config

cblecker:
This swaps around the version selections statements for the cluster-monitoring-operator configuration, so that we can support 4.9+

### DIFF
--- a/deploy/cluster-monitoring-config-non-uwm/config.yaml
+++ b/deploy/cluster-monitoring-config-non-uwm/config.yaml
@@ -3,8 +3,8 @@ selectorSyncSet:
   matchLabelsApplyMode: "OR"
   matchExpressions:
     - key: hive.openshift.io/version-major-minor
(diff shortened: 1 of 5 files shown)


deploy/cluster-monitoring-config-non-uwm/config.yaml
```diff
@@ -3,8 +3,8 @@ selectorSyncSet:
   matchLabelsApplyMode: "OR"
   matchExpressions:
     - key: hive.openshift.io/version-major-minor
-      operator: NotIn
-      values: ["4.6", "4.7", "4.8"]
+      operator: In
+      values: ["4.5"]
```
fahlmant, line 7: how many 4.5s do we have left? This specific condition can hopefully go away soon

fahlmant (approved):
/lgtm

PR #917, opened 2021-09-02
OSD-6030: Remove OpenShift_SRE from non-hive clusters

chopraapooja:


deploy/sre-pruning/120-pruning.openshiftsreidp.CronJob.yaml 
```diff
@@ -0,0 +1,43 @@
...
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            args:
+            - /bin/bash
+            - -c
+            - >-
+              oc get users | grep OpenShift_SRE | awk '{
+                users[i++]=$1; for(i in users)
+                {
+                  print "oc delete", users[i], "--as=backplane-cluser-admin"
```
wanghaoran1988, line 40: We shouldn't use --as=backplane-cluter-admin, instead, you should bind the sre-pruner-sa related permissions.

chopraapooja: done

deploy/sre-pruning/120-pruning.openshiftsreidp.CronJob.yaml 
```diff
@@ -0,0 +1,41 @@
...
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            args:
+            - /bin/bash
+            - -c
+            - >-
+              oc get users | grep OpenShift_SRE | awk '{
+                users[i++]=$1; for(i in users)
+                {
+                  print "oc delete", users[i]
```
wanghaoran1988, line 40:
```suggestion
                  print "oc delete user", users[i] 
```

wanghaoran1988:
Please create a clusterrole and clusterrolebinding with this:
```
kind: ClusterRole
metadata:
  name: sre-pruner-sre-users
rules:
# deployment pruning
- apiGroups:
  - "user.openshift.io"
  resources:
  - users
  verbs:
  - list
  - get
  - delete

kind: ClusterRoleBinding
metadata:
  name: sre-pruner-users-pruning
roleRef:
  kind: ClusterRole
  name: sre-pruner-sre-users
  apiGroup: rbac.authorization.k8s.io
subjects:
- kind: ServiceAccount
  name: sre-pruner-sa
  namespace: openshift-sre-pruning

```

wanghaoran1988:
/approve
@cblecker I added some more commits to make this works, has been tested also, is there any reason that you put a hold here?

deploy/sre-pruning/120-pruning.openshiftsreidp.CronJob.yaml 
```diff
@@ -0,0 +1,41 @@
...
+            imagePullPolicy: Always
+            args:
+            - /bin/bash
+            - -c
+            - >-
+              oc get users | grep OpenShift_SRE | awk '{
+                users[i++]=$1; for(i in users)
+                {
+                  print "oc delete user", users[i] 
+                }}' | sh
```
cblecker, line 41:
```suggestion
             oc get users | grep Google | awk '{print $1}' | xargs oc delete users --ignore-not-found --wait=false
```

This pipe is simpler, and idempotent.

wanghaoran1988: thanks, but we should grep "OpenShift_SRE" not "Google".

wanghaoran1988: This has some problems when there is no users matched in the cluster:
```
 oc get users | grep OpenShift_SRE | awk '{print $1}' | xargs oc delete users --ignore-not-found --wait=false
error: resource(s) were provided, but no name was specified

```
I will keep use this script, and update the rbac.

wanghaoran1988:
Tested on staging env.
/lgtm
/hold cancel

PR #919, opened 2021-09-02
Rollout dynamic system reserved sizing to all 4.9+ clusters

cblecker:
Fixes [OSD-6857](https://issues.redhat.com/browse/OSD-6857)


deploy/kubelet-config/01-kubelet-config.yaml
```diff
@@ -1,9 +1,10 @@
...
 metadata:
-  name: system-reserved-worker
+  name: custom-kubelet
 spec:
-  autoSizingReserved: true
   machineConfigPoolSelector:
-    matchLabels:
-      pools.operator.machineconfiguration.openshift.io/worker: ""
+    matchExpressions:
+    - key: machineconfiguration.openshift.io/mco-built-in
```
jharrington22, line 8: What is this label?

jharrington22: Otherwise looks good.

cblecker: it's the common built in label for both the master and worker MCO groups: `oc get machineconfigpool --show-labels`

jharrington22: This is safe for masters? I thought we were excluding masters?

cblecker: Yes, it is safe for masters. I confirmed that with the node team. We are not currently excluding masters for DPTP

PR #921, opened 2021-09-06
OSD-6381 remove ssh on non-hive clusters

feichashao:
WIP: need to test and verify it.

- Limit the SSHD resource to hive only.
- Move the namespace file `00-openshift-sre-sshd.Namespace.yaml` to the same directory as the SSHD resource, as only the SSHD resource using that namespace.


fahlmant:
/hold
Can you please test this on a cluster and see how cloud-ingress-operator reacts?

fahlmant:
Removing the namespace could cause issues for the operator as well. Could you also test deleting the ns and looking at the logs?

feichashao:
> Removing the namespace could cause issues for the operator as well. Could you also test deleting the ns and looking at the logs?

@fahlmant  I did a try in my environment, I deleted the ns `openshift-sre-sshd` but i don't see any error in the CIO log. I remove the CIO pod to let it re-create and the CIO pod started successfully and I didn't see anything special.  I will not delete the ns in the PR though, let's have a discussion offline, maybe i missed something.
------

For this PR, also did a test removing the SSHD CR from a gcp cluster, LB got deleted and I didn't see any obvious issue.

PR #923, opened 2021-09-08
OSD-7566 sre source builds

dofinn:
Locked egress evironments prevent the current sre-build-test from completing due to requiring pull data from github/registry.npmjs.org.

This build is performed locally form filesystem using `--binary` and passing the source into the builds rather then retrieving it remotely. To complete the build, a new image is now pushed to the internal registry of which extra garbage collection has been added to remove created images to prevent polluting the registry for customers.

### DIFF
--- a/deploy/sre-build-test/50-source-build-test.CronJob.yaml
+++ b/deploy/sre-build-test/50-source-build-test.CronJob.yaml
@@ -50,7 +50,8 @@ spec:
 
               # cleanup traps on exiting, so these always run
               cleanup () {
-                echo "$(date): Done, deleting build NS=$NS"
+                echo "$(date): Done, deleting build images and build NS=$NS"
+                oc -n "${NS}" get build -o jsonpath='{.items[0].status.output.to.imageDigest}' --ignore-not-found | xargs oc -n "${NS}" delete images
                 oc delete ns "${NS}"
               }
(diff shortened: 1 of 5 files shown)


deploy/sre-build-test/50-source-build-test.CronJob.yaml
```diff
@@ -81,7 +82,18 @@ spec:
...
+              package main
+              import (
+                     "fmt"
+              )
+
+              func main() {
+                      fmt.Println("Hello Openshift SRE :)")
+              }
+              EOF
+              oc -n "${NS}" start-build "${JOB_PREFIX}" --from-file=/tmp/main.go
```
NautiluX, line 96: wouldn't it be good enough to do `oc start-build -w -n ...` instead of the loop?

NautiluX (commented):
/lgtm
I like it, the comment is about code you didn't touch, but I thought it could be something we could adjust while we're working on it anyways.

deploy/sre-build-test/50-source-build-test.CronJob.yaml
dofinn: instead of the `while` loop? Then we would have to assure the `-w` never drops, in which case we would wrap `-w` with `while true; do ..... -w;done` ?

PR #924, opened 2021-09-09
OSD-8306, Added Reader permissions to the openshift-backplane-managed…

qhua948:
…-scripts namespace


qhua948:
This duplicates the https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-delete-backplane-script-resources/00-delete-backplane-script-resources.namespace.yml but i don't think there is any harm

PR #925, opened 2021-09-10
add missing odf addon for mt-sre backplane permissions

erdii:


cblecker:
/close

covered by #928

PR #931, opened 2021-09-14
Update apiVersion for the rebalance-infra-nodes CronJob to batch/v1beta1

blrm:
This is a small change to update the apiVersion for the infra node rebalance cronjob from `v1` to `v1beta1`.


cblecker:
/hold

This will cause this job to only function on 4.8

cblecker (approved):
/lgtm
/approve
/hold cancel

cblecker:
Actually, this change was going the opposite say to *fix* things on pre-4.8. My bad.

PR #934, opened 2021-09-15
[OSD-7738] Document all namespaces that are managed by Add-Ons

aliceh:
Added a script that finds all namespaces for managed-tenants addons and outputs them as a yaml document


resources/addons-namespaces/main.yaml
```diff
@@ -0,0 +1,21 @@
+ocs-converged-dev: openshift-storage
```
dustman9000, line 1: To valid yaml, there needs to be a root element:
```suggestion
---
addon-namespaces:
   ocs-converged-dev: openshift-storage
```

scripts/addons-namespaces/generate-add-ons-list.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_argument(
+        "--path",
+        "-p",
+        required=True,
+        help="The path to managed-tenants/addons/ directory [required]",
+    )
+    parser.add_argument(
+        "--output",
+        "-o",
+        help="The destination of the output yaml file. Default managed-cluster-config/resources/addons-namespaces/main.yaml ",
```
dustman9000, line 26: No need to indicate the default in the help message. Just add a default value here.

scripts/addons-namespaces/generate-add-ons-list.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    parser.add_argument(
+        "--path",
+        "-p",
+        required=True,
+        help="The path to managed-tenants/addons/ directory [required]",
+    )
+    parser.add_argument(
+        "--output",
+        "-o",
```
dustman9000, line 25:
```suggestion
        "-o",
        default="managed-cluster-config/resources/addons-namespaces/main.yaml",
```

scripts/addons-namespaces/generate-add-ons-list.py
```diff
@@ -0,0 +1,58 @@
...
+                namespace = ""
+                if file.endswith('addon.yaml'):
+                    file_to_search = open(file,"r")
+                    for line in file_to_search.readlines():
+                        if "targetNamespace" in line:
+                            ns = line.split(":")[1].strip()
+                            if namespace != ns:
+                                namespace = ns
+                                namescpace_dict[operator] = namespace 
+    if arguments.output == None :
```
dustman9000, line 51: No need for the check, just use the arguments.output value if you put the default value in the argument definition.

scripts/addons-namespaces/generate-add-ons-list.py
```diff
@@ -0,0 +1,58 @@
...
+        output="managed-cluster-config/resources/addons-namespaces/main.yaml"
+    else:
+        output=arguments.output
+    with open(output, 'w') as file:
+        yaml.dump(namescpace_dict, file )
```
dustman9000, line 56: Add extralines to separate functions.
```suggestion
        yaml.dump(namescpace_dict, file )
        
        
```

resources/addons-namespaces/main.yaml
aliceh: I fixed the script to generate a nested dict, so the yaml it generates has a root element.

PR #935, opened 2021-09-16
[OSD-6381]Remove openshift-sre-sshd ns from none hive clusters.

wanghaoran1988:
This remove openshift-sre-sshd from none hive clusters

### DIFF
--- a/deploy/osd-ingress/config.yaml
+++ b/deploy/osd-ingress/config.yaml
@@ -1,0 +1,3 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+    resourceApplyMode: Upsert
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -3324,6 +3324,12 @@ objects:
         - 'true'
     resourceApplyMode: Sync
     resources:
+    - apiVersion: v1
+      kind: Namespace
+      metadata:
+        name: openshift-sre-sshd
+        labels:
+          managed.openshift.io/service-lb-quota-exempt: 'true'
     - apiVersion: cloudingress.managed.openshift.io/v1alpha1
       kind: SSHD
       metadata:
@@ -6762,11 +6768,6 @@ objects:
       matchLabels:
         api.openshift.com/managed: 'true'
     resourceApplyMode: Sync
-    resources:
-    - apiVersion: v1
-      kind: Namespace
-      metadata:
-        name: openshift-sre-sshd
     patches:
     - apiVersion: operator.openshift.io/v1
       applyMode: AlwaysApply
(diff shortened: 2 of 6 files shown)


fahlmant (commented):
Can you clarify how the namespace will be limited to hives?

feichashao:
> Can you clarify how the namespace will be limited to hives?

It is moving to a directory scoping to hive only. 
-->  deploy/cloud-ingress-operator-configuration/sshd/00-openshift-sre-sshd.Namespace.yaml
https://github.com/openshift/managed-cluster-config/blob/master/deploy/cloud-ingress-operator-configuration/sshd/config.yaml

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -6762,11 +6768,6 @@ objects:
       matchLabels:
         api.openshift.com/managed: 'true'
     resourceApplyMode: Sync
-    resources:
-    - apiVersion: v1
-      kind: Namespace
-      metadata:
-        name: openshift-sre-sshd
```
cblecker, line 6769: Removing it from this section will delete the namespace

cblecker (changes requested):
/hold

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -3324,6 +3324,12 @@ objects:
         - 'true'
     resourceApplyMode: Sync
     resources:
+    - apiVersion: v1
+      kind: Namespace
+      metadata:
+        name: openshift-sre-sshd
```
wanghaoran1988, line 3330: @cblecker It's added here.

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
wanghaoran1988: I moved the ns file to a different location.

cblecker: That doesn't matter. As the SSS is in "Sync" mode, it will still delete it, even if it's re-added to another SSS. Each SSS is separate in that way.

deploy/osd-ingress/config.yaml
```diff
@@ -0,0 +1,3 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+    resourceApplyMode: Upsert
```
wanghaoran1988, line 3: @cblecker Use Upsert here, so the openshift-sre-ssh ns removed here will not delete the ns in OSD.

cblecker: We want this NS deleted on everything.. *except* hive clusters

wanghaoran1988:
New one created : https://github.com/openshift/managed-cluster-config/pull/943

PR #937, opened 2021-09-20
CEE should be allowed to delete MustGathers

sedroche:
This test indicates that CEE should be able to delete MustGathers - https://github.com/openshift/osde2e/blob/main/pkg/e2e/operators/mustgather.go#L53


sedroche:
cc @wanghaoran1988

PR #938, opened 2021-09-20
OSD-8028: Patch IngressController.spec.replicas to 3 for specific clusters

fahlmant:
As a solution of [OSD-8028](https://issues.redhat.com/browse/OSD-8028), this PR applies a replica count of 3 to cluster IDs specified by `ROUTER_REPLICA_CLUSTER_IDS`

### DIFF
--- a/deploy/cloud-ingress-operator-configuration/routerreplicas-OSD-8028/config.yaml
+++ b/deploy/cloud-ingress-operator-configuration/routerreplicas-OSD-8028/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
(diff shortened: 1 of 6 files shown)


deploy/cloud-ingress-operator-configuration/routerreplicas-OSD-8028/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: api.openshift.com/id
+      operator: In
+      values: ["${{ROUTER_REPLICA_CLUSTER_IDS}}"]
```
cblecker, line 6: If this is defined as an array, I don't think it should be wrapped in square brackets

PR #940, opened 2021-09-20
Specify namespace of the default ingress controller

fahlmant:
Turns out you also need to supply a namespace when working with Kubernetes objects. This PR specifies which namespace the ingresscontroller is in that should be patched.

It's definitely a Monday


cblecker (approved):
/lgtm
/approve

PR #926, opened 2021-09-13
bz1997062: Adds DaemonSet to clean up pids

chopraapooja:
https://bugzilla.redhat.com/show_bug.cgi?id=1997062

Solution: https://access.redhat.com/solutions/6304881

PR #941, opened 2021-09-20
workaround for BZ1997062 - deploy daemonset to clean /run fs

yithian:
push out the workaround from https://access.redhat.com/solutions/6304881 to all clusters to stop the shift primary from having to do it manually after an alert

### DIFF
--- a/deploy/bz1997062/config.yaml
+++ b/deploy/bz1997062/config.yaml
@@ -1,0 +1,3 @@
+deploymentMode: "Direct"
+direct:
+    environments: ["production", "stage", "integration"]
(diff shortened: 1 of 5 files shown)


deploy/bz1997062/config.yaml
```diff
@@ -0,0 +1,3 @@
+deploymentMode: "Direct"
+direct:
+    environments: ["production", "stage", "integration"]
```
cblecker, line 3: I don't think this is what you intend. You only want this to go to hive clusters?

cblecker:
I think this is a dupe of https://github.com/openshift/managed-cluster-config/pull/926

PR #943, opened 2021-09-22
remove openshift-sre-ssh ns on none hive

wanghaoran1988:
close [OSD-6381](https://issues.redhat.com/browse/OSD-6381)
1. This added a config.yaml under osd-ingress, and keeps use the same sss name for the openshift-sre-sshd namespace, so the ns on hive will not be deleted.
2. Moves the ingress patch to a different sss.


cblecker:
/lgtm
/hold

This should work. You can remove the hold when you're ready to test @wanghaoran1988

PR #951, opened 2021-10-04
OSD-8513 delete failed sre build projects

dofinn:
To remove stress on control plane due to project count. Lets delete all the failed jobs and then re-deploy the job once its validated.


cblecker:
/hold

No, we can't do this.

dofinn:
/

PR #952, opened 2021-10-05
Fix cleanup bugs in sre-build-test script

cblecker:
There was a number of small possible bugs in the way this script was working.

Changes:
- Changes clusterwide permissions to be only looking at namespaces
- Uses a SubjectPermission to add more broad permissions to the new namespaces created based on a pattern
- Sets a 15minute timeout on the job
- Create future namespaces with a `managed.openshift.io/sre-build-test` label so we can query/delete them all at once
- Move the cleanup trap until after the variables are set
- Remove the extra cleanup step of looking into the build, as when we delete the NS all objects (including the images) will get cleaned up
- Create a workaround delete command that we can remove afterwards as unfortunately we can't query previously created namespaces based on a label yet. That line can be removed after this has been promoted and run for a day.


deploy/sre-build-test/30-build-test-SubjectPermission.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: sre-build-test
+  namespace: openshift-rbac-permissions
+spec:
+  permissions:
+  - allowFirst: true
+    clusterRoleName: admin
+    namespacesAllowedRegex: "^openshift-build-test-.*$"
```
dofinn, line 10: nice!

deploy/sre-build-test/50-source-build-test.CronJob.yaml
```diff
@@ -119,8 +134,14 @@ spec:
 
                     if [[ -n "${JOBS_TO_DELETE}" ]] ; then
                       echo "$(date): Selected jobs for deletion: ${JOBS_TO_DELETE}"
-                      echo "${JOBS_TO_DELETE}" | xargs oc -n "${POD_NS}" delete job
+                      echo "${JOBS_TO_DELETE}" | xargs oc -n "${POD_NS}" delete job --ignore-not-found --wait=false
                     fi
+
+                    # clean up any old namespaces, other than the current one
+                    echo "$(date): Cleaning up old namespaces"
+                    oc delete ns -l "managed.openshift.io/sre-build-test,kubernetes.io/metadata.name!=${NS}" --ignore-not-found --wait=false
```
dofinn, line 142: In future we could use something like `oc get ns -l "managed.openshift.io/sre-build-test,kubernetes.io/metadata.name!=${NS}" --sort-by=.metadata.creationTimestamp` if we wanted to save the last X jobs? Not going to block merged based on this

PR #960, opened 2021-10-08
Enable write-request-body audit profile in prod

gsleeman:
/hold until [this PR](https://github.com/openshift/splunk-forwarder-operator/pull/119) is promoted


gsleeman:
Moving to https://github.com/openshift/splunk-forwarder-operator

PR #962, opened 2021-10-13
use average value to determine if the sync failed

bmeng:
fixes [OSD-7200](https://issues.redhat.com/browse/OSD-7200)

### DIFF
--- a/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
@@ -70,12 +70,12 @@ spec:
       annotations:
         summary: "Node drain failed in the given time period which is not caused by the PDB"
         description: "node drain takes too long and cannot be finished in the given time period during cluster upgrade"
-    - alert: UpgradeConfigSyncFailureOver1HrSRE
-      expr: upgradeoperator_upgradeconfig_synced > 0
-      for: 1h
+    - alert: UpgradeConfigSyncFailureOver2HrSRE
+      expr: avg_over_time(upgradeoperator_upgradeconfig_synced[2h]) == 1
+      for: 5m
       labels:
         severity: warning
         namespace: openshift-monitoring
       annotations:
-        summary: "UpgradeConfig has not successfully synced in 1 hour."
-        description: "This clusters UpgradeConfig has not been synced in 1 hour and may be out of date"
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -70,12 +70,12 @@ spec:
       annotations:
         summary: "Node drain failed in the given time period which is not caused by the PDB"
         description: "node drain takes too long and cannot be finished in the given time period during cluster upgrade"
-    - alert: UpgradeConfigSyncFailureOver1HrSRE
-      expr: upgradeoperator_upgradeconfig_synced > 0
+    - alert: UpgradeConfigSyncFailureOver2HrSRE
+      expr: avg_over_time(upgradeoperator_upgradeconfig_synced[2h]) == 1
```
mrbarge, line 74: This looks good, I like the change.
How does the `for` (line below) factor in to this when using `avg_over_time` though - am I right in thinking that this can be dropped to something like `5m` rather than `1h` (if it is even necessary at all..?)?

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
```diff
@@ -70,12 +70,12 @@ spec:
       annotations:
         summary: "Node drain failed in the given time period which is not caused by the PDB"
         description: "node drain takes too long and cannot be finished in the given time period during cluster upgrade"
-    - alert: UpgradeConfigSyncFailureOver1HrSRE
-      expr: upgradeoperator_upgradeconfig_synced > 0
+    - alert: UpgradeConfigSyncFailureOver2HrSRE
```
mrbarge, line 73: Just a note to please have an ops-sop commit prepared that renames our alert SOP when this goes live.

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
bmeng: good catch, yes, we do not need 1 hour more here, I will check it to 5 minutes to fire the alert

deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
bmeng: sure, I will make a PR for that

PR #969, opened 2021-10-28
chore(osd-rebalance-infra-nodes): bump speed

georgettica:
now it does it every hour

related to [OSD-8668](https://issues.redhat.com/browse/OSD-8668)


georgettica:
Close in regard to #973

PR #970, opened 2021-11-01
OSD-8555 allow CEE to read node-logs

feichashao:


feichashao:
/hold   not sure if `nodes/proxy` is too wide. Need to double check and maybe go with managed-scripts.
https://docs.openshift.com/container-platform/4.9/rest_api/node_apis/node-core-v1.html#apiv1nodesnameproxypath

feichashao:
close this and use managed-scripts instead.

PR #973, opened 2021-11-03
infra-rebalance every 15 minutes after osd-cluster-ready has succeeded

blrm:
This increases the frequency of the rebalance cronjob to every 15 minutes and adds a check to the beginning of the script to ensure that the `osd-cluster-ready` job has succeeded before it runs.


dustman9000:
Nice work, thanks!
/lgtm
/approve

PR #974, opened 2021-11-03
Remove unneeded variables from MCC config

cblecker:
The `IDENTITY_*` variables are no longer needed as the legacy OpenShift_SRE IdP has been removed.

The `ENABLE_MANAGEMENT_API_SERVER_INGRESS` is no longer needed as we only have v4 shards left, so this should be `true` for everyone.


jharrington22 (approved):
/lgtm

PR #977, opened 2021-11-05
Move openshift-sre-sshd roles and deployment that watches that NS to MCC

fahlmant:
https://github.com/openshift/cloud-ingress-operator/pull/211

`openshift-sre-sshd` namespace and roles should only be deployed to hives. The role, rolebinding and watch namespace list are copied from the cloud-ingress-operator deploy


fahlmant:
Closing as this will not work. OLM will reinstall the CSV which will revert the Deployment

PR #978, opened 2021-11-08
feat(osd-sts-ack): auto-add ack [OSD-8884]

georgettica:


deploy/osd-sts-ack/4.9/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major/minor
+    operator: In
+    values: ["4.8"]
+  - key: api.openshift.com/sts-4.9-upgradeable
```
cblecker, line 7: Are we changing this key every new version?

georgettica: No, as the migration is only 4.8->4.9 AFAICT

cblecker: No, there will potentially be STS changes with each minor version

georgettica: solved based on your suggestion

georgettica:
json patch is incorrect, will create a PR to fix this 
```
failed to apply patch 0: Invalid JSON Patch
```

PR #990, opened 2021-11-29
OSD-8656 explicitly schedule uwm components to workers

dofinn:
Currently the UWM config is not set enabling defaults to be deployed by the openshift-user-workload-monitoring/prometheus-operator. This is undesirable as the operator schedules itself to master nodes. 

```
[~ {production} (osd-v4prod-aws:default)]$ oc -n openshift-user-workload-monitoring get pods prometheus-operator-68fd48b89d-6tp7s -o json | jq '.spec.tolerations[]'
{
  "effect": "NoSchedule",
  "key": "node-role.kubernetes.io/master",
  "operator": "Exists"
}
{
  "effect": "NoExecute",
  "key": "node.kubernetes.io/not-ready",
  "operator": "Exists",
  "tolerationSeconds": 300
}
{
  "effect": "NoExecute",
  "key": "node.kubernetes.io/unreachable",
  "operator": "Exists",
  "tolerationSeconds": 300
}
{
  "effect": "NoSchedule",
  "key": "node.kubernetes.io/memory-pressure",
  "operator": "Exists"
}
```

Thus, this PR explicity schedules UWM components onto workers inclusive of the default prometheus configuration as seen below. 

```
[~ {production} (osd-v4prod-aws:default)]$ oc -n openshift-user-workload-monitoring get cm user-workload-monitoring-config -o yaml
apiVersion: v1
data:
  config.yaml: |
    prometheus:
      retention: 24h
      resources:
        requests:
          cpu: 200m
          memory: 2Gi
```

Open question: should the openshift-user-workload-monitoring/prometheus-operator itself be seen as an infra component? Its not obvious: https://docs.openshift.com/dedicated/osd_policy/policy-responsibility-matrix.html#policy-customer-responsibility_policy-responsibility-matrix

### DIFF
--- a/deploy/osd-user-workload-monitoring/cluster-monitoring-config.yaml
+++ b/deploy/osd-user-workload-monitoring/cluster-monitoring-config.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: user-workload-monitoring-config
+  namespace: openshift-user-workload-monitoring
+data:
+  config.yaml: |
+    prometheusOperator:
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+    prometheus:
+      retention: 24h
+      resources:
+        requests:
+          cpu: 200m
+          memory: 2Gi
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+    thanosRuler:
+      nodeSelector:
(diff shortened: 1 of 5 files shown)


cblecker:
We can't force down a version of this config map, as customers need the flexibility to configure things like remote write, etc.

/close

PR #996, opened 2021-12-02
OSD- 8643 : Revised expected upgrade control plane duration window for MUO 

MitaliBhalla:
In the context of https://issues.redhat.com/browse/OSD-8643 , SREs are often alerted unnecessarily for false positive upgrade control plane timeouts which makes their shifts noisier.
Thus, The expected upgrade control plane duration window for MUO has been revised from 90 to 120. 
This change comes from inferences and conclusions made through the values based on data found via CCX.


NautiluX (commented):
/lgtm

PR #1001, opened 2021-12-09
Add back deployment of the openshift-sre-sshd namespace to all clusters

fahlmant:
This is needed to have rbac work for cloud-ingress-operator


jharrington22 (approved):
/lgtm

PR #1002, opened 2021-12-09
Revert "Add back deployment of the openshift-sre-sshd namespace to all clusters"

fahlmant:
Reverts openshift/managed-cluster-config#1001


jharrington22 (approved):
/lgtm

PR #1009, opened 2021-12-13
Place definition of cloud-ingress-operator in MCC so deployment is owned

fahlmant:
tl;dr this PR ensures that MCC owns the cloud-ingress-operator clusterrole for future removal.

When cloud-ingress-operator is deployed, OLM deploys a clusterrole called `cloud-ingress-operator.v1.<CSV Version>`. For example `cloud-ingress-operator.v0.1.485-c329c1d-5679d97985`.  Previously, the cloud-ingress-operator deployment also included a definition of a clusterrole called `cloud-ingress-operator`, deployed by a syncset. As a result, both clusterroles were present with exactly the same permission set. The selectorsyncset of the `cloud-ingress-opreator`clusterrole has been removed, so on new clusters there will only be the OLM-deployed version with the CSV name appended. However, when the `cloud-ingress-operator` clusterrole was removed from the syncset, the sync mode was Upsert, so existing clusterroles on clusters were not removed. This PR serves as a way to regain control of the deployment of those clusterroles so they can removed via MCC in the future.

### DIFF
--- a/deploy/cloud-ingress-operator-configuration/clusterrole/10-cloud-ingress-operator-clusterrole.yaml
+++ b/deploy/cloud-ingress-operator-configuration/clusterrole/10-cloud-ingress-operator-clusterrole.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: ClusterRole
+metadata:
+  name: cloud-ingress-operator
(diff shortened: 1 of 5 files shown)


jharrington22 (approved):
/lgtm

PR #1011, opened 2021-12-14
 [ ODD - 9141 ] MUO's cluster-config role/binding as a workaround for CSV permissions issue

a7vicky:
https://issues.redhat.com/browse/OSD-9141
As a workaround, this selectorsyncset will create that role and rolebinding.

### DIFF
--- a/deploy/managed-upgrade-operator-rbac/110-monitoring_reader_role.yaml
+++ b/deploy/managed-upgrade-operator-rbac/110-monitoring_reader_role.yaml
@@ -1,0 +1,25 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: muo-monitoring-reader
+  namespace: openshift-monitoring
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps  
+  - serviceaccounts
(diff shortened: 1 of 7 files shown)


mrbarge:
/hold
To be tested in conjunction with https://github.com/openshift/managed-upgrade-operator/pull/294

mrbarge:
@a7vicky can you please:
- add an `OWNERS` file is defined in the base directory that matches that of `managed-upgrade-operator-config`.
- add a `README.md` that describes the reason for each role being added.

fahlmant:
cloud-ingress-operator also does this, but uses the olm-artifacts-template.yaml SelectorSyncSet: https://github.com/openshift/cloud-ingress-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L168-L481

This ensures the rbac for CIO is also still in its own repo. Could this be moved to the MUO equivalent? https://github.com/openshift/managed-upgrade-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml

a7vicky:
Thanks @fahlmant  
I have followed the same strategy to have role,rolebindings as part of MUO repo.

PR #1016, opened 2021-12-20
OSD-9131 OSD-8791 cluster proxy alerting

dofinn:
This PR enables alerting for the conditions of:

- proxy CA expiring in less then 30 days - warning
- proxy CA expiring in less then 15 days - critical
- proxy CA failed validation - warning

These will need to hit the pager and be service logged manually until ocm-agent is ready. 

SOPS will be provided later as not a priority against change freeze deadline. 

This PR should not be merged until https://github.com/openshift/osd-metrics-exporter/pull/72 is in production.

### DIFF
--- a/deploy/sre-prometheus/100-cluster-proxy.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-cluster-proxy.PrometheusRule.yaml
@@ -1,0 +1,36 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-cluster-proxy-alerts
+    role: alert-rules
+  name: sre-cluster-proxy-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+    - name: sre-cluster-proxy-alerts
+      rules:
+        - alert: ClusterProxyCAExpiringSRE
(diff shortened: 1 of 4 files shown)


dofinn:
/hold 

pending https://github.com/openshift/osd-metrics-exporter/pull/72

mrbarge:
I'm not personally sold on the idea that `ClusterProxyCAExpiringSRE` becomes a critical 15 days out because it's not a critical situation for SRE or the cluster, and the responsibility currently falls on the cluster owner to do their due diligence in keeping tabs on their cluster CA. 

I would be more inclinced to have a critical `ClusterProxyCAExpiredSRE` which occurs once the CA has actually expired, as that can have actual impact to the cluster's operational ability and health, and will likely be the harbinger of future issues.

dofinn:
@mrbarge thanks for feedback. I use the proposed conditions in light of how a customer will receive them via service log, so they may receive the warning and forget/do-nothing, but then they receive a critical and should act. We can also add expired if we like ? else i think the Critical sev covers that case.

mrbarge:
I have made some SOP changes here to cater for these PRs:
https://github.com/openshift/ops-sop/pull/1652
@dofinn if you could give that PR a quick look to ensure it lines up with what is being done here, that'd be great.

/lgtm

PR #1021, opened 2022-01-12
add remoteWrite to non-UWM

dofinn:
remoteWrite config is not apart of non-UWM config.


dofinn:
looks like remote Write is only available from 4.9 https://github.com/openshift/cluster-monitoring-operator/blob/master/CHANGELOG.md#49 - would need to understand implications of deployment remoteWrite to pre 4.9 cluster else filter for 4.9+

PR #1022, opened 2022-01-12
fix(osd-cluster-acks): change label

georgettica:
this makes it pass the current admin gate


georgettica:
/hold
in case the change will be done in the backend side

georgettica:
confirmed this is the correct format

PR #1025, opened 2022-01-19
set sync failure alert to high

bmeng:
To catch the sync issue between cluster and OCM earlier due to the cluster owner is unavailable.


mrbarge:
Can you also please ensure that either the SOP is renamed (https://github.com/openshift/ops-sop/blob/master/v4/alerts/UpgradeConfigSyncFailureOver2HrSRE.md) 

Other than that,
/lgtm

PR #1026, opened 2022-01-19
Don't deploy CIO/MVO service accounts to STS clusters

jharrington22:
For STS clusters the cloud-ingress-operator and managed-velero-operator are not deployed so we don't need to ensure the service accounts for those operators exist.

We can likely remove these sync sets that are ensuring SA's exist since OLM no longer creates owner references on SA objects (which caused them to be removed in the first place).

This PR will fix ClusterSync errors for STS clusters and is safe. Another PR removing these objects will come after. 

https://issues.redhat.com/browse/OSD-9681


jharrington22:
@cblecker reping - I'll follow this up with the removal after some testing.

PR #1027, opened 2022-01-19
OSD-9307 add file integrity operator manifests for deployment into FR clusters

tonytheleg:
- Addresses [OSD-9307](https://issues.redhat.com/browse/OSD-9307)
- Adds all manifests, CRDs, and CR needed to deploy the file integrity operator to fedramp clusters only
- Updates managed namespaces to include the file integrity operators namespace


tonytheleg:
Going a different route, backing this out for now

PR #1034, opened 2022-01-27
add configmap for sts policies

pvasant:
cc: @vkareh 

The idea is to have these policies available in a ConfigMap in the hive/provisioning cluster so that OCM can read from it and prompt the user to create those policies in their cloud
 
https://issues.redhat.com/browse/SDA-4890


vkareh:
There is a SelectorSyncSet that probably shouldn't be there.

I wonder if the MCC repo allows for a Makefile that compiles the final YAML file. This way we can pull the policies straight from the `resources/sts/` dir and be always in sync.

/cc @jharrington22

pvasant:
> There is a SelectorSyncSet that probably shouldn't be there.
> 
> I wonder if the MCC repo allows for a Makefile that compiles the final YAML file. This way we can pull the policies straight from the `resources/sts/` dir and be always in sync.
> 
> /cc @jharrington22

I did read about the make file in ReadMe but wanted a head start so we can discuss

wanghaoran1988:
Is there a jira tracking this? shouldn't the policy created during the installation phase, how we gonna consume this configmap?

fahlmant:
@pvasant @vkareh We already have these defined here https://github.com/openshift/managed-cluster-config/tree/master/resources/sts/4.9

fahlmant (changes requested):
@pvasant @vkareh We already have these defined here https://github.com/openshift/managed-cluster-config/tree/master/resources/sts/4.9

Why are they needing to be added to hive?

vkareh:
@fahlmant 
> Why are they needing to be added to hive?

OCM needs access to these so that we can create roles/policies on behalf of the user on their cloud account. Also, the ROSA CLI needs to have these served from the OCM API to guide the user through the process of creating them. The idea is to decouple these policies from binary releases of the CLI.

fahlmant:
Any reason not to just grab them from github though?

pvasant:
@fahlmant  we need it in two places both in ocm and ROSA so instead of copying the files in both places, we need a single source of truth.

fahlmant:
This is meant to be the source of truth https://github.com/openshift/managed-cluster-config/tree/master/resources/sts/4.9

pvasant:
The idea is 
1) We need to decouple the policy from rosa cli so we don't want to release rosa every time policy changes 
it is right now in   https://github.com/openshift/rosa/tree/master/templates/policies 

2) We also need to use these policies in OCM API to enable users to create the roles and policies on behalf of them

vkareh:
> This is meant to be the source of truth https://github.com/openshift/managed-cluster-config/tree/master/resources/sts/4.9

@pvasant - this is a good point. Maybe we should just use the GitHub API to fetch these files in OCM (which I think you suggested initially). I don't like this, to be honest, as then OCM is bound to the file structure of MCC, which is why I suggested instead pushing these files to a ConfigMap that OCM can read.

fahlmant:
Currently we don't have a makefile for anything in `resources`. Feel free to add one. As long as the config is drawing from there we can move forward; that way any changes to the source of truth files will automatically be reflected in the config

pvasant:
@cblecker @jharrington22  Can you please help with this one. Using GitHub API has its own limitation of rate-limiting etc and it is not scalable though we could restrict it to run every day. But if we can get this as a config map it will be really helpful. 
@fahlmant  can you give an example of the make file  cc: @vkareh

maorfr:
as discussed on the OSD sync call, this is more fitting for being deployed through app-interface.
more information in https://issues.redhat.com/browse/OSD-9896

maorfr:
internal discussion going on in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/33572

maorfr:
replaced by https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/33598

PR #1044, opened 2022-02-11
[OSD-9676] adding suricata-sa so the DS has the necessary permissions, ran make

rhdedgar:
Adding this back to the DS config. Somehow this change didn't make it in with the rest of the DS code during the original merge.


dustman9000:
Verified suricata-as already exists.
/unhold

PR #1045, opened 2022-02-14
Removes the DNSErrors05MinSRE alerts

iamkirkbater:
These alerts have proven to be noise at best, and auto-resolve on their
own within 5-10m of being fired. Removing these alerts will reduce the
number of high alerts that page by approximately 33% (data from Feb
5-13).


dofinn:
@iamkirkbater 
* are there cases where this alert has been validated at all ?
* Could we look to improve the conditions rather then completely remove it ?

iamkirkbater:
Closing in favor of #1050.

To answer your questions @dofinn - There has been a single case in recent memory where this alert had alerted us to something actually wrong, but I'm not even sure if we found something was actually impacted on the cluster outside of this check failing because `redhat.com` as a TLD might not have been the same as the required egress checks here: https://docs.openshift.com/rosa/rosa_getting_started/rosa-aws-prereqs.html#osd-aws-privatelink-firewall-prerequisites

PR #1055, opened 2022-02-17
Remove osd-token-refresher for FR clusters

dustman9000:
No need for token refresher since there's no observatorium


deploy/osd-token-refresher/config.yaml
```diff
@@ -0,0 +1,4 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet:
+  matchExpressions:
+    - { key: api.openshift.com/fedramp, operator: DoesNotExist }
```
cblecker, line 4: nit: can we split this out onto multiple lines so it's consistent with other config files

also, functionally, we should be checking for the true value rather than exists, because `api.openshift.com/fedramp: "false"` would actually this to not be pushed out either, which is incorrect/unintuitive

```suggestion
    - key: api.openshift.com/fedramp
      operator: NotIn
      value: 'true'
```

cblecker (changes requested):

cblecker (approved):
/lgtm
/approve

PR #1056, opened 2022-02-21
[ OSD-9965 ] Add OCMAgent CR to MCC

a7vicky:
This PR adds ocm-agent custom resource to MCC:
- Currently ocm-agent is proposed to be deployed in staging and integration environment
- OAO watches the OCMAgent CR and creates ocm-agent deployment


deploy/ocm-agent-custom-resource/10-ocm-agent-cr.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: ocmagent.managed.openshift.io/v1alpha1
+kind: OcmAgent
+metadata:
+  name: ocmagent
+  namespace: openshift-ocm-agent-operator
+spec:
+  ocmBaseUrl: ${OCM_BASE_URL}
+  replicas: 1
+  services:
+  - service_log
```
bmeng, line 10: to match the name of the service, I think it should be `service_logs`

deploy/ocm-agent-custom-resource/10-ocm-agent-cr.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: ocmagent.managed.openshift.io/v1alpha1
```
bmeng, line 1: the directory to place those files, should be better to rename to `ocm-agent-operator-resources`? just like the other managed service. I am not pretty sure about this, maybe discuss within team

mrbarge: +1 I prefer `ocm-agent-operator-resources`

deploy/ocm-agent-custom-resource/config.yaml
```diff
@@ -0,0 +1,7 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Upsert
```
mrbarge, line 3: I'd prefer this as `Sync` myself, that way if we ever needed to suddenly withdraw OA from all clusters we have that ability. Not sure that having it as `Upsert` is really giving us any advantages.

a7vicky: I am having same thought and modified it to Sync in my last commit

deploy/ocm-agent-custom-resource/10-ocm-agent-cr.yaml
a7vicky: Thank you for review. Updated the directory name to `ocm-agent-operator-resources`

deploy/ocm-agent-custom-resource/10-ocm-agent-cr.yaml
a7vicky: Thank you for the review. To align with OCM api `/api/service_logs` updated the list of spec.services to service_logs

PR #1060, opened 2022-02-22
Update managed-namespaces configmap

tnierman:
Updates `managed-namespaces` configMap by following instructions at https://github.com/openshift/managed-cluster-config/tree/master/deploy/osd-managed-resources#managed-resources-configmap

### DIFF
--- a/resources/managed/all-osd-resources.yaml
+++ b/resources/managed/all-osd-resources.yaml
@@ -50,12 +50,12 @@ Resources:
   - name: openshift-build-test
   - name: openshift-cloud-ingress-operator
   - name: openshift-codeready-workspaces
-  - name: openshift-compliance
   - name: openshift-custom-domains-operator
   - name: openshift-customer-monitoring
   - name: openshift-logging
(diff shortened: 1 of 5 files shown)


resources/managed/all-osd-resources.yaml
```diff
@@ -83,18 +83,12 @@ Resources:
     name: osd-delete-backplane-serviceaccounts
   - namespace: openshift-build-test
     name: sre-build-test
-  - namespace: openshift-cloud-ingress-operator
```
mrbarge, line 86: Is it expected that this and the other service accounts are being removed as managed resources in this PR? It looks like it is removing all the SAs that are listed/deployed via `deploy/osd-serviceaccounts`.

tnierman: Interesting, looks like rerunning the script fixed the missing serviceaccounts. Thanks for catching this!

NautiluX (commented):
/lgtm

cblecker:
/lgtm cancel

Needs a `make` to regenerate the templates

PR #1073, opened 2022-03-04
Add openshift-sre-token-scraper to fedramp hives

mjlshen:
Adds openshift-sre-token-scraper to fedramp hives
https://issues.redhat.com/browse/OSD-10149


dustman9000:
Nice! I've verified the hive label (ext-managed.openshift.io/hive-shard) exists for the managed hives in FR.
/lgtm

PR #1075, opened 2022-03-08
[OSD-9948] push oauth server sli to observatorium-mst

mrWinston:
This PR implements part of [OSD-9948](https://issues.redhat.com/browse/OSD-9948).

It makes sure, that oauth-server's availability is pushed to observatorium for validating our SLOs. A new recording rule was created to create a simpler alias for pushing only the desired metrics to observatorium.

If that is not desired, it's also possible to use a much more complicated relabel config to select only `apiserver_request_total{container="oauth-openshift"}` for pushing ( see linked ticket for more information ).

This PR must only be merged, after [gitlab/lifecycler#3](https://gitlab.cee.redhat.com/service/lifecycler/-/merge_requests/3) has also been merged.

@dofinn

### DIFF
--- a/deploy/sre-prometheus/100-oauth-server.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-oauth-server.PrometheusRule.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-oauth-server
+    role: recording-rules
+  name: sre-oauth-server
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-oauth-server
+    rules:
+      # aggregate the metrics under a new name to make sending these via remoteWrite easier
+    - record: oauth_server_health_checks
+      expr: apiserver_request_total{container="oauth-openshift"}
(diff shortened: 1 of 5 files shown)


deploy/sre-prometheus/100-oauth-server.PrometheusRule.yaml
```diff
@@ -0,0 +1,15 @@
...
+    role: recording-rules
+  name: sre-oauth-server
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-oauth-server
+    rules:
+      # aggregate the metrics under a new name to make sending these via remoteWrite easier
+    - record: oauth_server_health_checks
+      expr: apiserver_request_total{container="oauth-openshift"}
```
dofinn, line 15: We should consider dropping the pods,instance labels here rather in observatorium

```
sum without (pod, instance) (apiserver_request_total{container="oauth-openshift"})
```

We need to standardize this approach and review all metrics we are currently sending

mrWinston: Just summing them and dropping the pod and instance label will cause issues with continuity further down the line. If you take the `rate` of this sum, it will have weird jumps in it when pods are restarting: https://prometheus.io/docs/prometheus/latest/querying/functions/#rate
> Note that when combining rate() with an aggregation operator (e.g. sum()) or a function aggregating over time (any function ending in _over_time), always take a rate() first, then aggregate. Otherwise rate() cannot detect counter resets when your target restarts.


The only thing we could be doing, is sending the sum of the rate over to observatorium, but i'm unsure right now if that is sufficient for our use case

dofinn: good catch !

PR #1080, opened 2022-03-09
OSD-10249 allow token-refresher to use additional CA bundle

mrbarge:
It was identified during testing of SDE-1471 that the `token-refresher` will need to trust the cluster owner's additional CA if they are deploying the cluster with a cluster-wide proxy.

This PR updates the deployment of the `token-refresher` to include a ConfigMap that will have the system+user CA bundles injected into it and subsequently mounted into the host.

As the `token-refresher` image is alpine based, the usual approach of deploying it to `/etc/pki/ca-trust/extracted/pem` does not work. Instead I've opted to overwrite the image's existing `/etc/ssl/certs/ca-certificate.crt` with the contents of the system+user CA bundle.

This has been tested on both a cluster with no proxy or custom CA configured, as well as on a cluster with a proxy configured. When a cluster does not have a proxy or custom CA configured, the system CA trust store is mounted through.

### DIFF
--- a/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
+++ b/deploy/osd-token-refresher/01-token-refresher.Deployment.yaml
@@ -32,6 +32,14 @@ spec:
         - effect: NoSchedule
           key: node-role.kubernetes.io/infra
           operator: Exists
+      volumes:
+      - configMap:
+          defaultMode: 420
+          items:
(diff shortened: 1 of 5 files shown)


mrbarge:
/hold
whilst it gets eyes.

jharrington22 (approved):
/lgtm

PR #1084, opened 2022-03-14
MTSRE-467 | provide MTSRE the backplane access to credentialsrequests CR

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <ykukreja@redhat.com>

For the addons relying on CRO, we (MTSRE) would require access to reading the `CredentialsRequest` CR to debug problems in case of incidents associated with CRO's failures with provisioning RDS instances.

### DIFF
--- a/deploy/backplane/mtsre/managed-odh/10-mtsre-odh-admins.SubjectPermission.yml
+++ b/deploy/backplane/mtsre/managed-odh/10-mtsre-odh-admins.SubjectPermission.yml
@@ -8,5 +8,8 @@ spec:
   - allowFirst: true
     clusterRoleName: admin
     namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
+  - allowFirst: true
+    clusterRoleName: system:openshift:cloud-credential-operator:cluster-reader
+    namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
   subjectKind: Group
   subjectName: system:serviceaccounts:openshift-backplane-mtsre
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -3042,6 +3042,9 @@ objects:
         - allowFirst: true
           clusterRoleName: admin
           namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
+        - allowFirst: true
+          clusterRoleName: system:openshift:cloud-credential-operator:cluster-reader
+          namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
         subjectKind: Group
         subjectName: system:serviceaccounts:openshift-backplane-mtsre
 - apiVersion: hive.openshift.io/v1
--- a/hack/00-osd-managed-cluster-config-production.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-production.yaml.tmpl
@@ -3042,6 +3042,9 @@ objects:
         - allowFirst: true
           clusterRoleName: admin
           namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
+        - allowFirst: true
+          clusterRoleName: system:openshift:cloud-credential-operator:cluster-reader
+          namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
         subjectKind: Group
         subjectName: system:serviceaccounts:openshift-backplane-mtsre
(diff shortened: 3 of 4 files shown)


deploy/backplane/mtsre/10-mtsre-credentials-request-apis.ClusterRole.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: backplane-mtsre-credentials-request-apis-reader
+rules:
+  - apiGroups:
+    - "cloudcredential.openshift.io"
+    resources:
+      - "credentialsrequests"
```
fahlmant, line 9: Is there a specific `credentialsrequest` MTSRE needs access to?

yashvardhan-kukreja: Nope, there's no dedicated `credentialsrequest` resource we'd need access. We need to all-in-all access to any `credentialsrequest` resource which CRO might be reconciling depending on the addon in the situation. Though we just require read-level access as we only need visibility to `credentialsrequest` CR for debugging purposes, therefore, there's no reason to have any write-level access.

cblecker (changes requested):
A new ClusterRole isn't needed as one already exists: `system:openshift:cloud-credential-operator:cluster-reader`

What you need to do is create a per-namespace binding to that (the current PR is cluster-wide, which is too permissive).

For example, if you were adding this for ODH, you would do something like in [this](https://github.com/openshift/managed-cluster-config/blob/94fdb2351183880b43db26bd100c8b3797a3da08/deploy/backplane/mtsre/managed-odh/10-mtsre-odh-admins.SubjectPermission.yml#L7-L10) section:
```yaml
  - allowFirst: true
    clusterRoleName: "system:openshift:cloud-credential-operator:cluster-reader"
    namespacesAllowedRegex: (^rhods-notebooks$|^redhat-ods-applications$|^redhat-ods-monitoring$|^redhat-ods-operator$)
```

And then the rbac-permissions-operator would add a rolebinding for that clusterrole to each of the namespaces in the list

yashvardhan-kukreja:
> A new ClusterRole isn't needed as one already exists: system:openshift:cloud-credential-operator:cluster-reader

Fantastic, just checked this out. Let me get rid of the clusterrole and clusterrolebinding, and instead create the `SubjectPermission` CR. Thanks!

> the current PR is cluster-wide, which is too permissive

I get this point but I made the privileges to be cluster-wide-read so as to help with potentially other addons in the future depending upon CRO.

I suppose a fairly restrictive namespace bound would be `redhat-*`, what do you think? :)

deploy/backplane/mtsre/30-mtsre-credentials-request-readers.SubjectPermission.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: backplane-mtsre-credentials-request-readers
+  namespace: openshift-rbac-permissions
+spec:
+  permissions:
+  - allowFirst: true
+    clusterRoleName: system:openshift:cloud-credential-operator:cluster-reader
+    namespacesAllowedRegex: (^redhat-.*)
```
wanghaoran1988, line 10: We'd better list out the namespaces here and update it when needed.

yashvardhan-kukreja: As mentioned [here](https://github.com/openshift/managed-cluster-config/pull/1084#issuecomment-1067731927), I'd suggest to keep it `redhat-.*` only because access to the `CredentialsRequest` CR (at least read-only access) would be potentially required in other namespaces as well, considering that any other addon could depend on CRO in a different namespace.

wdyt :)

wanghaoran1988: There are some addons that maintained by other teams and namespaces prefixed with redhat-* as well, so we'd better list what NS that your team is managing.

yashvardhan-kukreja: Being the MT-SRE team, we're supposed to manage/SRE for all the addons out there.

So, any namespace where any addon could be installed i.e. `redhat-.*` comes under what we are supposed to deal with :)

yashvardhan-kukreja: Alright then, let me migrate this `SubjectPermission` under `mtsre/managed-odh` with namespaces as (^redhat-ods-applications$)

yashvardhan-kukreja: Hi @cblecker, do you mind checking the above comment out in case you missed it :)
Thanks!

cblecker: @yashvardhan-kukreja Adding specific permissions for new addons is already required, as outlined in the subdirectories of https://github.com/openshift/managed-cluster-config/tree/master/deploy/backplane/mtsre

On further thought, this likely should just get added to the specific subject permissions file for that addon.

yashvardhan-kukreja: Alright, @cblecker . Updated the PR with that :)

PR #1086, opened 2022-03-15
Add updated pid limit to fleet

cblecker:
fixes [OSD-10564](https://issues.redhat.com/browse/OSD-10564)


jewzaam:
Discussed with @cblecker.  Generally OK with this with 2 changes:  expand all the 4.8 versions (so 4.8 upgrades get this) and use Upsert apply mode (so it cannot be removed).  For any cluster that requires this without an upgrade SRE can apply the ContainerRuntimeConfig directly.

jharrington22 (approved):

PR #1087, opened 2022-03-17
adding kube job metrics for OSD-9951

reetika-vyas:
This for OSD 9951, Adding metrics kube job -* to  observatorium


dofinn:
@reetika-vyas there should be labels on this metric to isolate only the build jobs?

reetika-vyas:
> @reetika-vyas there should be labels on this metric to isolate only the build jobs?

I have added the Prometheus rule and recorded it to two expr to calculate count/sum

dofinn:
/lgtm
/approve

happy dashboarding :)

PR #1091, opened 2022-03-21
[OSD-10448] updated api URL to use livez

dofinn:
As per https://kubernetes.io/docs/reference/using-api/health-checks/ the /healthz endpoint is being depreciated in favour of the more explicit /livez and /readyz. 

With the consideration of SLOs being generated by RMO and this endpoint, the /livez endpoint appears to be more applicable as its used to catch container failures and triggers container restarts as per https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/


georgettica:
Should we also change our other operators to handle liveness/readiness with these urls?

PR #1092, opened 2022-03-21
ControlPlane Timeout Window upgrade from 120 to 130

MitaliBhalla:
The recent data from ccx shows that we are currently facing upgrade in the time bracket of 122 which fails our current window i.e 120 and triggers an alert to us. This PR is created to fulfill for the same.


bmeng:
@MitaliBhalla Could you provide from which data shows that the upgrade time of 122 minutes?

MitaliBhalla:
> @MitaliBhalla Could you provide from which data shows that the upgrade time of 122 minutes?

The data from ccx can be accessed via permissions granted. I had these permissions as I upgraded the window in past as well. I ll ping you with the data on slack privately since it is not considered good by ccx to make it public.

PR #1093, opened 2022-03-21
Add policies for 4.10

pvasant:
https://issues.redhat.com/browse/SDA-5683

We need to reconcile all the latest policy from ROSA cli to MCC repo so app-interface can read it from and feed to CS


pvasant:
@cblecker @jharrington22 @vkareh PTAL

fahlmant:
Please add a description

pvasant:
> Please add a description

Done. Can you please review

fahlmant (commented):
Looks good generally. Is there also a set for a PrivateLink installation? 
https://github.com/openshift/managed-cluster-config/tree/master/resources/privatelink

pvasant:
@fahlmant  yes i see we already have those permissions 

```
"ec2:DescribeVpcEndpointServiceConfigurations",
                "ec2:CreateVpcEndpointServiceConfiguration",
                "ec2:DeleteVpcEndpointServiceConfigurations",
                "ec2:ModifyVpcEndpointServicePermissions",
                "ec2:DescribeVpcEndpointServices"
```

fahlmant:
@pvasant We need a set of 4.10 permissions that are specific to PrivateLink.

pvasant:
> @pvasant We need a set of 4.10 permissions that are specific to PrivateLink.

What are those? I see only those six permissions in the link you shared. Can you please send me those extra permissions i can add it.

pvasant:
> @pvasant We need a set of 4.10 permissions that are specific to PrivateLink.

Ok i see what you are saying. Yeah but CS or ROSA or app-interface doesnt care about that folder. I can add this to the MR

fahlmant:
Because PrivateLink and Standard STS clusters require different permissions. For example, PrivateLink installs don't need permissions to create VPCs, Subnets, etc. The point of STS is least privilege, so anytime an STS cluster is installed via PrivateLink it should use those policies instead. The rosa cli should be using these for those cases

jharrington22 (approved):
/lgtm

PR #1095, opened 2022-03-22
[OSD-10326] Initial FedRAMP token vaulter configuration

mjlshen:
Stores the apiserver URL (ClusterRole to get/list ClusterDeployments) and token (Role to get/list secrets in the openshift-sre-token-scraper namespace) which will be needed when running `oc login`

I was able to use Pod dnsConfig/dnsPolicies to get DNS resolution to FR Vault

```yaml
dnsConfig:
   nameservers:
     - "10.125.32.103"
   searches:
     - "vault.fedramp.devshift.net"
dnsPolicy: None
```

Verified this works in integration hive and it's sharing the MCC "config.yaml" with the token scraper so it'll be targeted to run on all FR Hives

### DIFF
--- a/deploy/osd-fedramp-token-scraper/20-vaulter-cronjob.yaml
+++ b/deploy/osd-fedramp-token-scraper/20-vaulter-cronjob.yaml
@@ -1,0 +1,56 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: openshift-sre-token-vaulter
+  namespace: openshift-sre-token-scraper
+spec:
+  schedule: "*/30 * * * *"
+  concurrencyPolicy: Replace
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          automountServiceAccountToken: true
+          serviceAccountName: openshift-sre-token-vaulter
+          containers:
+            - name: vaulter
+              image: "quay.io/mshen/openshift-sre-token-vaulter@sha256:db7807beeb4ead02f7b1575899472efb5040d35c7c9f218b9b73948cda6134ec"
+              env:
+                - name: VAULT_APPROLE_ROLE_ID
+                  valueFrom:
+                    secretKeyRef:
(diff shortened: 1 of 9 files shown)


deploy/osd-fedramp-token-scraper/20-vaulter-cronjob.yaml
```diff
@@ -0,0 +1,55 @@
...
+                  readarray -t tokens < <(oc get secrets -n openshift-sre-token-scraper -l app=openshift-sre-token-scraper -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}');
+                  for token in "${tokens[@]}"; do
+                    readarray -t components < <(echo "${token}" | sed 's/-sre-/\n/g');
+                    CLUSTER="${components[0]}";
+                    SRE="${components[1]}";
+                    API="$(oc get cd -n ${components[0]} -ojson | jq -r '.items[0].status.apiURL')";
+                    SECRET="$(oc get secret -n openshift-sre-token-scraper ${token} -ojson | jq -r '.data.token' | base64 -d)";
+                    vault kv put "osd-fedramp/srep-sa-tokens/${CLUSTER}/${SRE}" api="${API}" token="${SECRET}";
+                  done;
+                  sleep 3600;
```
dustman9000, line 48: Would it be better to put a smaller sleep in the for loop to throttle vault api calls?

dustman9000:
Looks good, nice work! Just had a minor question on the use of the sleep command. Feel free to cancel the hold.
/lgtm
/hold

deploy/osd-fedramp-token-scraper/20-vaulter-cronjob.yaml
mjlshen: Agreed! Removed this big sleep and added a 3 second sleep inside the for loop

mjlshen:
Closing due to proactively respecting OSD-9896

PR #1098, opened 2022-03-25
OHSS-11084 Fix clusters impacted by BZ2067978

mrbarge:
This adds a temporary job to identify clusters that have a `node-resolver` daemonset with missing `ownerReference` indicating that it can be a risk factor to encounter [BZ2067978](https://bugzilla.redhat.com/show_bug.cgi?id=2067978).

A workaround is to delete the daemonset and let it be recreated. Deleting the `dns-operator` forces the recreation.

### DIFF
--- a/deploy/bz2067978/01-osd-delete-noderesolver-daemonset.Role.yaml
+++ b/deploy/bz2067978/01-osd-delete-noderesolver-daemonset.Role.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-delete-noderesolver-daemonset-bz2067978
+  namespace: openshift-dns
+rules:
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  verbs:
+  - get
+  - list
+  - delete
--- a/deploy/bz2067978/10-osd-delete-noderesolver-daemonsets.CronJob.yaml
+++ b/deploy/bz2067978/10-osd-delete-noderesolver-daemonsets.CronJob.yaml
@@ -1,0 +1,42 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: osd-delete-noderesolver-daemonset-bz2067978
+  namespace: openshift-dns
+spec:
+  failedJobsHistoryLimit: 3
+  successfulJobsHistoryLimit: 3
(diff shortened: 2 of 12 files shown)


deploy/bz2067978/01-osd-delete-noderesolver-daemonset.Role.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-delete-noderesolver-daemonset-bz2067978
```
cblecker, line 5: This role is named the same as the the other one.. should this be combined?

mrbarge: The distinguishing factor is that one role sits in `openshift-dns` and the other in `openshift-dns-operator` (as it has to do things in both)

cblecker: Oh I missed that. Thanks!

deploy/bz2067978/10-osd-delete-noderesolver-daemonsets.CronJob.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: osd-delete-noderesolver-daemonset-bz2067978
+  namespace: openshift-dns
+spec:
+  failedJobsHistoryLimit: 3
```
wking, line 8: Can we add a [`ttlSecondsAfterFinished`][1] too?  So that if we fail a few rounds of this but then the next few rounds pass, the failed runs will eventually be automatically pruned?  Because it seems unlikely that we want folks digging in to investigate and clear out any hiccup failures manually.

[1]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#jobspec-v1-batch

cblecker: This would require moving to the batch/v1 version of the API. If we are targeting 4.8+ (k8s 1.21) then we can probably do that safely.

mrbarge: Fair point! Now integrated.

PR #1104, opened 2022-03-29
[WIP][OSD-10254] deploy the managedNofitication to osd-v4* clusters

bmeng:
Deploy the managedNofitication and paired prometheus rules to our owned production clusters before push to all the productions.

To make sure the ocm-agent and ocm-agent-operator work well there.


mrbarge:
As-is, this would result in the `ManagedNotification` no longer being distributed anywhere, as `config.yaml` clauses need to match. So it would be looking for staging/integration clusters with those names.

I'd recommend creating a subdirectory under this one, maybe just named OSD-10250, which has its own `config.yaml` and ManagedNotification CR, where the config.yaml would:
- Distribute to the 'production' `api.openshift.com/environment`
- Use `api.openshift.com/id` as the way to target a select group of clusters (`name` also works, but safer and more reliable to target the ID, as names are not unique)

bmeng:
Thanks! @mrbarge I was wondering how would the multiple items in the match will work together. I think make a new file makes sense.

mrbarge:
This all looks fine to me, thanks - just one minor ask, could you please change the directory name from `OSD-10250` to `osd-10254-ocm-agent-prod-verification` - just to fit with existing conventions.

Then I think this is good to put on hold until SDB-2683 is complete.

bmeng:
@fahlmant thanks! Description added.

PR #1105, opened 2022-03-29
Removing MCC management of the openshift-sre-token-scraper namespace

mjlshen:
https://issues.redhat.com/browse/OSD-10326 after discussion with AppSRE, since this is targeting hive clusters it should go in app-interface.


fahlmant:
Please add a description

PR #1109, opened 2022-03-31
OSD-10019 Add metrics for router SLI

feichashao:
Adding a prometheus record `sre_router_default_available`.  This records if the number of running router pod >= 1. If there's a least 1 pod running, it will have value 1, otherwise the value is 0.

To calculate the router availability, we can use:
```
avg_over_time(sre_router_default_available[24h])
```


feichashao:
Hi @dofinn , could you help to review this?

feichashao:
Closing this one and will make a new one due to https://issues.redhat.com/browse/MON-2374

PR #1114, opened 2022-04-01
Update pending timer for etcdDivergentRevisionsSRE

cblecker:
Update timer for this alert for 10m to 120m, and remove runbook link so that we can create an ops-sop SOP for this.


jharrington22 (approved):
/lgtm

PR #1118, opened 2022-04-04
OSD-10734 Add pruning for duplicate OAO servicemonitor

mrbarge:
In [OSD-9942](https://issues.redhat.com/browse/OSD-9942) we added metrics to the `ocm-agent-operator`, but a consequence of this was that we needed to change the name of OAO's `service` and `servicemonitor` to not be `ocm-agent-operator-metrics`. This change, when promoted, will leave an orphaned service and servicemonitor with those name on Production clusters.

This MCC PR is a reinstatement of a similar pruning cronjob (commit 8a4fee6c), modified to clean up these orphaned resources and prevent them from causing `TargetDown` alerts.

It will be removed upon successful execution.

refs [OSD-10734](https://issues.redhat.com/browse/OSD-10734)

### DIFF
--- a/deploy/sre-pruning/105-pruning.rbac.ClusterRole.yaml
+++ b/deploy/sre-pruning/105-pruning.rbac.ClusterRole.yaml
@@ -40,3 +40,20 @@ rules:
   - builds
   verbs:
   - delete
+# service and servicemonitor pruning
+- apiGroups:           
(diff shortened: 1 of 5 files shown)


dofinn:
/lgtm

pending prometheus remote write investigations.

PR #1123, opened 2022-04-08
OSD-10920 add job to set proxy for token-refresher

mrbarge:
This adds:
- a new CronJob to `openshift-monitoring` named `osd-set-token-refresher-proxy`
- a supporting ClusterRole/Binding and Role/Binding named `osd-set-token-refresher-proxy`

The goal of the CronJob is to reflect the state of the cluster's cluster-wide-proxy in the `token-refresher` Deployment's environment variable configuration.

If the cluster is not set up with a cluster-wide proxy, the environment variables will be set with an empty value which has the same effect as not using a Proxy.

- I have verified that ClusterSync will not overwrite the environment variables in the Deployment after they are set on-cluster.
- I have verified that `token-refresher` respects the setting of `HTTP_PROXY` and `HTTPS_PROXY` by verifying that it reports errors communicating via a proxy when the proxy is offline.

Refs [OSD-10920](https://issues.redhat.com/browse/OSD-10920)

### DIFF
--- a/deploy/osd-set-token-refresher-proxy/00-osd-set-token-refresher-proxy.ServiceAccount.yaml
+++ b/deploy/osd-set-token-refresher-proxy/00-osd-set-token-refresher-proxy.ServiceAccount.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-set-token-refresher-proxy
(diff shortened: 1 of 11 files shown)


mrbarge:
/hold
being discussed in OSD-10920

mrbarge:
Going to attempt to handle this a different way - CronJob is not the way to be managing this.

/close

PR #1124, opened 2022-04-08
[OSD-9779] Add policies for Cloud Ingress Operator STS Deployment

mrWinston:
Jira: https://issues.redhat.com/browse/OSD-9779

This PR adds the updated aws policy documents for the cloud ingress operator. See [Here](https://github.com/openshift/cloud-ingress-operator/pull/173) for the related cio PR.

### DIFF
--- a/resources/sts/4.9/openshift_cloud_ingress_operator_cloud_credentials_policy.json
+++ b/resources/sts/4.9/openshift_cloud_ingress_operator_cloud_credentials_policy.json
@@ -1,0 +1,48 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+      {
+          "Effect": "Allow",
+          "Action": [
+              "elasticloadbalancing:AddTags",
+              "elasticloadbalancing:CreateListener",
+              "elasticloadbalancing:CreateLoadBalancer",
+              "elasticloadbalancing:DeleteLoadBalancer",
+              "elasticloadbalancing:DescribeLoadBalancers",
+              "elasticloadbalancing:DescribeTags",
+              "elasticloadbalancing:DescribeTargetGroups",
+              "ec2:DescribeAccountAttributes",
+              "ec2:DescribeAddresses",
+              "ec2:DescribeInternetGateways",
+              "ec2:DescribeSecurityGroups",
+              "ec2:DescribeSubnets",
+              "ec2:DescribeVpcs",
+              "ec2:DescribeVpcClassicLink",
+              "ec2:DescribeInstances",
+              "ec2:DescribeNetworkInterfaces",
+              "ec2:DescribeClassicLinkInstances",
+              "ec2:DescribeRouteTables",
+              "ec2:AuthorizeSecurityGroupEgress",
+              "ec2:AuthorizeSecurityGroupIngress",
+              "ec2:CreateSecurityGroup",
+              "ec2:DeleteSecurityGroup",
+              "ec2:DescribeInstanceAttribute",
+              "ec2:DescribeInstanceStatus",
+              "ec2:DescribeNetworkAcls",
+              "ec2:RevokeSecurityGroupEgress",
+              "ec2:RevokeSecurityGroupIngress",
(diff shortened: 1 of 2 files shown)


fahlmant:
Will look over this today

resources/sts/4.9/openshift_cloud_ingress_operator_cloud_credentials_policy.json
```diff
@@ -0,0 +1,48 @@
...
+              "ec2:DescribeVpcs",
+              "ec2:DescribeVpcClassicLink",
+              "ec2:DescribeInstances",
+              "ec2:DescribeNetworkInterfaces",
+              "ec2:DescribeClassicLinkInstances",
+              "ec2:DescribeRouteTables",
+              "ec2:AuthorizeSecurityGroupEgress",
+              "ec2:AuthorizeSecurityGroupIngress",
+              "ec2:CreateSecurityGroup",
+              "ec2:DeleteSecurityGroup",
```
jharrington22, line 28: @mrWinston are these actually needed by CIO?

jharrington22 (changes requested):

jharrington22:
/hold

@mrWinston could you ensure each of these actions are actually used by CIO?

resources/sts/4.9/openshift_cloud_ingress_operator_cloud_credentials_policy.json
mrWinston: Looking only at the code itself would give the following permissions:
```
{
  "Version": "2012-10-17",
  "Statement": [
      {
          "Effect": "Allow",
          "Action": [
              "elasticloadbalancing:AddTags",
              "elasticloadbalancing:CreateListener",
              "elasticloadbalancing:CreateLoadBalancer",
              "elasticloadbalancing:DeleteLoadBalancer",
              "elasticloadbalancing:DescribeLoadBalancers",
              "elasticloadbalancing:DescribeTags",
              "elasticloadbalancing:DescribeTargetGroups",
              "ec2:DescribeSubnets",
              "ec2:DescribeInstances",
              "ec2:DescribeRouteTables",
              "route53:ChangeResourceRecordSets",
              "route53:ListHostedZones",
              "route53:ListHostedZonesByName",
              "route53:ListResourceRecordSets",
          ],
          "Resource": "*"
      }
  ]
}
```

Which is vastly different to what the currently deployed CIO is allowed to do.
Either, the CIO has a lot more permissions than what's actually needed *or* i'm doing something wrong when determining the permissions.

Anyway, i'm currently checking the permissions against the non-sts version.

mrWinston: The CIO does actually work with only the permissions mentioned in the previous comment. changing the publishingstrategy from internal to external and back works fine with that limited set.

Still, I'm not entirely confident that i can accurately judge the blast radius of this change.

fahlmant: In non-STS mode, the cluster has Administrator Access in the AWS account, so each operator can technically get whatever permissions it wants. That doesn't mean it needs all those permissions.

fahlmant: Correct. My point is the Credentials Request currently used by CIO doesn't necessarily reflect the actual minimal permissions because that hasn't been a requirement before.

mrWinston: So, can i resolve this thread then? I've checked if the CIO works with the given permissions by modifying the credentials request. Is there anything else I should be doing?

fahlmant (approved):
/hold cancel
/approve

jharrington22 (approved):

jharrington22:
@mrWinston thanks for looking at this, these permissions look much better.

/lgtm

PR #1133, opened 2022-04-18
Pruning SRE job to warning severity [OSD-10995]

dofinn:
PruningCronjobErrorSRE currently pages SRE with severity of Critical. This alert does not pose immediate danger to data loss or cluster stability. Thus proposal is to reduce this to a warning alert.


mrbarge:
Seems like discussion on 'should this not be critical' has come up several times before with no outcome. Most recently OSD-9243 and OSD-7810. I would be in favour of us moving this to warning.

PR #1137, opened 2022-04-20
add customer bearer token allowed secrets in openshift-customer-monitoring

maorfr:
### What type of PR is this?
feature

There are cases where metrics endpoints need authentication through Bearer tokens. The way to configure a prometheus target with this ability is defining a ServiceMonitor with the BearerTokenSecret attribute pointing to a secret in the same namespace as the ServiceMonitor. Basically the ServiceMonitor needs a reference to a Secret name, and the key to read the bearer token from: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint

The list of Secrets dedicated-admins can create is limited, and this PR adds some "generic" names for customers to be able to manage for this purpose.

https://issues.redhat.com/browse/APPSRE-4907

This is an alternative to #1138

### DIFF
--- a/deploy/osd-customer-monitoring/05-prometheus-k8s-role.yaml
+++ b/deploy/osd-customer-monitoring/05-prometheus-k8s-role.yaml
@@ -17,5 +17,10 @@ rules:
   - alertmanager-instance
   - prometheus-auth-proxy
   - alertmanager-auth-proxy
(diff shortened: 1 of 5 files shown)


jewzaam:
/hold

Hold for discussion in JIRA.

PR #1138, opened 2022-04-20
enable dedicated-admins to manage secrets in openshift-customer-monitoring

maorfr:
### What type of PR is this?
feature

There are cases where metrics endpoints need authentication through Bearer tokens. The way to configure a prometheus target with this ability is defining a ServiceMonitor with the BearerTokenSecret attribute pointing to a secret in the same namespace as the ServiceMonitor. Basically the ServiceMonitor needs a reference to a Secret name, and the key to read the bearer token from: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint

The list of Secrets only limits cli interaction with the secret, but in reality -
a customer can run any pod they want in the namespace, and that pod can mount any secret it wants, which grants access for a customer to the secret content.

This PR removes the secret name limitation altogether, as it is limiting, but not securing.

https://issues.redhat.com/browse/APPSRE-4907

This is an alternative to #1137

### DIFF
--- a/deploy/osd-customer-monitoring/05-prometheus-k8s-role.yaml
+++ b/deploy/osd-customer-monitoring/05-prometheus-k8s-role.yaml
@@ -8,14 +8,5 @@ rules:
   - ""
   resources:
   - secrets
-  resourceNames:
-  - alertmanager-main
-  - alertmanager-main-proxy
-  - prometheus-additional-scrape-config
-  - prometheus-additional-scrape-config-2
-  - prometheus-additional-alertmanager-config
-  - alertmanager-instance
-  - prometheus-auth-proxy
-  - alertmanager-auth-proxy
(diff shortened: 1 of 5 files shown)


jewzaam:
/hold

Hold for discussion in JIRA.

jewzaam:
Reviewed the state of things and discussed in JIRA.  We don't have a requirement to restrict secret access anymore and this PR can be merged.  I'll approve and look for someone else to lgtm as a sanity check.  Long term we could grant full admin permissions to openshift-customer-monitoring to dedicated-admins, but no reason to block this PR on that.

/hold cancel

/approve

rogbas:
/lgtm
With the suggestion of moving the custom prometheus to your own (non-openshift) NS in the future.

PR #1141, opened 2022-04-21
[OSD-9964] fix alert expression

a7vicky:
### What type of PR is this?
_fix_

Currently the alert will be in firing state if the expression is true and returning any value. To fix this we need to set the expression value == 1 so that it will fire once the condition reached to value "1" over 4 hr time period.  

This will be fixing below PR:
 https://github.com/openshift/managed-cluster-config/pull/1139


mrbarge:
@a7vicky could you please update the description with a link to the Jira and a bit more information about what/why this change is needed? Thanks.

PR #1143, opened 2022-04-22
[WIP] Commenting out DS until disk issues corrected

robotmaxtron:
### What type of PR is this?
Cleanup

Removes DS for osd-suricata to avoid disk issues on hive clusters

Disk issue


dustman9000:
Instead of deleting the DS, can we just target int/stage environments and target just worker nodes? We need Suricata in FedRAMP.

PR #1146, opened 2022-04-26
OSD-1007 Use fixed version of the watcher images in ClamAV workloads

mjlshen:
### What type of PR is this?
Bug

It was found that the previous version of this image had a misnamed binary that was preventing it from starting correctly. A new image has been built to remediate that: https://quay.io/repository/app-sre/watcher?tab=tags&tag=latest

[OSD-10007](https://issues.redhat.com//browse/OSD-10007)

This change affects FedRAMP only


rhdedgar (approved):
+1

PR #1147, opened 2022-04-27
Add new alert for MNMO reconcile failures.

bergmannf:
The alert will trigger, when the change rate of errors during
reconciliation exceeds or equals 5 in 5 minutes as perper [OSD-9911](https://issues.redhat.com//browse/OSD-9911).

Merging should probably wait until is merged https://github.com/openshift/ops-sop/pull/1905, so the link to the SOP doesn't point into nothingness.

### What type of PR is this?

Feature

With the managed-node-metadata-operator making it into production, additional alerting is desired to catch errors early, when the operator fails.

Fixes [OSD-9911](https://issues.redhat.com//browse/OSD-9911)

### DIFF
--- a/deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
@@ -1,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-managed-node-metadata-operator-alerts
+    role: alert-rules
+  name: sre-managed-node-metadata-operator-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-managed-node-metadata-operator-alerts
+    rules:
+    - alert: MNMOTooManyReconcileErrors15MinSRE
+      # If the amount of reconciliation errors in the last 15 minutes is > 5 trigger an alert.
+      # Currently the window is 90 minutes, because the reconcile loop uses some backoff mechanism for retries and with a 90 minute window the alert will keep on firing.
+      expr: increase(controller_runtime_reconcile_total{controller="machineset_controller", service="openshift-managed-node-metadata-operator-metrics-service", result="error"}[90m]) >= 5
+      for: 15m
+      labels:
+        severity: critical
+        namespace: "{{ $labels.namespace }}"
+        maturity: "immature"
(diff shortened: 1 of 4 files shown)


bergmannf:
This alert makes no sense in an OSD cluster, as MNMO is not running there.

bergmannf:
@shibumi I actually also had to recheck with @iamkirkbater , but you are correct, so I will reopen the MR after all.

bergmannf:
I tested the new alert in a staging cluster and it worked - I had to put the `for` time very low though, because there is a short burst of errors in the reconcile loop of `MNMO`, that will even out relatively quickly:

MNMO Query when introducing errors

dofinn:
Whats the potential impact to the cluster if this alert fires? cc @openshift/sre-alert-sme

bergmannf:
When the alert fires it is likely that changes to `labels` and `taints` the customer wants to set on his MachineSets did not propagate, so the nodes should all keep running, but there might be problems with the scheduling and thus customer workloads if the `taints` or `labels` are used for that.

Maybe @iamkirkbater or @katherinelc321 can evaluate this better, as they have worked a lot more with the MNMO than me.

iamkirkbater:
@dofinn - What @bergmannf mentioned above is true.  The impact is mostly just that MNMO cannot push out the custom taints/labels that the customer specified leading to unpredictability in the customer's workload scheduling if they're relying on those labels/taints.

Generally we _shouldn't_ ever see this error fire as OCM should be validating the Labels and Taints when the customer attempts to apply them, before they're pushed to the cluster, however this gives us a failsafe to be able to find gaps in their validation and push those back upstream.

dofinn:
is there webhooks preventing customer from interacting with these objects outside of OCM? I would like to see the threshold increased to maybe 15 mins - 5 mins might trigger during node reboot or some ephemeral event? Also - can we please have the SOP written/merged prior to merging this alert as its of serverity critical.

bergmannf:
@dofinn I increased the alert - I think there is no webhook currently checking these configurations, but I opened a PR to at least check the validity of taints in ocm as well (https://github.com/openshift-online/ocm-cli/pull/384)

dofinn:
Great work @bergmannf  - last request would be to consider using and help developing maturity labels of alerts as defined [here](https://docs.google.com/document/d/1PnsZ0L_lzzgLMQX79OiHyV6DzmtLOaeECbWeXuz9a7A/edit#).

dofinn:
awesome, thanks @bergmannf cc @aweiteka :)

PR #1151, opened 2022-04-29
Add monitoring CRs permissions for mtsre backplane users.

sugarraysam:
### What type of PR is this?
feature

The MTSRE team needs to manage monitoring resources in the addon-operator namespace and in all addons. This permission is missing when logging in as the backplane mtsre user.

Also, it seems `cluster-admin` role does not provide access to CRD or other resources created from extending the core k8s API. 

Maybe another solution would be to create a new `cluster-admin-mtsre` clusterrole, which would extend the native `cluster-admin` clusterrole. 

_Fixes #_


cblecker:
@sugarraysam Is MTSRE making manual changes to `servicemonitors`, `monitoringstacks`, and `thanosqueriers`? Wouldn't those be made via the deployment process?

sugarraysam:
@cblecker It will make it easier to test changes in stage if we have write access. We have an issue right now where our generated `ServiceMonitor` object has bugs and issues and there was no easy way for us to modify and try different things.

sugarraysam:
@cblecker can we merge this? Did you see my updates in our thread: https://coreos.slack.com/archives/C01L83EMDUZ/p1652104621238909?thread_ts=1651601614.723179&cid=C01L83EMDUZ

sugarraysam:
Final arguments for MTSRE team to gain write access listed here (ref: https://coreos.slack.com/archives/C01L83EMDUZ/p1652104621238909?thread_ts=1651601614.723179&cid=C01L83EMDUZ):

1. Most addons are hosted on clusters created by tenants so it is not easy for MTSRE to get cluster-admin access
2. Many clusters in the OCM prod environment are effectively stage/test clusters (to get around OCM stage auto cluster-deletion)
3. I see MTSRE using a workflow where we scale down the addon-operator replicas to 0 (stopping reconciliation), and then manually editing any monitoring CR to debug/troubleshoot

PR #1152, opened 2022-05-01
feat(alertmanager): duplicate ES alerts

georgettica:
CURRENT_STATE: CI is passing and names are modified

as we want the alerts to show up only on a specific set of clusters, we need to duplicate them only where we want them

this PR will do that (currently WIP but will be better later)

### DIFF
--- a/deploy/sre-prometheus/extended-logging/config.yaml
+++ b/deploy/sre-prometheus/extended-logging/config.yaml
@@ -1,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  # if we ever remove this, do not remove the resources
+  resourceApplyMode: "Upsert"
+  matchExpressions:
+
+  - key: ext-managed.openshift.io/extended-logging-support
+    operator: In
+    values: ["true"]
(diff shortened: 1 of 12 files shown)


georgettica:
related to OSD-11273

deploy/sre-prometheus/extended-logging/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  # if we ever remove this, do not remove the resources
```
fahlmant, line 3: Why?

fahlmant (commented):
I'm not clear on why the alerts are in two places; resources and deploy. Resources isn't used for any automation, its just information, but if the alerts are already in deploy I don't see a reason for them to be in both places.

georgettica:
Some are pulled from a cluster, and others are what we want to deploy (I manipulate the list a bit) see the YQ commands

deploy/sre-prometheus/extended-logging/config.yaml
georgettica: Not sure off the top of my head.
Will TAL later and reply

georgettica:
Yeah, I saw the rebase as a complex task (wasn't sure which `make` I should use)
Feel free to take the files and recreate a PR

cblecker:
/close

Closing in favour of #1185

PR #1166, opened 2022-05-11
Make the alert a warning to not block upgrades.

bergmannf:
This should remain until the alert is mature enough.

### What type of PR is this?
Bug

Right now the alert is immature, but can block upgrades.

This is blocking a cluster upgrade, so it should be put to warning instead.

Fixes https://issues.redhat.com/browse/OSD-11437


NautiluX:
/hold cancel
we can add a link later. This is blocking cluster upgrades.

PR #1168, opened 2022-05-13
[OSD-11487] Scan tester addition

rhdedgar:
### What type of PR is this?
Feature, osd-scanning detecting this container starting up can singularity verify end-to-end scanning functionality.

Adds an OpenShift CronJob to the same environment of clusters that have osd-scanning enabled.

Tracked in [OSD-11487]

Tested locally with crc.

### DIFF
--- a/deploy/osd-scan-tester/00-osd-scan-tester-Namespace.yaml
+++ b/deploy/osd-scan-tester/00-osd-scan-tester-Namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: scan-tester 
--- a/deploy/osd-scan-tester/10-osd-scan-tester-cronjob.yaml
+++ b/deploy/osd-scan-tester/10-osd-scan-tester-cronjob.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: scan-tester
+  namespace: scan-tester
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "04 04 * * 4"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - preference:
+                  matchExpressions:
+                  - key: topology.kubernetes.io/region
+                    operator: In
+                    values:
+                    - us-gov-west-1
+                    - us-gov-east-1
+                weight: 1
+          serviceAccountName: scan-tester-sa
+          restartPolicy: Never
+          containers:
+          - name: scan-tester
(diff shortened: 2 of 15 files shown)


deploy/osd-scan-tester/config.yaml
```diff
@@ -0,0 +1,7 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet:
+  matchExpressions:
+    - key: api.openshift.com/fedramp
+      operator: In
+      values:
+        - "true"
```
dustman9000, line 7: This means it would install to all clusters. Similar to ids-tester, change to only non-production: https://github.com/openshift/managed-cluster-config/blob/9916ac4591ea4dcfe9c12287360d1c3f3bb2a44b/deploy/osd-ids-tester/config.yaml#L8-L11

deploy/osd-scan-tester/00-osd-scan-tester-Namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: scan-tester
```
dustman9000, line 4: Can we name it osd-scan-tester to be consistent?

deploy/osd-signature-updater/config.yaml
```diff
@@ -0,0 +1,12 @@
+# Placeholder until there's a permanent cluster to run this.
```
dustman9000, line 1: I thought the cluster to run this was going to be the appsre clusters. Is there a reason this is going into MCC now?

rhdedgar: This is a more general-purpose test, not for one particular cluster, but to ensure every cluster's scanning implementation is indeed scanning new pods and reporting its findings. I'll answer the slack specific questions on slack.

deploy/osd-scan-tester/config.yaml
afreiberger: +1 to make this change for non-production first.  Change control tracking will be required for production, and this could be the proof of testing before production release.

deploy/osd-scan-tester/10-osd-scan-tester-cronjob.yaml
```diff
@@ -0,0 +1,33 @@
...
+                    operator: In
+                    values:
+                    - us-gov-west-1
+                    - us-gov-east-1
+                weight: 1
+          serviceAccountName: scan-tester-sa
+          restartPolicy: Never
+          containers:
+          - name: scan-tester
+            image: quay.io/dedgar/scan-tester:v0.0.1
```
dustman9000, line 31: For change management purposes, this needs to be a RH quay repo, not a personal one.

dustman9000:
I don't think this repo is a great place for workloads that are intended for e2e testing. Can we move the ids-tester and scan-tester to https://github.com/openshift/osde2e instead?

rhdedgar:
Closing to re-locate to e2e.

PR #1173, opened 2022-05-25
Remove addons-namespaces configmap

cblecker:
### What type of PR is this?
cleanup

This configmap is not needed as we don't want to manage or care about which namespaces addons are using

_Fixes #_


blrm:
The changes to the script here look good and will remove the `addons-namespaces` configmap. I'll defer to @tnierman on how this may impact anything that is consuming that configmap. From discussing this in slack, it looks like all consumers have been updated to address this, but just want to call it out as a possible concern. 

/lgtm

PR #1179, opened 2022-06-02
FedRAMP cluster monitoring

theautoroboto:
### What type of PR is this?
FedRAMP alteration

This will provide a FedRAMP specific config for cluster monitoring

[OSD-11807](https://issues.redhat.com/browse/OSD-11807)

_Fixes #_
Conflicting configuration for FedRAMP clusters

### DIFF
--- a/deploy/cluster-monitoring-config/fedramp/config.yaml
+++ b/deploy/cluster-monitoring-config/fedramp/config.yaml
@@ -7,7 +7,3 @@ selectorSyncSet:
     - key: ext-managed.openshift.io/uwm-disabled
       operator: NotIn
       values: ["true"]
-    - key: api.openshift.com/fedramp
-      operator: NotIn
-      values:
-        - "true"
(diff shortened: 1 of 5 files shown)


deploy/cluster-monitoring-config/fedramp/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
```
tonytheleg, line 3: Think we're missing an expression here to only target FedRAMP clusters, since the other config still exists for Non-FR clusters

deploy/cluster-monitoring-config/fedramp/config.yaml
```diff
@@ -0,0 +1,9 @@
...
+    - key: hive.openshift.io/version-major-minor
+      operator: NotIn
+      values: ["4.5"]
+    - key: ext-managed.openshift.io/uwm-disabled
+      operator: NotIn
+      values: ["true"]
```
tonytheleg, line 9: Think we're missing an expression here to only target FedRAMP clusters, since the other config still exists for Non-FR clusters

theautoroboto:
@cblecker Is there anything I can do to move this forward?

cblecker:
Per discussions in slack, closing this as we are exploring other options with Suricata

/close

PR #1182, opened 2022-06-06
Update and sync monitoring configs

cblecker:
### What type of PR is this?
cleanup

Removes the limits from the Thanos querier as the bug referenced is fixed in all supported versions of OSD/ROSA.

Also syncs the openshiftStateMetrics scheduling config between the UWM and non-UWM configs

https://bugzilla.redhat.com/show_bug.cgi?id=1906496
#1181

### DIFF
--- a/deploy/cluster-monitoring-config-non-uwm/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config-non-uwm/cluster-monitoring-config.yaml
@@ -70,13 +70,10 @@ data:
         - effect: NoSchedule
           key: node-role.kubernetes.io/infra
           operator: Exists
-    # https://access.redhat.com/solutions/5685771
-    # https://bugzilla.redhat.com/show_bug.cgi?id=1906496
     thanosQuerier:
-      resources:
-        limits:
-          cpu: 500m
-          memory: 2048Mi
-        requests:
-          cpu: 5m
-          memory: 125Mi
+      nodeSelector:
+        node-role.kubernetes.io/infra: ""
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/infra
+          operator: Exists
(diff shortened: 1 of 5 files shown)


deploy/cluster-monitoring-config-non-uwm/cluster-monitoring-config.yaml
```diff
@@ -70,13 +70,17 @@ data:
...
+          key: node-role.kubernetes.io/infra
+          operator: Exists
     thanosQuerier:
-      resources:
-        limits:
-          cpu: 500m
-          memory: 2048Mi
-        requests:
-          cpu: 5m
-          memory: 125Mi
```
jewzaam, line 82: This should not be removed as this configuration is applied only to 4.5 per the `config.yaml`.  So, this applies only to impacted clusters.

deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
```diff
@@ -96,16 +96,7 @@ data:
...
-    # https://access.redhat.com/solutions/5685771
-    # https://bugzilla.redhat.com/show_bug.cgi?id=1906496
     thanosQuerier:
-      resources:
-        limits:
-          cpu: 500m
-          memory: 2048Mi
-        requests:
-          cpu: 5m
-          memory: 125Mi
```
jewzaam, line 108: This config will apply to 4.6 clusters, which do not have the fix.

jewzaam (changes requested):
/hold

deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
cblecker: 4.6 clusters are no longer supported as of Aug 2021: https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-life-cycle.html#rosa-life-cycle-dates_osd-life-cycle

jewzaam (approved):
Discussed version support internally.  +1 for the change.

/lgtm

PR #1184, opened 2022-06-07
Change InfraNodeNeedsResizing from warning to critical

mrbarge:
### What type of PR is this?
Feature

This changes the severity of the `InfraNodesNeedResizingSRE` alert from `warning` to `critical`.

The alert has been missed by SRE Primaries because Primary typically is not reviewing warning-level alerts.

There are three clusters actively firing this alert currently. 

An analysis of PagerDuty stats has shown this to only fire on five unique clusters from 2022 onwards, so this is not expected to create much more additional stress to SRE Primary if changed to critical.


jewzaam (approved):
Alert SOP needs tweaked to indicate it fires for critical now, it says it's warning.

/lgtm

PR #1192, opened 2022-06-16
[OSD-11496] Onboard elasticsearch alert to ocm agent

a7vicky:
### What this PR does / why we need it?
This PR include additional alert rule in existing prometheusrule and notificationtemplate in Managednotification CR for onboarding elasticsearch related alert to the OCM agent for servicelog automation.

### Which Jira/Github issue(s) this PR fixes?

_Fixes #_  [OSD-11496](https://issues.redhat.com/browse/OSD-11496)

### Pre-checks (if applicable):
- [x] Tested latest changes against a cluster
- [ ] Included documentation changes with PR

PR #1193, opened 2022-06-21
Deploy OAO to managednotifications and prometheusrules production

jwai7:
### What type of PR is this?
Feature

Adds OAO to production

_Fixes #_ https://issues.redhat.com/browse/OSD-10264


mrbarge:
I think it would be easier to just remove the `matchExpressions` section altogether.

mrbarge:
/hold
as I'd rather #1192 be merged first so that it can go out at the same time as this.

PR #1194, opened 2022-06-21
[SDE-1703] Tag STS policy files

feichashao:
### What type of PR is this?
Feature

- We want to limit the AWS actions to be only allowed on resources with a red-hat-managed tag.
- Not all services and actions supports the tag based authorization, so this PR tries to restrict as many as possible.

SDE-1703

- [x ] Tested latest changes against a cluster


resources/sts/4.10-tag-policies/sts-installer-preflight-checks.json
```diff
@@ -0,0 +1,94 @@
+{
```
jharrington22, line 1: @feichashao what was this file used for?

feichashao: This was used to workaround the pre-flight check SDA-6233.

PR #1195, opened 2022-06-21
Add PrometheusRule for BZ1980755

mrbarge:
### What type of PR is this?
Bug

Adds a new `warning` level alert `HighSubscriptionSyncRateSRE` intended to detect clusters that are impacted by bug [BZ1980755](https://bugzilla.redhat.com/show_bug.cgi?id=1980755) whereby operators will install but not be updateable due to  the CSV and Subscription not being correctly associated. This puts the operators in a state where OLM will no longer attempt to update them.

Analysis of Telemetry suggests that this will impact around 67 clusters. As the bug tends to occur at cluster install-time, that number will only grow. 

I'm not expecting SRE to actually do anything about this yet, so a SOP has not yet been written, though I would like to explore remediation via a managed script. 

The alert is checking SRE operators specifically because there is nothing SRE could/should do about customer-installed operators. 

We can, in the future, perhaps use a separate alert and OCM Agent Automation to inform the cluster owner that their operators are impacted by the bug.

This was first detected in OHSS-12724.

### DIFF
--- a/deploy/sre-prometheus/bz1980755/100-sre-bz1980755.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/bz1980755/100-sre-bz1980755.PrometheusRule.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  name: sre-bz1980755
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-bz1980755
+    rules:
+    - expr: subscription_sync_total * on(name) group_right label_replace(catalogsource_ready, "name", "$1", "name", "(.*)-(registry|catalog)")
+      record: sre:subscription_sync_total
+    - alert: HighSubscriptionSyncRateSRE
+      annotations:
+        message: "The sync rate for the {{ $labels.name }} operator Subscription is high and may indicate it has not installed successfully."
+      expr: |
+        rate(sre:subscription_sync_total{exported_namespace=~"openshift.*",exported_namespace!~"openshift-operators.*"}[30m]) * 100 > 10
+      for: 5m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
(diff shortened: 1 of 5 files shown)


deploy/sre-prometheus/bz1980755/100-sre-bz1980755.PrometheusRule.yaml
```diff
@@ -0,0 +1,21 @@
...
+    rules:
+    - alert: HighSubscriptionSyncRateSRE
+      annotations:
+        message: "The sync rate for the {{ $labels.name }} operator Subscription is high and may indicate it has not installed successfully."
+      expr: |
+        rate(subscription_sync_total{name=~"managed-upgrade-operator|cloud-ingress-operator|osd-metrics-exporter|must-gather-operator|managed-node-metadata-operator|custom-domains-operator|ocm-agent-operator|rbac-permissions-operator|route-monitor-operator|openshift-splunk-forwarder-operator|managed-velero-operator|configure-alertmanager-operator"}[30m]) * 100 > 10
+      for: 5m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
```
jewzaam, line 21: Join subscription_sync_total to catalogsource_ready on name, except strip "-registry" from the name label.  All SRE operators follow this pattern and therefore you can build a generic expression for the alert using exported_namespace.

```suggestion
    rules:
    - expr: subscription_sync_total * on(name) group_right label_replace(catalogsource_ready, "name", "$1", "name", "(.*)-registry")
      record: sre:subscription_sync_total
    - alert: HighSubscriptionSyncRateSRE
      annotations:
        message: "The sync rate for the {{ $labels.name }} operator Subscription is high and may indicate it has not installed successfully."
      expr: |
        rate(sre:subscription_sync_total{exported_namespace=~"openshift.*",exported_namespace!~"openshift-operators.*"}[30m]) * 100 > 10
      for: 5m
      labels:
        severity: warning
        namespace: openshift-monitoring
```

jewzaam (changes requested):

deploy/sre-prometheus/bz1980755/100-sre-bz1980755.PrometheusRule.yaml
mrbarge: Thanks, this is a great suggestion.

I sadly note that three SRE/Hive-distributed operators do not have a catalog that follows the pattern (addon-operator, managed-upgrade-operator, and splunk-forwarder-operator) - they instead end in `-catalog`. However, updating the regex to be `(.*)-(registry|catalog)` was successful in including them without inadvertently including any others, so that's what I've now done.

PR #1203, opened 2022-07-01
[OSD-12087] update the prometheusrule for UpgradeConfigSyncFailureOver4HrSRE

bmeng:
### What type of PR is this?
_feature_

Update the UpgradeConfigSyncFailureOver4HrSRE alert rule to use the new introduced metric in https://github.com/openshift/managed-upgrade-operator/pull/325

_Fixes #OSD-12087_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
@@ -64,7 +64,7 @@ spec:
       # Alert if MUO has been unable to successfully sync with its upgrade policy provider for a four-hour window
       # It will also be used for detecting the connection/authentication between cluster and OCM
       # Should be moved to OCM Agent service eventually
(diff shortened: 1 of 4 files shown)


mrbarge:
/lgtm

This looks good to me, but I've put a hold on it for now. Reason being that we need to make sure that MUO is promoted and running across the fleet at latest release before we promote this, otherwise we will cause an alert flood for any cluster that doesn't have MUO running at the latest version (that is complicated by BZ1980755 which is impacting a portion of the fleet)

Additionally because this alert can now fire if MUO is not running at all, we might want to update our SOP to note that the alert can fire if MUO is not running and SRE should now check that possibility (possibly as the very first step). 

My general suggestion would be:
- Promote MUO.
- Verify MUO has promoted across the fleet by verifying in Telemeter that the fleet is on the expected CSV
- Merge this to staging, verify in stage.
- Promote MCC.

PR #1208, opened 2022-07-07
Add CSSRE permissions to scale operator deployment

robshelly:
### What type of PR is this?
_feature_

Gives CSSRE permission to scale RHOAM deployment as the sometimes need to be scaled down to debug incidents.

https://issues.redhat.com/browse/CSSRE-1835

### DIFF
--- a/deploy/backplane/cssre/rhoam/01-rhoam-cssre-admins-project.ClusterRole.yaml
+++ b/deploy/backplane/cssre/rhoam/01-rhoam-cssre-admins-project.ClusterRole.yaml
@@ -123,3 +123,10 @@ rules:
   - '*'
   verbs:
   - '*'
+# CS SRE can scale deployments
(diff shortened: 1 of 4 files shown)


jewzaam:
/hold

This was modified in the srep permissions.
And the scope of what can be scaled is cluster wide.
The intent in CSSRE-1835 is to enable scaling of RHOAM related operators.

jewzaam (changes requested):
Scope needs to be for CSSRE and focus on RHOAM only per CSSRE-1835.

robshelly:
:man_facepalming: Fixed the mistake. Modified the correct file now. Thanks

jewzaam (changes requested):
Need to run `make` to generate updated templates.

jewzaam (approved):

PR #1210, opened 2022-07-11
[OSD-12236]expose the metric for ocm-agent to observatorium

bmeng:
### What type of PR is this?

Expose the metric for ocm-agent to observatorium 

_Fixes #OSD-12236_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -16,7 +16,7 @@ data:
           writeRelabelConfigs:
           - sourceLabels: [__name__]
             action: keep
-            regex: '(cluster_admin_enabled|identity_provider|probe_success|probe_duration_seconds|ingress_canary_route_reachable|oauth_server_requests_total|imageregistry_http_requests_total)'
+            regex: '(cluster_admin_enabled|identity_provider|probe_success|probe_duration_seconds|ingress_canary_route_reachable|oauth_server_requests_total|imageregistry_http_requests_total|ocm_agent_service_log_sent_total)'
           queueConfig:
             capacity: 2500
             maxShards: 1000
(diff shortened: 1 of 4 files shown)


bmeng:
> @bmeng is there a sample of this metric? has its cardinality potential been calculated? Is it for fleet or cluster or both insights?

Hi Dom,

We are reviewing the metric here, may update the metric to expose later, as the existing one might be reset with the container restarted.

The target metric would be a permanent one which will be used to count the total service log being sent based on the managedNotification we defined in the cluster.

Here is an example of the metric on my testing cluster:
```
sh-4.4# curl -s http://10.129.2.84:8383/metrics | grep ^ocm_agent_service_log_sent_total
ocm_agent_service_log_sent_total{ocm_service="service_logs",template="TestTemplate"} 14
```

I am not sure if I understand your last question. The metric will be existed on all the clusters across the fleet.

And the dashboard being shown in observatorium will be an accumulate count of total service log across the fleet which are sent by OCM Agent, for each template which defined within the `managedNofitication`

dofinn:
>We are reviewing the metric here
Was there suppose to be a hyperlink here?

Its not clear to me if this metric is suppose to be sent to MST or if another one is being worked on and it will replace it?

bmeng:
/hold

A new metric will be introduced in: https://github.com/openshift/ocm-agent/pull/44 
Please help review there if possible.

Eventually, there will be a `gauge` value metric to be exposed to observatorium, the name should be `ocm_agent_service_log_sent_total` in the PR above.

The result here is if we deployed the managedNotification on all of the clusters, and the OCM agent sent out the Service Log on some of them, the metric will be a `gauge` value on each of the clusters. And the value in observatorium will be a sum of the value for all the clusters.

bmeng:
@dofinn the metric here updated to the latest. PTAL

/unhold

dofinn:
/approve
/lgtm

Lets assess this in stage - cluster + RHOBS

PR #1213, opened 2022-07-13
OSD-12458 Update Cronjobs from batch/v1beta1 to batch/v1

mjlshen:
### What type of PR is this?
cleanup

cronjobs.v1beta1.batch has been deprecated starting in K8s v1.21 and removed in v1.25+ and has since been replaced with batch/v1

[OSD-12458](https://issues.redhat.com//browse/OSD-12458)

Left https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-pruning/images-pre-4.6/01-pruning-images.CronJob.yaml alone as it targets pre-4.6 versions - should we delete this perhaps?

### DIFF
--- a/deploy/osd-curated-operatorsources-revert/10-osd-patch-subscription-source.CronJob.yaml
+++ b/deploy/osd-curated-operatorsources-revert/10-osd-patch-subscription-source.CronJob.yaml
@@ -1,4 +1,4 @@
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: osd-patch-subscription-source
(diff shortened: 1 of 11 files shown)


mrbarge:
Would this not break the cronjobs on the portion of the fleet which is 4.7 and below? (of which there's still a fairly sizable number). AFAIK v1 is only usable in 4.8+.

mjlshen:
It would, is there a good way to check? I was going off of https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-life-cycle.html#rosa-life-cycle-dates_osd-life-cycle which shows 4.7 has been EOL'd on Dec 17, 2021

PR #1218, opened 2022-07-18
[OSD-12523] sre slo probe_success recordingrules

dofinn:
### What type of PR is this?
feature

### What this PR does / why we need it?
Combines the probe_success metrics from RMO with standardised telemetry to send via remoteWrite. Enables standardised drilldowns. 

### Which Jira/Github issue(s) this PR fixes?
https://issues.redhat.com/browse/OSD-12523 

_Fixes #_

### Pre-checks (if applicable):
- [x] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1219, opened 2022-07-18
remoteWrite sre slo probe_success

dofinn:
### What type of PR is this?
feature

New recordingRule has been created to capture the needed labels of the `probe_success` metric. This new rule should be sent to centralised obs via remoteWrite. Dashboards will use this rule. Having only the labels we need will reduce cardinality and improve query performance/reliability. 

_Fixes #_ https://issues.redhat.com/browse/OSD-12530 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


dofinn:
Rules were created here: https://github.com/openshift/managed-cluster-config/pull/1218

PR #1228, opened 2022-07-20
Set remote write config maxBackoff to default

PhilipGough:

The current 100ms value is hyper aggressive compared to the default value of 5s.
The 100ms has potential to exacerbate problems for a Prometheus remote write endpoint that is already under pressure.

This change relates to a [previous incident](https://issues.redhat.com/browse/APPSRE-4549) where the one of the action items from the [RCA](https://docs.google.com/document/d/1006VnOTyCg-FW-H0Y76LMEg0UXVMEI_7-eBMT-g2e3Q/edit#) was around remote write tuning.

We had a later [incident](https://issues.redhat.com/browse/MON-2553) where an action item identified via the RCA was to provide some guides for tenants configuration.

In our [onboarding documentation](https://docs.google.com/document/d/1Laphmtpb_n-YV20CPKe2MSikgHxuS14p8Iud2_CLso0/edit#heading=h.bupciudrwmna) for new tenants, we do not recommend deviating from the default settings for the remote write queue config. 

However the majority of the existing settings look here look unproblematic so I suggest we leave as-is. I do believe we should increase the max backoff setting to the default values of 5s however to effectively support backpressure in an overloaded system (both ingress and remote write).

Note that the Prometheus client will continue to double the wait from the minimum 30ms to max 5s with each failed attempt, so in scenarios where the system recovers quickly, this change would have little to no impact from the client perspective.

### DIFF
--- a/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -24,7 +24,7 @@ data:
             maxSamplesPerSend: 500
             batchSendDeadline: 60s
             minBackoff: 30ms
-            maxBackoff: 100ms
+            maxBackoff: 5s
       nodeSelector:
         node-role.kubernetes.io/infra: ""
       tolerations:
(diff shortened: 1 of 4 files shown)


cblecker:
/hold
/assign @dofinn 

Would like to get Dom's take on this change, but seems reasonable to me.

dofinn:
Thanks for the review @PhilipGough - changes look sane. 
/lgtm
/approve

PR #1240, opened 2022-08-01
Handle additional csv/sub combinations for BZ1980755 fix

mrbarge:
### What type of PR is this?
Bug

This makes a small correction to the BZ1980755 cronjob.

This fix caters for the splunk-forwarder-operator whose subscription name starts with `openshift-`, unlike the CSV.

[OSD-12547](https://issues.redhat.com//browse/OSD-12547)


NautiluX:
ignore my comment above, it won't work like that.
/lgtm

PR #1242, opened 2022-08-02
Node startup timeout

NautiluX:
### What type of PR is this?
bug

This change is to accommodate metal instance types and that the change is + 15 minutes since the [default](https://github.com/openshift/machine-api-operator/blob/128c5c90918c009172c6d24d5715888e0e1d59e4/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L47) is 10.

Fixes [OSD-12657](https://issues.redhat.com//browse/OSD-12657)

### DIFF
--- a/deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
+++ b/deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
@@ -21,3 +21,4 @@ spec:
     timeout: "480s"
     status: "Unknown"
   maxUnhealthy: 3
+  nodeStartupTimeout: 25m
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -9454,6 +9454,7 @@ objects:
           timeout: 480s
           status: Unknown
         maxUnhealthy: 3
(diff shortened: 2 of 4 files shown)


arendej:
correction:
AWS *.metal instances have longer startup times.

deploy/osd-machine-api/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
```diff
@@ -21,3 +21,4 @@ spec:
     timeout: "480s"
     status: "Unknown"
   maxUnhealthy: 3
+  nodeStartupTimeout: 25m
```
jharrington22, line 24: What is the default node start up time?

jharrington22: Never mind its 10 minutes. https://github.com/openshift/machine-api-operator/blob/128c5c90918c009172c6d24d5715888e0e1d59e4/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L47

jharrington22:
@NautiluX can you update the description to reference this change is to accommodate metal instance types and that the change is + 15 minutes since the [default](https://github.com/openshift/machine-api-operator/blob/128c5c90918c009172c6d24d5715888e0e1d59e4/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L47) is 10.

PR #1248, opened 2022-08-10
encrypt nodes' root volumes

yithian:
### What type of PR is this?
_cleanup/_

add a patch sss that will encrypt worker node root volumes per
https://issues.redhat.com/browse/OSD-3822

https://issues.redhat.com/browse/OSD-3822

this has been enabled by default starting in 4.5

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-machine-api/009-machine-api.srep-worker-rootvol-encryption.MachineSet.yaml
+++ b/deploy/osd-machine-api/009-machine-api.srep-worker-rootvol-encryption.MachineSet.yaml
@@ -1,0 +1,6 @@
+kind: MachineSet
+apiVersion: machine.openshift.io/v1beta1
+applyMode: AlwaysApply
+patch: |-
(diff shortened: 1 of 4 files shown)


cblecker:
/hold

@yithian which machineset are you patching? the patch doesn't specify

yithian:
[OSD-3822](https://issues.redhat.com/browse/OSD-3822) specifies the worker machinesets need to be patched but machineset names appear to be dynamically generated:

```
[~ {staging} (achvtest.xqmr.s1:openshift-machine-api)]$ oc get machineset
NAME                               DESIRED   CURRENT   READY   AVAILABLE   AGE
achvtest-tl9h2-infra-us-east-1a    2         2         2       2           4h19m
achvtest-tl9h2-worker-us-east-1a   4         4         4       4           4h43m
```

because of that, i was trying to just patch all machinesets to encrypt the root volumes. can a patch like this target machinesets based on wildcard names?

yithian:
/hold

after discussion with @NautiluX , i suspect this PR will be rejected and that we will take a different approach

PR #1250, opened 2022-08-11
[MTSRE-643] feat: adding nvidia gpu backplane config

Ajpantuso:
### What type of PR is this?
_feature_

Adds Backplane Role/Rolebinding for Nvidia GPU Addon Dev team to utilize admin permissions withing the addon namespace when installed.

https://issues.redhat.com/browse/MTSRE-643

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/nvidia-gpu/00-nvidia-gpu.namespace.yml
+++ b/deploy/backplane/nvidia-gpu/00-nvidia-gpu.namespace.yml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-backplane-nvidia-gpu
--- a/deploy/backplane/nvidia-gpu/config.yaml
+++ b/deploy/backplane/nvidia-gpu/config.yaml
@@ -1,0 +1,5 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchLabels:
+    api.openshift.com/addon-nvidia-gpu-addon: "true"
+  matchLabelsApplyMode: "OR"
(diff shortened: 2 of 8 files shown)


deploy/backplane/nvidia-gpu/20-nvidia-gpu-readers.ClusterRoleBinding.yml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: backplane-nvidia-gpu-readers
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: backplane-nvidia-gpu-readers-cluster
```
cblecker, line 8: Do you need view access to the whole cluster, including customer namespaces?

Ajpantuso: This role will only be used by the addon's engineers to support pilot customers. Outside of that context I would say no, but the expanded view access would be needed for hands-on customer support. Again that only applies since this role is not the same that would be used by SRE's for LA/GA support.

Ajpantuso: @cblecker did that explanation answer your question or are there still concerns about the scope of view access?

jewzaam: This sounds like better suited for privilege escalation which has a process behind it to ensure such access is justified.  Since there is not going to be a touch point to triage this access when transitioning out of pilot I would like to hear how escalation is insufficient.  

https://github.com/openshift/ops-sop/blob/master/v4/howto/backplane-elevate-privileges.md

jewzaam (changes requested):
/hold

PR #1251, opened 2022-08-11
Osd 12363 remove cbc ciphers

tonytheleg:
### What type of PR is this?
feature?

This PR is to remove weak ciphers from server/client SSH configs, specifically CBC ciphers.  
The change specifically:
* Adds a new file under /etc/ssh/ssh_config.d to specify the allowed ciphers instead of using the default
* Recreates the entire sshd_config file with a line added for the allowed ciphers
     * The reason for updating the entire sshd_config file is because RHCOS does not have an sshd_config.d directory to drop files in, nor the include directive in the sshd_config, so either method would require updating the sshd_config. The modified file is the default config with comments removed and just a new line at the end
     * MCO does not support appending the file so rewrite of file is required
   
These changes will only be applied to FedRAMP clusters. This was a finding in scanning our clusters for vulnerabilities and threats and has been asked to be remediated.

If you'd like to decode the content since they are URL encoded for Ignition, this KB article is super handy: https://access.redhat.com/solutions/4719071

https://issues.redhat.com/browse/OSD-12363

_Fixes #_

I'm not a huge fan of rewriting the entire sshd_config file but don't see another method of handling that with MCC/MCO since we can't do append. I can't imagine the default config changes very much between RHCOS versions so I might just be worrying more than I should but am open to suggestions. There is the potential to use crypto-policies and add a new module that removes the ciphers but that requries running a command to update it which I'm not sure we have something in place to do that on each node? 

- [x ] Tested latest changes against a cluster (ran this in my dev test cluster in FedRAMP)
- [] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-fedramp-machineconfig/99-fedramp-ssh-cipher-configs-master.yaml
+++ b/deploy/osd-fedramp-machineconfig/99-fedramp-ssh-cipher-configs-master.yaml
@@ -1,0 +1,21 @@
+---
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-master-ssh-cipher-config
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    storage:
+      files:
+        - contents:
+            source: data:,Ciphers%20aes256-gcm%40openssh.com%2Caes256-ctr%2Caes128-gcm%40openssh.com%2Caes128-ctr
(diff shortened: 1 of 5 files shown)


jaybeeunix:
these changes would be helpful in commercial too. for example, CPSYN-100 flagged these ciphers as a problem too.

PR #1253, opened 2022-08-15
OSD-12881 stop patching grafana route on 4.11

mrbarge:
### What type of PR is this?
Bug

ClusterSyncs are currently failing for 4.11 clusters because there's a Patch trying to apply to the `grafana` route, which is no longer a thing in 4.11.

Currently happening in two places
- osd-hsts-routes (all versions)
- osd-samesite-cookie (versions 4.6+)
- 
This PR makes that patch exclusive to clusters 4.10 and below.

[OSD-12881](https://issues.redhat.com//browse/OSD-12881)


cblecker:
/hold

I think this will break things. Need to have a closer look.

cblecker:
/hold cancel
Had another look and this should be fine. Thanks @mrbarge!

PR #1254, opened 2022-08-16
OSD-12547 Rework BZ1980755 script to be more efficient

mrbarge:
### What type of PR is this?
Feature

This CronJob attempts to address clusters that have operators impacted by bug [BZ1980755](https://bugzilla.redhat.com/show_bug.cgi?id=1980755).

**Please note this change only impacts staging environments**. Once it is proven to be working fine, a separate change will update the `config.yaml` to remove the environment restriction.

This is a rewrite of the CronJob script to address several issues found when it was last trialled in Production:
- Switch from `openshift/cli` to `openshift/tools` image so that `jq` is available.
- On 4.8 clusters and below, it took too long to execute as it made a lot of `oc` calls which got API-throttled. The number of `oc` calls is now greatly reduced. If a cluster is not impacted at all, only 1 call will be made. (made possible by switching to an image with `jq` in it)
- When the fix was applied to some clusters, OLM did not re-install the operator. The `catalog-operator` pod had gotten into a state requiring a restart. If the script replaces an impacted operator, it now bounces the `catalog-operator` pod before recreating the Subscription to workaround this issue.

[OSD-12547](https://issues.redhat.com//browse/OSD-12547)

### DIFF
--- a/deploy/bz1980755/10-bz1980755.CronJob.yaml
+++ b/deploy/bz1980755/10-bz1980755.CronJob.yaml
@@ -29,7 +29,7 @@ spec:
           restartPolicy: Never
           containers:
           - name: bz1980755
-            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            image: image-registry.openshift-image-registry.svc:5000/openshift/tools:latest
             imagePullPolicy: Always
             args:
             - /bin/bash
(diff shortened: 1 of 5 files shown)


deploy/bz1980755/10-bz1980755.CronJob.yaml
```diff
@@ -40,35 +40,46 @@ spec:
...
+                          oc delete clusterserviceversions.operators.coreos.com -n "${ns}" "${csv}"
+                      fi
                   done
-              done
+
+                  echo "Removing subscription ${sub}"
+                  oc delete subscriptions.operators.coreos.com  -n "${ns}" "${sub}"
+
+                  echo "Restarting catalog-operator to clear cache"
+                  oc delete pod -n openshift-operator-lifecycle-manager -l app=catalog-operator --wait=true
```
bmeng, line 80:
```
Restarting catalog-operator to clear cache
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:openshift-sre-pruning:bz1980755" cannot list resource "pods" in API group "" in the namespace "openshift-operator-lifecycle-manager"
```

I am seeing error when testing on my staging cluster. Seems the clusterrole is missing the permissions for the pods?

mrbarge: Nice, thanks for finding and spotting that! The `ClusterRole` did indeed require additional permissions. I've updated it to reflect that in the PR now.

bmeng:
tested and passed.

/lgtm

PR #1255, opened 2022-08-16
[Hypershift] Add operator policies for hypershift-specific operators

samira-barouti:
### What type of PR is this?
_feature_

This PR adds the operator policies for Hypershift-specific operators.

_Relates SDA-6223_

The policies are taken from the Hypershift repo: https://github.com/openshift/hypershift/blob/45676147134e22f876aaec3ea4239ef3c120fa69/api/v1alpha1/hostedcluster_types.go#L946

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.11/openshift_capa_controller_manager_credentials_policy.json
+++ b/resources/sts/4.11/openshift_capa_controller_manager_credentials_policy.json
@@ -1,0 +1,86 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Action": [
+          "ec2:AllocateAddress",
+          "ec2:AssociateRouteTable",
+          "ec2:AttachInternetGateway",
+          "ec2:AuthorizeSecurityGroupIngress",
+          "ec2:CreateInternetGateway",
+          "ec2:CreateNatGateway",
+          "ec2:CreateRoute",
+          "ec2:CreateRouteTable",
+          "ec2:CreateSecurityGroup",
+          "ec2:CreateSubnet",
+          "ec2:CreateTags",
+          "ec2:DeleteInternetGateway",
+          "ec2:DeleteNatGateway",
+          "ec2:DeleteRouteTable",
+          "ec2:DeleteSecurityGroup",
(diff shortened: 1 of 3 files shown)


samira-barouti:
@vkareh @pvasant PTAL

pvasant:
@cblecker These are needed for hypershift. What is the process to get this merged .

cblecker:
This adds the following additional permissions:
```
ec2:CreateLaunchTemplate
ec2:CreateLaunchTemplateVersion
ec2:DeleteLaunchTemplate
ec2:DeleteLaunchTemplateVersions
ec2:DescribeLaunchTemplateVersions
ec2:DescribeNetworkInterfaceAttribute
ec2:DisassociateAddress
```

@arendej @wgordon17 Any concerns with any of these? I don't really have any.

arendej:
I'm OK with these permissions.

cblecker:
@samira-barouti I tried to match all these up with the hypershift policies, and the ones I don't see is related to the [ingress](https://github.com/openshift/hypershift/blob/main/cmd/infra/aws/iam.go#L249) and [kms](https://github.com/openshift/hypershift/blob/main/cmd/infra/aws/iam.go#L307).

Are those included in some other policy?

samira-barouti:
@cblecker We are not using the OIDC feature of Hypershift; CS creates and populates the OIDC bucket itself, and provides the issuer URL to the Hypershift via HostedCluster CR. That's why we do not have those policies covered.
cc: @vkareh

wgordon17:
@samira-barouti, are these intended to be HyperShift-only policies? Meaning the OIDC permissions would be coming from a different policy (since we should still be creating the OIDC link in "auto" mode).

samira-barouti:
@wgordon17 that's right... these are Hypershift-only policies. Hypershift is still going to use the 6 operator policies that are used by Classic ROSA (which are defined within the same dir as these policies).

arendej:
FYI, it may become 5 operator policies, as we're currently looking at removing the CCO role and policy.

PR #1259, opened 2022-08-18
OSD-10920 Proxy-enable token-refresher

mrbarge:
### What type of PR is this?
Feature

- Creates a `token-refresher` ServiceAccount which the `token-refresher` Deployment now runs as.
- Adds an init container to the token-refresher `Deployment` which checks if the cluster has a cluster-wide proxy enabled, and if so, populates a new `token-refresher-proxy` `Secret` with the related values. Those values are then mounted from the `Secret` into the `token-refresher` pod as environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`).
- Adds RBAC for the `token-refresher` ServiceAccount to be able to do the above task.

If the cluster has no proxy enabled, the `Secret` is still created and mounted, but the proxy values will be empty.

This is a desirable change because it is a common occurrence for clusters using a cluster-wide-proxy to end up with a crashlooping `token-refresher` as it cannot successfully contact `sso.redhat.com`. We do have a [documented requirement](https://docs.openshift.com/rosa/networking/configuring-cluster-wide-proxy.html#cluster-wide-proxy-network-prereqs_cluster-wide-proxy-configuration) to exclude `sso.redhat.com` from proxy traffic, but this would eliminate the need for that requirement.

[OSD-10920](https://issues.redhat.com//browse/OSD-10920)

This was previously attempted in #1123 but that PR was rightly rejected because a `CronJob` is not the way to solve this. Hopefully this is more amenable.

- [X] Tested latest changes against a cluster (proxy-enabled and non-proxy-enabled)

### DIFF
--- a/deploy/osd-token-refresher/01-token-refresher.ClusterRole.yaml
+++ b/deploy/osd-token-refresher/01-token-refresher.ClusterRole.yaml
@@ -1,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: token-refresher
+rules:
+- apiGroups:
+    - config.openshift.io
+  resources:
+    - proxies
+  verbs:
+    - get
+    - list
+- apiGroups:
+    - ""               
+  resources:           
+    - secrets             
+  verbs:
+  - get
+  - update         
+  - create
+  - patch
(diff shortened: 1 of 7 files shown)


jharrington22:
@mrbarge nice, however the pod wont react to changes in the clusters proxy configuration correct?

jharrington22 (commented):
@mrbarge nice, however the pod wont react to changes in the clusters proxy configuration correct?

mrbarge:
@jharrington22 that's correct. However, a change of proxy configuration also results in a rolling reboot of all nodes, so, the pod should eventually restart as a result of that process and end up with the right configuration.

PR #1262, opened 2022-08-26
Allow OCM to manage policy versions

vkareh:
### What type of PR is this?

_bug_

It adds two new permissions to the OCM admin policy, which allows OCM to manage policy permissions to ensure OCM doesn't get a `LimitExceeded` from the AWS IAM API.

Fixes https://issues.redhat.com/browse/SDA-6659

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


vkareh:
@pvasant @tbrisker - please take a look

/cc @cblecker @jharrington22

cblecker:
@vkareh You're only adding this to the 4.10 policy. Should this be added to others?

vkareh:
@cblecker I'm adding it to both 4.10 and 4.11 in this PR. Older versions are not supported for admin mode in OCM.

cblecker:
Ah okay, I can see that.

/lgtm
/approve

PR #1270, opened 2022-09-12
Only look at the last sequential build status

iamkirkbater:
### What type of PR is this?
_bug_

In certain cases, the build test creation can encounter issues.  The build is Cancelled.  Multiple retries to create the build are done.

The script that checks the status of the build, though, does not take this into account and you'll get _all_ the build statuses as output, and not just the last one, which leads to strings like the following being compared in the switch statement:

```
Cancelled
Cancelled
Cancelled
Complete
```

Where the first 3 builds in the example above are cancelled before they're started, and we retry creation afterwards.  The last build is successful, but the entire string including newlines is being compared, which ends up failing to compare.

_Fixes OSD-13202_

- [ ] Tested latest changes against a cluster (tested the subcommand with the pipe in a cluster)

### DIFF
--- a/deploy/sre-build-test/50-source-build-test.CronJob.yaml
+++ b/deploy/sre-build-test/50-source-build-test.CronJob.yaml
@@ -103,7 +103,7 @@ spec:
               echo "$(date): Waiting for build to complete."
               while :
               do
-                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers | tail -1)
+                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --sort-by=.status.startTimestamp --no-headers | tail -1)
                 case ${ST} in
                   "")
                     # if build status is blank, assume we are still starting the build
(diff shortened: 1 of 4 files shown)


deploy/sre-build-test/50-source-build-test.CronJob.yaml
```diff
@@ -103,7 +103,7 @@ spec:
               echo "$(date): Waiting for build to complete."
               while :
               do
-                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers)
+                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --no-headers | tail -1)
```
cblecker, line 106:
```suggestion
                ST=$(oc -n "${NS}" get build -o custom-columns=STATUS:.status.phase --sort-by=.status.startTimestamp --no-headers | tail -1)
```

Great catch! Can we add in an explicit sort by the starting timestamp so that we are sure the last one listed is the most recent?

cblecker:
/lgtm

@iamkirkbater Feel free to remove the hold when testing is done

iamkirkbater:
Tested and validated with your suggested changes against the particular cluster where that was failing.

/hold cancel

Thanks all!

PR #1277, opened 2022-09-15
OSD-13103 Adjust alert to >8GB consumption for SDN pods to tune false positives

mjlshen:
### What type of PR is this?
bug

The alert is currently misleading and leading to a lot of false positives. As it has a very specific focus at the moment, tuning the SDN pod memory threshold to 8GB for now.

[OSD-13103](https://issues.redhat.com//browse/OSD-13103)


iamkirkbater:
/lgtm

Thanks Michael!

PR #1278, opened 2022-09-16
deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak: CronJob pruner

wking:
Similar to the pruner in openshift/release#32338, as linked from [OCPBUGS-1341][2].

[2]: https://issues.redhat.com/browse/OCPBUGS-1341

### What type of PR is this?

bug

The pruner removes excess resources leaked via node churn, until the network operator can be patched to stop leaking.  Leaked resources are more likely in managed clusters because of their node churn.  Managed-cluster use of MachineHealthChecks and scaling compute up around OCP-core updates both contribute to node churn across the managed fleet, as do customer decisions to use autoscalers in a subset of the managed fleet.  Significant numbers (tens thousands) of leaked resources lead to excessive memory use, and if left unpruned, will eventually require expensive control-plane scale-ups.

Fixes [OCPBUGS-1341][2].

I'm not SD, and this is my first managed-cluster-config pull request.  I'm trying to pattern-match prior commits, but :shrug: don't assume I know what I'm doing ;).

- [ ] Tested latest changes against a cluster (sort of.  I've tested openshift/release#32338).
- [ ] Included documentation changes with PR (documentation?)r
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, which I am, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak/config.yaml
+++ b/deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak/config.yaml
@@ -1,0 +1,10 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+  - key: api.openshift.com/environment
+    operator: In
+    values: ["staging", "integration"]
+  - key: hive.openshift.io/version-major-minor
+    operator: In
+    values: ["4.8","4.9","4.10","4.11"]
--- a/deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak/00-OCPBUGS-1341.ServiceAccount.yaml
+++ b/deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak/00-OCPBUGS-1341.ServiceAccount.yaml
@@ -1,0 +1,8 @@
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: sre-pod-network-connectivity-check-pruner
+  namespace: openshift-network-diagnostics
+  annotations:
+    kubernetes.io/description: Mitigate https://issues.redhat.com/browse/OCPBUGS-1341 by pruning leaked PodNetworkConnectivityChecks.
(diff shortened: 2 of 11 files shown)


mrbarge:
Thanks for this PR. I have tested it manually on a cluster and it seems to be behaving fine.

A couple of notes/questions:
- This should be restricted by version because `openshift-network-diagnostics` does not exist on `4.6` and `batch/v1` isn't available until 4.8. Should it be limited to 4.8 through 4.12? If so, please update the `config.yaml` to apply that restriction similar in execution to [this example](https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-samesite-cookie/pre-4.11/config.yaml).
- I am still a little unfamiliar with the underlying issue and consequences of the proposed fix to know whether executing this on the fleet is sound or not. Are there the consequences of removing the `PodNetworkConnectivityChecks`? When executing the job, there always seems to be at least one which is removed on each execution even though there's been no node churn on the cluster (example, `network-check-source-ip-10-0-211-116-to-network-check-target-service-cluster`; the `service-cluster` suffix seems to be common each time it runs). This then gets immediately recreated. Would that be expected?

wking:
> Should it be limited to 4.8 through 4.12?

Done with 29220fe -> b15ff46.

> Are there the consequences of removing the PodNetworkConnectivityChecks?

I don't see how PodNetworkConnectivityChecks between nodes that no longer exist could be useful, but I'm not a maintainer for that component, and the component maintainers have yet to weigh in in [OCPBUGS-1341|https://issues.redhat.com/browse/OCPBUGS-1341].  We have yet to notice negative consequences as a result of deploying openshift/release#32338, and that limits risk.

> ...the service-cluster suffix seems to be common each time it runs...

I've pushed b15ff46 -> 6c4c7a0, adding a `grep -v service-cluster` to avoid this churn.

deploy/OCPBUGS-1341-PodNetworkConnectivityCheck-leak/config.yaml
```diff
@@ -0,0 +1,13 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+  - key: api.openshift.com/environment
+    operator: In
+    values: ["staging", "integration"]
```
mrbarge, line 7: This block can probably be removed, we don't need to isolate it to staging/int environments.

wking: Chatting out of band, we're ok keeping this limit while we soak in stage/int before promoting to prod.

But we do want to drop the FedRAMP filter, to allow this to run there too, so I've pushed that with 6c4c7a0 -> c5bd4c0.

PR #1280, opened 2022-09-22
deploy/ocpbugs-1341-pod-network-connectivity-check-leak: Promote to production

wking:
Follow up on e275e5d01c (#1278) and c42a7c78df (#1279) with a promotion to production, because it seems to be working well in stage, based on Telemetry:

```
  max by (_id, url) (
    cluster:usage:resources:sum{_id=~"...REDACTED_IDS...",resource="podnetworkconnectivitychecks.controlplane.operator.openshift.io"}
    + on (_id) group_left (url)
    topk by (_id) (1, console_url)
  )
```


mrbarge:
/hold
Looks like there's an issue with this on staging that will need to be addressed before this goes any further. I'll update here with more details in a later comment.

mrbarge:
This is currently failing on clusters 4.10 and below:

```
13m         Warning   FailedCreate                 job/sre-pod-network-connectivity-check-pruner-27735841   Error creating: pods "sre-pod-network-connectivity-check-pruner-27735841-2x4qn" is forbidden: unable to validate against any
 security context constraint: [pod.metadata.annotations.container.seccomp.security.alpha.kubernetes.io/pod-network-connectivity-check-pruner: Forbidden: seccomp may not be set provider "anyuid": Forbidden: not usable by user or serv
iceaccount provider "nonroot": Forbidden: not usable by user or serviceaccount provider "pcap-dedicated-admins": Forbidden: not usable by user or serviceaccount provider "hostmount-anyuid": ......
```

To get it to run on clusters 4.8 through 4.11 is going to require removing the setting of `seccompProfile` in the container security context.

I tried that out and it seemed to run fine on a test 4.8 and 4.11 cluster.

wking:
> This is currently failing on clusters 4.10 and below...

I've opened #1284 to drop `seccompProfile`.  That might trigger `PodSecurityViolation` on 4.11; let me know if you think it's worth sharding into two chunks; one for 4.8 through 4.10 (without `seccompProfile`) and another for 4.11 (with `seccompProfile`).

mrbarge:
Re `seccompProfile` in 4.11 - that would be good, if it's not a problem. Seems like 4.11 clusters are already throwing `PodSecurityViolation` but it would be good to not exacerbate that.

To implement it requires that the 4.11-specific be placed in a different dir (ie subdir) with a different `config.yaml`.
You could possibly keep all the version-universal resources in the outer dir with the existing `config.yaml` and have specific 4.8/9/10 and 4.11 subdirs that contain the cronjob, I think that'd work.

This seems to be running fine on staging today after the fix.

wking:
> Seems like 4.11 clusters are already throwing `PodSecurityViolation` but it would be good to not exacerbate that.

I've opened #1285 with the `seccompProfile` sharding, using your suggested subdir approach.

mrbarge:
This seems good to go everywhere now based on my observations in staging. Thanks!

/lgtm
/unhold

PR #1283, opened 2022-09-24
Add missing RBAC for sre-build-test to get serviceaccounts

mjlshen:
### What type of PR is this?
cleanup

* The job sometimes times out because it spins on trying to get a serviceaccount it doesn't have the necessary RBAC to get.
* ~~The `--docker-image` flag has been deprecated in favor of `--image`~~ Reverted due to review feedback as this may not work as expected on <4.9 clusters

Fixes [OSD-13298](https://issues.redhat.com//browse/OSD-13298)

### DIFF
--- a/deploy/sre-build-test/20-build-test-ClusterRole.yaml
+++ b/deploy/sre-build-test/20-build-test-ClusterRole.yaml
@@ -13,3 +13,9 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - get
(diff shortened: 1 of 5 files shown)


cblecker:
What version was --docker-image deprecated?

mjlshen:
Seems like 4.9? 
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1995291
PR: https://github.com/openshift/oc/pull/767

cblecker:
If we move to this new flag, I'm concerned about clusters that don't have the new flag yet breaking things. Could we separate that change out?

mjlshen:
Oh, hm I think I see your point since it pulls the image from `image-registry.openshift-image-registry.svc:5000/openshift/cli:latest`. Is that image on cluster dependent on the OCP version running?

cblecker:
Yes, that image contains an `oc` client matching the version of the cluster

PR #1286, opened 2022-09-27
 deploy/ocpbugs-1341-pod-network-connectivity-check-leak: Set base config

mrbarge:
### What type of PR is this?
Feature

The base directory of `deploy/ocpbugs-1341-pod-network-connectivity-check-leak` requires a `config.yaml` to only distribute it to:

- staging/integration clusters
- clusters in the 4.8 through 4.11 range

Presently it will distribute everywhere.

Removal of the staging/integration restriction will occur as part of #1280 

Remediation of OCPBUGS-1341.


NautiluX (commented):
/lgtm

PR #1288, opened 2022-09-29
[WIP] Split installer BYO VPC policy

jharrington22:
### What type of PR is this?
feature

Creates a base installer policy that does not include permissions to create VPC related resources. Adds a new separate policy that contains permissions to create VPCs.

[SDA-4889](https://issues.redhat.com//browse/SDA-4889)

@hunterkepley can you please test the following cases:

1. New ROSA BYO VPC cluster where the installer policy includes only `resources/sts/4.11/sts_installer_permission_policy.json`
2. New ROSA BYO cluster where the installer policy includes only `resources/sts/4.11/sts_installer_permission_policy.json` and remove the statement "BaseInstallerAllowS3Endpoint" - I'm not sure if the installer creates a S3 endpoint within the VPC for public clusters too, if it doesn't the install should succeed and these policies should be covered in the private link policy (yet to be created)
3. New ROSA cluster where the installer policy includes `resources/sts/4.11/sts_installer_permission_policy.json` and `resources/sts/4.11/sts_vpc_permission_policy.json`
    ```


hunterkepley:
Will test starting now, I'll comment on here how the tests turn out

hunterkepley:
@jharrington22 Each test produced a `Ready` cluster with no issues. The only thing I want to make sure is that these are the most minimal possible policies, here is a list Priya provided me when I started out, it's incomplete but it has a lot of the required policies with nothing un-required: [Line 39 here](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/permissions.go#L39) (PermissionCreateBase)

We can talk on slack if you want to discuss this, Priya just brought up in a 1:1 meeting after ours that we should make sure there's nothing unnecessary.

hunterkepley:
Trying on 4.10 and 4.9 now

hunterkepley:
All active versions worked, as well as with a privatelink test (with my own privatelink policy attached) -- this seems like it LGTM

hunterkepley:
@jharrington22 I think we should actually put this new policy into a new file rather than editing the current one.
Older versions will still use the old policy, and it would be best to keep them separate.

Maybe putting it in a new file called `sts_core_installer_policy.json` would be good.

Otherwise, once this is merged, I will be ready to begin development and integration into the newest version via ROSACLI.

resources/sts/4.11/sts_installer_permission_policy.json
```diff
@@ -178,16 +157,26 @@
...
                 "ec2:DescribeVpcEndpointServicePermissions",
                 "ec2:DescribeVpcEndpointServices",
-                "ec2:ModifyVpcEndpointServicePermissions",
-                "kms:DescribeKey",
-                "cloudwatch:GetMetricData"
+                "ec2:ModifyVpcEndpointServicePermissions"
             ],
             "Resource": "*"
-        }
     ]
```
jordigilh, line 181: invalid JSON. 
```
Error: Parse error on line 180: ..."Resource": "*" ]} ----------------------^ Expecting '}', ',', got ']'
```

jharrington22:
Closing in favor or #1385 and #1396

PR #1295, opened 2022-10-12
[OSD-12886] set check error increase over 20min window and trigger an alert on error > 0

T0MASD:
set check error increase over 20min window and trigger an alert on error > 0, update service name to match the name of the deployed service

### What type of PR is this?
_(feature)_

_Fixes #_ [OSD-12886](https://issues.redhat.com//browse/OSD-12886)

if I set the window to 15m I'm seeing gaps, I was told the gaps would make alert resolve, 20m window graph has no gaps, hence the setting. We should reduce the reconcile period to be able to use 15min window

### DIFF
--- a/deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
@@ -11,9 +11,8 @@ spec:
   - name: sre-managed-node-metadata-operator-alerts
     rules:
     - alert: MNMOTooManyReconcileErrors15MinSRE
-      # If the amount of reconciliation errors in the last 15 minutes is > 5 trigger an alert.
-      # Currently the window is 90 minutes, because the reconcile loop uses some backoff mechanism for retries and with a 90 minute window the alert will keep on firing.
-      expr: increase(controller_runtime_reconcile_total{controller="machineset_controller", service="openshift-managed-node-metadata-operator-metrics-service", result="error"}[90m]) >= 5
+      # If reconcile errors have occured in the last 15 minutes, trigger an alert.
+      expr: increase(controller_runtime_reconcile_total{controller="machineset_controller", service="managed-node-metadata-operator-metrics-service", result="error"}[20m])>0
       for: 15m
       labels:
         severity: warning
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
```diff
@@ -11,15 +11,14 @@ spec:
   - name: sre-managed-node-metadata-operator-alerts
     rules:
     - alert: MNMOTooManyReconcileErrors15MinSRE
-      # If the amount of reconciliation errors in the last 15 minutes is > 5 trigger an alert.
-      # Currently the window is 90 minutes, because the reconcile loop uses some backoff mechanism for retries and with a 90 minute window the alert will keep on firing.
-      expr: increase(controller_runtime_reconcile_total{controller="machineset_controller", service="openshift-managed-node-metadata-operator-metrics-service", result="error"}[90m]) >= 5
+      # If reconcile errors have occured in the last 15 minutes, trigger an alert.
+      expr: increase(controller_runtime_reconcile_total{controller="machineset_controller", service="managed-node-metadata-operator-metrics-service", result="error"}[20m])>0
```
iamkirkbater, line 15: Do we want to update this to check for the new node name label value we introduced with https://github.com/openshift/managed-node-metadata-operator/pull/44?

Am I reading this correctly that a single failure would trigger the alert?  This will potentially lead to flapping alerts, and unnecessary toil, so we might want to keep it so that it only triggers if it has had 5 failures on a single node in 15m.

iamkirkbater (changes requested):

deploy/sre-prometheus/100-managed-node-metadata-operator.PrometheusRule.yaml
T0MASD: from experience in querying `controller_runtime_reconcile_total` metrics, limiting with `controller`,`service` and `result` was sufficient, I fear adding additional constrain for `node_name` would be too granular for this rule at this point.

Wrt `single failure` triggering the alert, that's correct, with the available metrics I did not find a way to resolve the alert when condition was resolved in a timely manner (20min not 90min). 

I feel we will have better understanding when we review historical alert data as part of `operator observability`, mnmo alerts are currently in warning, meaning SRE will not handle them, we can adjust that when we have better understanding

iamkirkbater:
```Running 'make' caused changes.  Run 'make' and commit changes to the PR to try again.```

PR #1296, opened 2022-10-13
Remove ineffective role/binding

cblecker:
### What type of PR is this?
cleanup

Removes a role/rolebinding that doesn't have any effect. `packagemanifests` are available at the cluster scope, and `catalogsourceconfigs` don't exist anymore.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jharrington22 (approved):
/lgtm

PR #1298, opened 2022-10-17
Add fluentd job label to FluentdNodeDownSRE alert expr

ninabauer:
### What type of PR is this?
Fix

Adds the fluentd job label to the FluentdNodeDownSRE alert expression to prevent false alerts for clusters with older versions of CLO/ESO.

[OSD-13044](https://issues.redhat.com//browse/OSD-13044)

### DIFF
--- a/deploy/sre-prometheus/extended-logging/101-parsed_fluentd_openshift-logging_collector.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/extended-logging/101-parsed_fluentd_openshift-logging_collector.PrometheusRule.yaml
@@ -31,7 +31,7 @@ spec:
             message: Prometheus could not scrape fluentd {{ $labels.instance }} for more than 10m.
             summary: Fluentd cannot be scraped
           expr: |
-            absent(up{job="collector"} == 1) or absent(up{job="fluentd"} == 1)
+            absent(up{job="collector"} == 1) and absent(up{job="fluentd"} == 1)
           for: 10m
           labels:
             service: fluentd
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/extended-logging/101-parsed_fluentd_openshift-logging_collector.PrometheusRule.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             message: Prometheus could not scrape fluentd {{ $labels.instance }} for more than 10m.
             summary: Fluentd cannot be scraped
           expr: |
-            absent(up{job="collector"} == 1)
+            absent(up{job="collector"} == 1) or absent(up{job="fluentd"} == 1)
```
mjlshen, line 34: Would this cause the alert to fire on both new/old versions of the Cluster Logging Operator?

i.e.
* New version: `absent(up{job="fluentd"} == 1` would fire
* Old version: `absent(up{job="collector"} == 1` would fire

ninabauer: Fixed :)

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -17359,7 +17359,8 @@ objects:
               message: Prometheus could not scrape fluentd {{ $labels.instance }}
                 for more than 10m.
               summary: Fluentd cannot be scraped
-            expr: 'absent(up{job="collector"} == 1)
+            expr: 'absent(up{job="collector"} == 1) and absent(up{job="fluentd"} ==
+              1)
```
sam-nguyen7, line 17363: is there a reason this needs to be a new line?

ninabauer: These were auto generated from running make. This also happened with other expressions in the files.

sam-nguyen7: Gotcha, thanks!

PR #1299, opened 2022-10-17
Changing image URI to the version allowed in image.config.openshift.io

Makdaam:
### What type of PR is this?
bug

It changes the URI of the image from image-registry.openshift-image-registry.svc.cluster.local to image-registry.openshift-image-registry.svc since only the URI ending in svc is in spec.registrySources.allowedRegistries of image.config.openshift.io/cluster

Functionality doesn't change, it's the same image accessed via a different URI.

N/A

It addresses the same repository

- [N/A] Included documentation changes with PR 
- [N/A] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocpbugs-1341-pod-network-connectivity-check-leak/no-seccomp/10-OCPBUGS-1341.CronJob.yaml
+++ b/deploy/ocpbugs-1341-pod-network-connectivity-check-leak/no-seccomp/10-OCPBUGS-1341.CronJob.yaml
@@ -34,7 +34,7 @@ spec:
               operator: Exists
           containers:
           - name: pod-network-connectivity-check-pruner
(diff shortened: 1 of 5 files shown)


cblecker:
We reference the internal registry in a bunch of places in this repo. Should we change it in those other spots too?

mrbarge:
Looks like everywhere else in MCC is referring to it without the `.cluster.local` so this seems okay to go through.
/approve

mrbarge:
/remove-approve
@Makdaam can you please run `make` and commit the changes as well?

Makdaam:
@mrbarge added results of make.

PR #1305, opened 2022-10-24
osd-rebalance-infra-nodes to correct misscheduled dns-default pods

mjlshen:
### What type of PR is this?
feature

Per OSD-13536, `KubeDaemonsetMisScheduled` commonly manifests as a low alert because `openshift-dns/dns-default` pods get scheduled onto infra nodes before their taints get applied. We have this SOP: https://github.com/openshift/ops-sop/blob/master/v4/alerts/KubeDaemonSetMisScheduled.md which directs SREs to this handy script: https://github.com/openshift/ops-sop/blob/master/v4/utils/misscheduled_ds.sh which does this for the SRE.

However, by just putting this script into this osd-rebalance-infra-nodes CronJob, we can periodically check if there are any misscheduled dns-default pods and delete the ones running on infra nodes.


deploy/osd-rebalance-infra-nodes/07-osd-rebalance-infra-nodes.ConfigMap.yaml
```diff
@@ -80,6 +80,28 @@ data:
...
+      NS=$1
+      LABEL=$2
+    
+      readarray -t nodes < <(oc get no -l "${NODE_TAINT}" -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}');
+      for node in "${nodes[@]}"; do
+        misscheduled_pods="$(oc get po -n "${NS}" -l "${LABEL}" -o go-template="{{range .items}}{{if eq .spec.nodeName \"${node}\"}}{{.metadata.name}}{{\"\n\"}}{{end}}{{end}}")";
+        for pod in "${misscheduled_pods[@]}"; do
+          if [[ -n "${pod}" ]]; then
+            echo "Deleting misscheduled pod ${pod} running on ${node}"
+            oc delete po -n "${DS_NAMESPACE}" "${pod}"
```
mrbarge, line 97: Is this meant to be `${NS}` rather than `${DS_NAMESPACE}`?

mrbarge:
I gave this a quick test on a cluster and it seemed to run without issue.
/lgtm

PR #1306, opened 2022-10-26
enable the size based retention for platform prometheus

bmeng:
### What type of PR is this?
_(feature)_

Set size based prometheus retention so that we won't need to be alerted for the prometheus PV disk full for large clusters.

_Fixes #OSD-13455_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mrbarge:
:+1: on the implementation of this, looks good. @NautiluX @fahlmant any concerns from yourselves about this being applied to the fleet?

PR #1311, opened 2022-10-27
OSD-13445: adds vpceacceptance objects for FedRAMP

tonytheleg:
### What type of PR is this?
feature

This PR adds aws-vpce-operator VpcEndpointAcceptance objects to be deployed to FedRAMP hives to handle auto-acceptance of vpce endpoints in new customer clusters for Splunk. This allows us to forward audit logging in a Privatelink network from customer clusters to the FedRAMP splunk instance. Each hive should get a CR for each region to ensure all customer clusters in east and west are setup for logging.

Note, the `assumeRoleArn` and vpce service `id` have been converted to a variable that will be passed via saas file in the FedRAMP  gitlab instance. This seemed like it might be Fedramp-team-eyes-only sensitive data, so I converted it to variable. 

[OSD-13445](https://issues.redhat.com/browse/OSD-13445)

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-avo-resources/fedramp-vpce-acceptance/00-osd-avo-vpcendpointacceptance-us-gov-east-1.yaml
+++ b/deploy/osd-avo-resources/fedramp-vpce-acceptance/00-osd-avo-vpcendpointacceptance-us-gov-east-1.yaml
@@ -4,7 +4,7 @@ metadata:
   name: osd-fr-splunk-acceptance
   namespace: openshift-aws-vpce-operator
 spec:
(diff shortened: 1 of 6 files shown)


mjlshen (approved):
I'm good with this change, but would you mind also slightly refactoring the VpcEndpoint CRs too (totally fine with a followup PR) so that they go into something like a `vpcendpoint` subfolder like the acceptance CRs? Otherwise the folder structure under `avo-resources/` is a bit nonsensical haha

/lgtm

PR #1312, opened 2022-10-27
restructures AVO resources to account for future needs

tonytheleg:
### What type of PR is this?
bug

Follow up to request in #1311 , cleans up the folder structure as hypershift will most likely use AVO as well, leaving room for it now.
N/A
_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -9884,7 +9884,7 @@ objects:
       managed.openshift.io/gitHash: ${IMAGE_TAG}
       managed.openshift.io/gitRepoName: ${REPO_NAME}
       managed.openshift.io/osd: 'true'
-    name: osd-avo-resources-fedramp-vpce-acceptance
+    name: osd-avo-resources-fedramp-vpc-endpoints-us-gov-east-1
```
mjlshen, line 9887: Noted that this name change would probably leave a duplicate SSS behind that will need to be cleaned up

PR #1314, opened 2022-10-28
adding a new recording rule sre:sla:outage_minutes_28d

cattias:
new recording rule to record the number of minutes of outage that counts towards the contractual SLA over a 28d/4w rolling window


NautiluX (commented):
/lgtm

PR #1317, opened 2022-10-31
update the SLA recording rule with RMO

cattias:
idea of this change is to cater for when the api server does not record any metrics (no traffic) in that case we look into RMO and validate or not if we see a 5m outage there if so then it counts towards our SLA


NautiluX (commented):
/lgtm

NautiluX (commented):
/lgtm

PR #1324, opened 2022-11-04
New task to reschedule pods from 'openshift-' namespaces which aren't on infra nodes

bng0y:
### What type of PR is this?
_feature_

In [OSD-13621](https://issues.redhat.com/browse/OSD-13621) Customers are reporting (and complaining) that some of our operators in `openshift-*` are wrongfully scheduled on workers instead of infra nodes, reducing their available capacity.
The investigation has shown that this is due to the usage of `preferredDuringSchedulingIgnoredDuringExecution` in the node affinity, which will allow the pods to be scheduled on workers if the infra nodes don't exist yet. 

While the node affinity gets corrected, this PR intends to be a workaround (similar to [#1305](https://github.com/openshift/managed-cluster-config/pull/1305)). It will reschedule any pod from `openshift-*` that is intended to run on an infra node but is scheduled on a worker node. This allows the pod to be scheduled on infra nodes once they are available.


mjlshen (approved):
/lgtm

PR #1325, opened 2022-11-07
OSD-13649: updaed MUO configmap to ignore etcdGRPCRequestsSlow alert

shitaljante:
- etcdGRPCRequestsSlow alert auto-resolve by self in some time within 5-10 mins.
- This alert has been converted into warning as it is self resolvable in nature. 
- MUO can block the upgrade considering this alert as a critical raised by cluster. 
- To avoid the upgrade cancellations due to this scenario, updated the MUO configmap to ignore etcdGRPCRequestsSlow in pre/post upgrade checks. 

https://issues.redhat.com/browse/OSD-13649

### DIFF
--- a/deploy/managed-upgrade-operator-config/4.7/10-managed-upgrade-operator-configmap.yaml
+++ b/deploy/managed-upgrade-operator-config/4.7/10-managed-upgrade-operator-configmap.yaml
@@ -52,6 +52,7 @@ data:
       - ThanosRuleSenderIsFailingAlerts
       - ThanosRuleHighRuleEvaluationFailures
       - ThanosNoRuleEvaluations
+      # OSD-13649
       - etcdGRPCRequestsSlow
       ignoredNamespaces:
       - openshift-logging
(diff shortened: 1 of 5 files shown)


mrbarge:
We'll also want
`managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml`
updated as well, as that's the version that goes out to all clusters 4.8 and above.

deploy/managed-upgrade-operator-config/4.7/10-managed-upgrade-operator-configmap.yaml
```diff
@@ -52,6 +52,7 @@ data:
       - ThanosRuleSenderIsFailingAlerts
       - ThanosRuleHighRuleEvaluationFailures
       - ThanosNoRuleEvaluations
+      - etcdGRPCRequestsSlow
```
mrbarge, line 55: When adding an alert to this list, please add a leading comment line (`# ....`) that links to the Jira ticket where this decision was made.

shitaljante:
> We'll also want `managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml` [0] updated as well, as that's the version that goes out to all clusters 4.8 and above.
> 
> [0] https://github.com/openshift/managed-cluster-config/blob/master/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml

Sure, I will add. Do we need this change for lower versions i.e 4.5 and 4.6 as well ?

deploy/managed-upgrade-operator-config/4.7/10-managed-upgrade-operator-configmap.yaml
shitaljante: Added change, Thanks :) @mrbarge

PR #1332, opened 2022-11-15
Increase unhealthyConditions timeout as *.metal instances take a long time to reboot

mjlshen:
### What type of PR is this?
bug

When testing with `m5zn.metal` instances, it was noticed that when machine-config-operator works on these nodes and reboots them, this takes a long time and exceeds the existing 480s timeout for how long a node can have the `Ready: Unknown` status condition. I locally removed the timeout to validate and got data that a node took 13 minutes to get out of that `Unknown` status condition, so I am bumping that specific parameter to 15m

For readability, because math is hard, I also converted the existing 480s timeout to 8m (no-op)

OSD-13791

Validated this setting works in stage


fahlmant (approved):
/approve
/lgtm

PR #1333, opened 2022-11-16
Update backplane permissions

cblecker:
### What type of PR is this?
bug

Audits and updates backplane permissions to conform to established RBAC policies

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


iamkirkbater:
Are we sure we want to limit roles such as TAMs, MOBB, and CSAs to only be read-only to the managed namespaces?  I was under the impression that these roles may also need access to customer's workloads, as the roles themselves should be scoped only to organizations who are paying for those services via ocm-resources.

I also wonder how this will effect our ability to send servicelogs for things like PDBs that are blocking a drain if we can't list all the pods running on a node.

I see there's a JIRA ticket for tracking the tooling work, but is there a ticket tracking the work that was done to limit these roles?  I don't see one listed in the PR description.

PR #1334, opened 2022-11-17
Deploy infra-rebalance pieces only to relevant clusters

mjlshen:
### What type of PR is this?
bug

- Splits up RoleBindings into their own file due to a bug in the PR check: https://github.com/openshift/managed-cluster-config/blob/master/hack/app_sre_pr_check.sh#L46-L47 where it cannot handle multi-yaml files. Yes, there is a `yaml. safe_load_all`, but the script would have to change pretty significantly and it's easier to just split all the files apart.
- Fixes broken clustersyncs for STS/PrivateLink clusters as namespaces for CIO and velero operator do not exist by breaking out deploying their rolebindings based on specific cluster types.

ClusterSyncs for those clusters have a status condition:
```
    Result:                Success
    Failure Message:       failed to apply resource 13: error when creating "object": namespaces "openshift-cloud-ingress-operator" not found
    Last Transition Time:  2022-11-17T14:46:37Z
    Name:                  osd-rebalance-infra-nodes
    Observed Generation:   7
```

This is because a couple of our operators are not deployed fleet wide:
- non-FedRAMP:
  - Managed Node Metadata Operator
  - OCM Agent Operator
  - OSD Metrics Exporter
- non-FedRAMP and non-STS only:
  - Velero Operator
- non-FedRAMP and non-STS and non-PrivateLink:
  - Cloud Ingress Operator

[OSD-13621](https://issues.redhat.com//browse/OSD-13621)


mrbarge:
Sorry @mjlshen for causing the need for a rebase here. I'd learnt about this same issue for a different means (OSD-13956) today, as the failing clustersync was causing the infra rebalance job to not land on STS clusters, causing further issues. So I put in a quick fix for that in #1344 and only noticed this one once the 'needs rebase' mail went out! 

This PR is a much more thorough way of solving the issue. Feel free to revert my change if that is easier, or just ping me once this PR is in a mergeable state again and I'm happy to review this.

mjlshen:
/hold

FedRAMP does have CDO

sam-nguyen7 (commented):
/lgtm
/approve

PR #1338, opened 2022-11-21
OSD-13719: adds default allow all netpol's for FedRAMP compliance

tonytheleg:
### What type of PR is this?
bug

FedRAMP clusters currently run the [Compliance Operator](https://github.com/ComplianceAsCode/compliance-operator) to ensure we remain compliant with FedRAMP controls. One of the requirements is to have network policies defined on all Red Hat owned namespaces. The [current rule check](https://github.com/ComplianceAsCode/content/blob/master/applications/openshift/networking/configure_network_policies_namespaces/rule.yml) looks at namespaces starting with `openshift`, but does not account for some SREP specific namespaces used in OSD/ROSA.  This PR adds a default allow all `NetworkPolicy` to satisfy the rule. The 3 namespaces where the policy is being created have no workloads, this is essentially a no-op to remain compliant. 

This change is only being made to FedRAMP clusters. In the long term, I'd like to see if the compliance operator rule could account for namespaces with no pods which would then cover these for us, but that change would take more review, plus a release which may not fall in line with FedRAMP customer go-live.

_Fixes #_ [OSD-13719](https://issues.redhat.com/browse/OSD-13719)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/osd-fedramp-default-netpols/00-default-netpol-osd-experience.yml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-allow-all
+  namespace: osd-experience
```
cblecker, line 5: Does this exist on all clusters?

deploy/osd-fedramp-default-netpols/00-default-netpol-app-sre.yml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-allow-all
+  namespace: app-sre
```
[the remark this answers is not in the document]
tonytheleg: Good catch, this exists on Hive and AppSRE clusters only, I'll have to separate this out so it has its own config so it doesn't get applied to Non-Hive/Non-AppSRE clusters. Thanks

deploy/osd-fedramp-default-netpols/00-default-netpol-osd-experience.yml
tonytheleg: Ditto. Will make changes.

tonytheleg:
@cblecker updated. I made a separate config/dir for Hives. For AppSRE there are no labels i can trigger off of, going to see if this can be accomplished in A-I in FedRAMP separately through namespace files that exist.

PR #1344, opened 2022-11-23
OSD-13956 prevent CIO resources landing on STS clusters

mrbarge:
### What type of PR is this?
Bug

Prevents the `osd-rebalance-infra-nodes` SelectorSyncSet from trying to install a RoleBinding resource into the `openshift-cloud-ingress-operator` namespace for STS clusters, as that namespace will not exist (introduced in #1327)

This currently causes a ClusterSync failure for the SelectorSyncSet, which means that the resources relating to it are currently not being distributed to STS clusters. This in turn is impacting newly provisioned STS clusters which end up with their infra nodes unbalanced from a cluster-monitoring-operator perspective, which in turn impacts the ability to launch successful upgrades.

The renaming of the RoleBinding file is due to https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/mcc-modify-roleref.md

[OSD-13956](https://issues.redhat.com//browse/OSD-13956)

The rebalance CronJob script will still attempt to check in the `openshift-cloud-ingress-operator` namespace for pods that are not situated on infra nodes and need to be. However, this does not cause the job to fail. It simply prints an error and moves on. So I have left the `CIO` namespace in there rather than split the script out into two separate directories or anything smarter.

Example job execution on an STS cluster with a missing namespace:
```
INFO: Rebalancing misscheduled pods from openshift- namespaces
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:openshift-monitoring:osd-rebalance-infra-nodes" cannot list resource "pods" in API group "" in the namespace "openshift-doesnotexist"
INFO: no misscheduled pods in openshift-* namespaces
```

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mrbarge:
The build failure seems like a red herring for what this change is trying to do. I'll poke at it later.

NautiluX (commented):
/lgtm

PR #1356, opened 2022-11-30
Update permission for osde2e namespaces

ritmun:
Added osde2e-* to namespaces to which srep have access without elevating permissions

internal slack thread for ref  https://coreos.slack.com/archives/CCX9DB894/p1669841886918709

### What type of PR is this?
_feature_

for srep who need to debug osde2e pods. 

draft: need a jora

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


ritmun:
This is a proposed change to allow access to osde2e namespaces without elevated permissions. This is useful when debugging osde2e on stage clusters. This PR is meant to propose this change, as I haven't found a different way to do so.

PR #1359, opened 2022-12-02
Update Cronjobs from batch/v1beta1 to batch/v1

cblecker:
### What type of PR is this?
bug

cronjobs.v1beta1.batch has been deprecated starting in OpenShift 4.8 and removed in 4.12+ and has since been replaced with batch/v1

Additionally adds the timeout value from #1362 to the remaining cronjobs now that they are all batch/v1.

fixes [OSD-12458](https://issues.redhat.com//browse/OSD-12458)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/bz1980755/10-bz1980755.CronJob.yaml
+++ b/deploy/bz1980755/10-bz1980755.CronJob.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: bz1980755
@@ -11,6 +11,7 @@ spec:
   schedule: "25 3 * * *"
   jobTemplate:
     spec:
+      ttlSecondsAfterFinished: 86400
       template:
         spec:
           affinity:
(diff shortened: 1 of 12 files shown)


fahlmant:
@cblecker is the v1 version available in <4.8?

jewzaam:
Just to bring it into the conversation, this kind of question is why we have a support lifecycle.  Because trying to support all the minor versions for all time is not sustainable. 4.8 went EOL Sept 27, 2022.  https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-life-cycle.html#sd-life-cycle-dates_osd-life-cycle

cblecker:
@fahlmant it is not. However, all this means is that we might get some clustersync apply failures on these clusters, but the existing cronjobs will still run

jbpratt:
woops @cblecker, sorry about that, didn't notice this PR

 https://github.com/openshift/managed-cluster-config/pull/1364

fahlmant (approved):
/lgtm
/approve

PR #1369, opened 2022-12-05
[OSD-14160] Add alert to detect stuck node.

bergmannf:
Some nodes do not alert, but are broken - one indicator for this is pods being stuck in ContainerCreating as well as Terminating state on the same node.

This alert should detect nodes that have containers stuck in both states at the same time.

### What type of PR is this?
feature

Adds a new alert to detect broken nodes earlier.

_Fixes #[OSD-14160](https://issues.redhat.com//browse/OSD-14160)_

- [x] Tested latest changes against a cluster: tested in CRC cluster with stuck pods.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-node-stuck-terminating-creating.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-node-stuck-terminating-creating.PrometheusRule.yaml
@@ -1,0 +1,31 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-node-stuck
+    role: alert-rules
+  name: sre-node-stuck
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-node-stuck
+    rules:
+    - alert: KubeNodeStuckWithCreatingAndTerminatingPodsSRE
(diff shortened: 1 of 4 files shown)


T0MASD:
@bergmannf thanks for your PR! I was looking at the query which is working but seems little complex (there's a `count(count(...))` which I don't understand ). 

Generally to find issues with the nodes we should look at the `crio` and `kubelet` metrics:
```
count by (__name__) ({__name__=~".*crio.*"})
count by (__name__) ({__name__=~".*kubelet.*"})`
```
What is your opinion about splitting this query into 2 sepatate alerts for pods in our managed namespaces that are in non running states and using combination of crio/kublet metrics to determine failed node?

bergmannf:
@T0MASD Right now the problem is that we do not know if the problems comes from `crio` or maybe `networking` or maybe something else entirely, so I am don't think switching metrics will help a lot.

I think only having a metric that would be based on certain strings in the kubelet logs might work, but AFAIK that is not doable without using the logging-stack (and not even sure if it's doable then).

The nested counts are mostly to align the labels and reduce dimensions right - I can try to see if there is a way to lessen those however.

PR #1372, opened 2022-12-06
Adding a script to generate policy configuration for limited support HyperShift clusters

bdmiller3:
### What type of PR is this?
Feature

Why: Hypershift clusters that enter limited support need to have a ConfigMap (indicating limited support) managed by an ACM policy. There is an existing manifest for this ConfigMap (see https://issues.redhat.com/browse/OSD-13229 for further details)

https://issues.redhat.com/browse/OSD-13677
_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


lnguyen1401:
@bdmiller3 , can you rebase the PR and run `make` locally, then push the changes?

deploy/osd-limited-support/policy-generator-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+policyDefaults:
+    namespace: openshift-osd-metrics
+    placement:
+      clusterSelectors:
+        api.openshift.com/limited-support: true
+    remediationAction: enforce
+policies:
+  - name: osd-limited-support-configmap
+    manifests:
+      - path: 00-limited-support.ConfigMap.yaml
```
lnguyen1401, line 14: Do you still need this file? Because the `generate-policy-config.py` script would fill this out and save it with the other configs in the temp_directory. 

If the file is still needed, i think it should be moved to another directory, that's not /deploy (maybe /script or /resources). that's because this policy-generator-config.yaml is here, it will be included in the selectorsyncset.

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -14791,6 +14849,20 @@ objects:
...
+      policyDefaults:
+        namespace: openshift-osd-metrics
+        placement:
+          clusterSelectors:
+            api.openshift.com/limited-support: true
+        remediationAction: enforce
+      policies:
+      - name: osd-limited-support-configmap
+        manifests:
+        - path: 00-limited-support.ConfigMap.yaml
```
lnguyen1401, line 14865: This is what i meant from the first comment.

hack/00-osd-managed-cluster-config-production.yaml.tmpl
```diff
@@ -14791,6 +14849,20 @@ objects:
...
+      policyDefaults:
+        namespace: openshift-osd-metrics
+        placement:
+          clusterSelectors:
+            api.openshift.com/limited-support: true
+        remediationAction: enforce
+      policies:
+      - name: osd-limited-support-configmap
+        manifests:
+        - path: 00-limited-support.ConfigMap.yaml
```
lnguyen1401, line 14865: (same comment as #2)

hack/00-osd-managed-cluster-config-stage.yaml.tmpl
```diff
@@ -14791,6 +14849,20 @@ objects:
...
+      policyDefaults:
+        namespace: openshift-osd-metrics
+        placement:
+          clusterSelectors:
+            api.openshift.com/limited-support: true
+        remediationAction: enforce
+      policies:
+      - name: osd-limited-support-configmap
+        manifests:
+        - path: 00-limited-support.ConfigMap.yaml
```
lnguyen1401, line 14865: (same comment as above)

deploy/acm-policies/50-GENERATED-osd-limited-support.Policy.yaml
```diff
@@ -0,0 +1,61 @@
+---
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+    annotations:
+        policy.open-cluster-management.io/categories: CM Configuration Management
+        policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
+        policy.open-cluster-management.io/standards: NIST SP 800-53
+    name: osd-limited-support
+    namespace: openshift-rbac-policies
```
sam-nguyen7, line 10: Is there a better namespace for this policy to be in as it's not rbac related?

deploy/osd-limited-support/policy-generator-config.yaml
bdmiller3: Good catch, this does need to be removed.

lnguyen1401:
Thanks for the great work ! 

/lgtm

deploy/acm-policies/50-GENERATED-osd-limited-support.Policy.yaml
```diff
@@ -0,0 +1,61 @@
...
+                name: osd-limited-support
+            spec:
+                object-templates:
+                    - complianceType: musthave
+                      objectDefinition:
+                        apiVersion: v1
+                        kind: ConfigMap
+                        metadata:
+                            name: limited-support
+                            namespace: openshift-osd-metrics
```
fahlmant, line 27: What happens if this namespace isn't present on the HCP?

enxebre:
It's very unlikely we'll want to put a configmap in the guest cluster to signal this. What's the criteria for "limited support" so I can get more context?

fahlmant:
@enxebre whatever the mechanism, we are aiming to have it match between standard rosa and hypershift rosa. Currently, a configmap deployed to the cluster is how we're measuring this in classic ROSA. The criteria for limited support can be found here: https://docs.openshift.com/rosa/rosa_architecture/rosa_policy_service_definition/rosa-service-definition.html#rosa-limited-support_rosa-service-definition Its up to the discretion of SRE to place a cluster in limited support. Do you have a proposal for what we can use in both standard and hypershift rosa if not a configmap?

scripts/policy-paths.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - backplane
+    - backplane/srep
+    - ccs-dedicated-admins
+    - customer-registry-cas
+    - osd-openshift-operators-redhat
+    - osd-pcap-collector
+    - osd-limited-support
+
+limited-support-policies:
+  namespace: openshift-limited-support-policies
```
lnguyen1401, line 16: Hey Ben, this new Namespace needs to be created like this 
 `so if the new policies don't belong to openshift-managed-rbac-config, create a manifest to create a new namespace. Example: https://github.com/openshift/managed-cluster-config/blob/master/deploy/acm-policies/00-openshift-managed-rbac-policies.Namespace.yaml`

deploy/acm-policies/50-GENERATED-osd-limited-support.Policy.yaml
bdmiller3: Per https://coreos.slack.com/archives/CMD1L32M9/p1672776350672259 all policies will be in a single namespace moving forward. There will be an upcoming PR to change the name to something other than openshift-rbac-policies, so for now this will revert to openshift-rbac-policies.

PR #1378, opened 2022-12-08
[OSD-13637] Add new rule to monitor missing apiserver pods.

bergmannf:
Right now if one of the kube-apiservers is going missing no dedicated alert will fire, so this one should be turned into a critical alert once it's mature enough.

It compare the running pods matching the naming pattern of kube-apiserver pods to all master nodes.
This is based on the assumption that every master must run an apiserver and if it isn't something is wrong.

The SOP is added in this PR: https://github.com/openshift/ops-sop/pull/2369

### What type of PR is this?
feature

_Fixes #[OSD-13647](https://issues.redhat.com//browse/OSD-13647)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-kube-apiserver-missing-on-node.yaml
+++ b/deploy/sre-prometheus/100-kube-apiserver-missing-on-node.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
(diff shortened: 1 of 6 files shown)


mjlshen (approved):
/lgtm

bergmannf:
I double checked that the alerts are still triggered, by removing the manifests from one master-node:
image

As the SOP has been merged already, maybe we can go ahead with this PR as well.

PR #1384, opened 2022-12-10
Fix up generation

cblecker:
### What type of PR is this?
bug

In certain local cases, generation was failing opaquely. This cleans up the steps a bit, swaps out the image used, and increases logging verbosity.

It also uses case insensitive matching when sorting the files, so that it doesn't run differently depending on if it's run on Linux vs macOS/Windows.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,29 @@
-FROM openshift/origin-cli:v3.11 AS builder
-
-RUN cp /bin/oc /
+FROM openshift/origin-cli:4.12 AS builder
 
 # Multistage with python
-FROM python:2.7.15 AS runner
+FROM ubi8/python-39 AS runner
 
 # Bring oc binary to python image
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,29 +1,29 @@
-FROM openshift/origin-cli:v3.11 AS builder
-
-RUN cp /bin/oc /
+FROM openshift/origin-cli:4.12 AS builder
```
mjlshen, line 1: Should we be explicit with using quay here (otherwise it may default to Docker Hub)?

```suggestion
FROM quay.io/openshift/origin-cli:4.12 AS builder
```

cblecker: These are using imagestreams in the build cluster

PR #1387, opened 2022-12-14
[OSD-12648] Add pagerduty service orchestration rule configmap

a7vicky:
### What type of PR is this?
_feature_

- Add configmap which is used by pagerduty-operator for service orchestration rule to suppress warning alerts.

_Fixes_ #[OSD-12648](https://issues.redhat.com/browse/OSD-12648)

### DIFF
--- a/deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
+++ b/deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
@@ -1,0 +1,32 @@
+apiVersion: v1
+data:
+  service-orchestration.json: |
+    {
+    "orchestration_path": {
+    "catch_all": {
+        "actions": {}
+    },
+    "sets": [
+    {
+        "id": "start",
+        "rules": [
+        {     
+            "actions": {
+            "suppress": true
+            },
+            "conditions": [
+            {
+            "expression": "event.severity matches 'Warning'"
+            }
+            ]
+        }
+        ]
+    }
+    ],
+    "type": "service"
+    }
+    }
+kind: ConfigMap
+metadata:
+  name: osd-serviceorchestration
+  namespace: pagerduty-operator
--- a/deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
+++ b/deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
@@ -30,3 +30,5 @@ kind: ConfigMap
 metadata:
   name: osd-serviceorchestration
   namespace: pagerduty-operator
+  labels:
+    api.openshift.com/managed: "true"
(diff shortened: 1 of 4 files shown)


deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+            ]
+        }
+        ]
+    }
+    ],
+    "type": "service"
+    }
+    }
+kind: ConfigMap
+metadata:
```
bmeng, line 30: The PD operator will read the configmap which has the same label as the clusterdeployment, for filter.
So in here, I think we should have the label which indicates the cluster is managed as well.
```
metadata:
  labels:
    api.openshift.com/managed: "true"

File: deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+    {
+        "id": "start",
+        "rules": [
+        {     
+            "actions": {
+            "suppress": true
+            },
+            "conditions": [
+            {
+            "expression": "event.severity matches 'Warning'"
```
bmeng, line 19: I am not sure if it should be `warning` or `Warning`. From the alert payload. seems it should be `warning`

deploy/pagerduty-service-orchestration/00-osd-serviceorchestration-configmap.yaml
a7vicky: Thanks for the review @bmeng . I have addressed this in my latest commit

mrbarge:
This lgtm from a technical standpoint but I think deploying resources strictly onto Hive shards from MCC is a bit of an MCC-anti pattern that we would want to mostly avoid, in favour of app-interface.

If we could please take a moment to check how straightforward to bring this across to app-interface that would be appreciated. 

What I believe this would involve is:
- We create a `pagerduty-operator` directory for each Hive environment in `resources` (ie. `resources/hive-stage/pagerduty-operator`)
- We place our service orchestration configmap in that directory (ie. `resources/hive-stage/pagerduty-operator/service-orchestration.configmap.yaml`)
- We edit the `data/services/osd-operators/namespaces/*/pagerduty-operator.yml` file for each hive shard to define a new `openshiftResources` element (pointing to the environment-specific file):

```
- provider: resource
  path: /hive-stage/pagerduty-operator/service-orchestration.configmap.yaml
```

mrbarge:
I've made an attempt to do this in app-interface
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/54283

/hold 
whilst that is reviewed

mrbarge:
Okay, the app-interface approach is not really workable because managing it in `/resources` places it outside of our ability to self-manage changes to. Nor do I think that putting it in the pagerduty-operator's OLM template is an alternative, because this file does not really feel like something that should become part of the catalog IMO.

The option is a separate, parameterized template inside PDO that is referenced and populated via SaaS parameter, however, trying to get the rules being processed as a template parameter without having the format mangled is proving difficult if not impossible.

In the short term I'm not really seeing other options here other than to go with MCC and retrofit it into app-interface later if we can.

/unhold

PR #1392, opened 2022-12-16
Tune the queueConfig of the remote write of hypershift tenant

wanghaoran1988:
### What type of PR is this?

We need tune the queue config a bit to better align the rhobs tenant performance.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


supreeth7:
How about we tweak `maxBackOff` to 256s instead?
I believe this would give us a better network headroom.

PR #1394, opened 2022-12-19
Add Managednotification and Prometheus rule for multiple ingress controllers

hectorakemp:
### What type of PR is this?
_feature_

This pull request sends a service log to the customer if we get an alert that they have manually created an ingress controller outside of the `custom-domains-operator`. It uses the `splunk-audit-exporter` to pick up the non-system `create` verb, and follows the metric example set out in https://docs.google.com/presentation/d/1SEPCs9zOjy9KjgTXwznnJ9hCNMBOWpbz0vir5iCl_H8/edit#slide=id.g1a09e448a9b_0_139. 
https://issues.redhat.com/browse/OSD-12599
Depends on https://github.com/openshift/splunk-forwarder-operator/pull/167
- [x] Wording needs to be confirmed by the BU - as mentioned in the PR. (have corresponded with a PM about wording, waiting for approval).
- [x] Depends on https://github.com/openshift/splunk-forwarder-operator/pull/167 being merged

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -6,6 +6,14 @@ metadata:
 spec:
   notifications:
     - activeBody: |-
+        Your cluster requires you to take action as there are multiple ingress controllers detected. Red Hat SRE strongly recommends using Openshift-managed Custom Domains instead, as multiple ingress controllers, if improperly configured, can lead to problems with internal and external networking.
+      name: MultipleIngressControllersDetected
+      resendWait: 24
+      resolvedBody: |-
+        Your cluster no longer has multiple Ingress Controllers installed. No additional action on this issue is required.
+      severity: Info
+      summary: "Multiple ingress controllers detected"
+    - activeBody: |-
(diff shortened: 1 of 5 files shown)


deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
```diff
@@ -5,6 +5,14 @@ metadata:
   namespace: openshift-ocm-agent-operator
 spec:
   notifications:
+    - activeBody: |-
+        Your cluster requires you to take action as there are multiple ingress controllers detected. Red Hat SRE strongly recommends using Openshift-managed Custom Domains instead, as multiple ingress controllers can lead to problems with internal and external networking.
```
wgordon17, line 9:
```suggestion
        Your cluster requires you to take action as there are multiple ingress controllers detected. Red Hat SRE strongly recommends using Openshift-managed Custom Domains instead, as multiple ingress controllers, if improperly configured, can lead to problems with internal and external networking.
```

hectorakemp: Hi @wgordon17 - thank you for the review. I'll amend the message with your suggested changes now.

hectorakemp:
> @hectorakemp does this also consider the apps2 IngressController that cloud-ingress-operator creates?

Hi @fahlmant - it does, because any resource (no matter what kind) created by an operator will be prefixed with a `system:*` username. In the dependent [PR](https://github.com/openshift/splunk-forwarder-operator/pull/167), we target usernames that do NOT match this pattern, which filters out anything created by an operator or automated resource. 

So the short answer is yes.

PR #1373, opened 2022-12-07
Add env label for the metrics on management clusters

wanghaoran1988:
### What type of PR is this?
_(bug/feature/cleanup/documentation)_

### What this PR does / why we need it?
We will ship all stg hypershift metrics to production tenant due to capacity issue on staging hypershift tenant.
### Pre-checks (if applicable):
- [x] Tested latest changes against a cluster
- [x] Included documentation changes with PR
- [x] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1403, opened 2023-01-04
Adding a new label to identify metrics origin on RHOBS

Nikokolas3270:
### What type of PR is this?
_Feature_

Metrics shipped by Hypershift to RHOBS will come from 3 distincts sources:
- From the Prometheus handled by the CMO operator running in the Management Cluster.
- From the Prometheus handled by the OO operator running in the Management Cluster.
- From the Prometheus handled by the CMO operator running in the Data Plane.

The 2 first sources do not need to be distinguished one from the other and will be `source` labelled to value `HCP` for Hosted Control Plane. Value for the last source will be set to `DP` (Dala Plane). 
Be able to clearly identify the source of each metric will avoid confusion and potential conflicts.

This change is needed to properly support the alerts defined there:
https://gitlab.cee.redhat.com/service/rhobs-rules-and-dashboards/-/tree/main/rules/hypershift-platform
Those rules were defined while implementing the following tickets; they already assume that a `source` label is set to `HCP`:
- [OSD-12616](https://issues.redhat.com/browse/OSD-12616)
- [OSD-13687](https://issues.redhat.com/browse/OSD-13687)

Please tell me if a dedicated ticket is needed for that.

Change is very similar to this one also done with the same context in mind (Hypershift): https://github.com/openshift/managed-cluster-config/pull/1373

However, here we are adding the label at a lower level (through some `writeRelabelConfig`)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/hypershift-cluster-monitoring-config/cluster-monitoring-config.yaml
+++ b/deploy/hypershift-cluster-monitoring-config/cluster-monitoring-config.yaml
@@ -45,6 +45,8 @@ data:
             action: replace
           - sourceLabels: [__name__]
             action: keep
+          - targetLabel: source
+            replacement: HCP
           queueConfig:
             capacity: 2500
             maxShards: 1000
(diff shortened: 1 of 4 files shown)


Nikokolas3270:
A similar (but not dependent) change has been submitted on OO config.

PR #1406, opened 2023-01-07
Fix mount permissions for containerized generation

cblecker:
### What type of PR is this?
bug

This uses the `PODMAN_USERNS` env var to allow podman to correctly map the UID in the container to the user's filesystem UID.

fixes [OSD-14145](https://issues.redhat.com//browse/OSD-14145)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,8 @@ endif
 
 
 CONTAINER_ENGINE?=$(shell command -v docker 2>/dev/null || command -v podman 2>/dev/null)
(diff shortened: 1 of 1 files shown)


lnguyen1401:
hello Christoph, it looks like the Makefile is not working on my linux laptop. It's failing with this error:
`Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: unrecognized namespace mode keep-id:uid=1001 passed
`

PR #1410, opened 2023-01-10
Allow CEE to view Addons

gjbravi:
This PR is for the CEE team and allows us to view and list child objects of Addons. The change consists of displaying any resource that is a child of the "addons.managed.openshift.io" apiGroup. With this new feature, CEE will be able to easily view and enumerate the various resources associated with Addons, and improve our overall workflow.

This request originated from https://issues.redhat.com/browse/OSD-14426

Thanks.
gbravi

### DIFF
--- a/deploy/backplane/cee/01-cee-cluster-readers-ClusterRole.yml
+++ b/deploy/backplane/cee/01-cee-cluster-readers-ClusterRole.yml
@@ -30,3 +30,12 @@ rules:
   - get
   - list
   - watch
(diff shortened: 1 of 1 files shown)


mrbarge:
@gjbravi thanks for the PR.

Could you please also ensure that you run `make` from the base directory (per the [README](https://github.com/openshift/managed-cluster-config#how-to-use-this-repo)) and commit the additional files that will have been changed as a result?

PR #1411, opened 2023-01-10
Add delete/patch permissions to cpms for SREs

AlexVulaj:
As part of OCP 4.12, ControlPlaneMachineSet (CPMS) was added as a new custom resource in openshift-machine-api. SREs need these added permissions to properly service this CR in new clusters.

Linked JIRA:
https://issues.redhat.com/browse/OSD-14591

### DIFF
--- a/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
+++ b/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
@@ -115,6 +115,7 @@ rules:
   - machines
   verbs:
   - delete
+# SRE can delete and patch ControlPlaneMachineSet https://issues.redhat.com/browse/OSD-14581 
 - apiGroups:
   - machine.openshift.io
   resources:
--- a/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
+++ b/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
@@ -119,7 +119,7 @@ rules:
 - apiGroups:
   - machine.openshift.io
   resources:
-  - controlplanemachineset
+  - controlplanemachinesets
   verbs:
   - delete
   - patch
--- a/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
+++ b/deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
@@ -115,6 +115,13 @@ rules:
   - machines
   verbs:
   - delete
+- apiGroups:
+  - machine.openshift.io
+  resources:
+  - controlplanemachineset
+  verbs:
+  - delete
+  - patch
 # SRE can view api request counts
 - apiGroups:
   - apiserver.openshift.io
--- a/deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
@@ -134,6 +134,13 @@ spec:
                               verbs:
                                 - delete
                             - apiGroups:
+                                - machine.openshift.io
+                              resources:
+                                - controlplanemachinesets
+                              verbs:
+                                - delete
+                                - patch
+                            - apiGroups:
                                 - apiserver.openshift.io
                               resources:
                                 - apirequestcounts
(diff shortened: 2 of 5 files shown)


deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
```diff
@@ -115,6 +115,13 @@ rules:
   - machines
   verbs:
   - delete
+- apiGroups:
```
jewzaam, line 118:
```suggestion
# SRE can delete and patch ControlPlaneMachineSet https://issues.redhat.com/browse/OSD-14581 
- apiGroups:
```

deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
```diff
@@ -115,6 +115,13 @@ rules:
...
+  - machine.openshift.io
+  resources:
+  - controlplanemachineset
```
jewzaam, line 121:
```suggestion
  - controlplanemachinesets
```

jewzaam (changes requested):

AlexVulaj:
Generated files are missing from this PR (and likely causing the build failure) because I'm having issues running `make` on my m1 mac with podman. If anyone wants to pull this down and generate those it would be greatly appreciated, in the meantime I'll keep investigating.

mrbarge:
@AlexVulaj in case it helps, I raised a PR to your branch with the generated files as I couldn't commit directly to your own.
https://github.com/AlexVulaj/managed-cluster-config/pull/1

AlexVulaj:
@cblecker This is an action that will always be the exact same every time, and I don't think it'll necessarily be a regular thing SREs need to do daily.  I'm pinging @iamkirkbater for a second opinion here.

fahlmant:
@AlexVulaj Is there an associated Jira? 

Would patch be sufficient? IIUC, a delete will cause rolling reboots of the control plane nodes, is that necessary for common actions SREP should take? If not, it might be better to gate that one behind elevation because it could be more sensitive.

AlexVulaj:
@fahlmant Based on our conversations in slack, it looks like our path forward here is to go through with what's shown.  I've also added the JIRA ticket to this PR's description.

fahlmant:
@AlexVulaj thanks for the update. I don't see any description of what causes the rolling reboots though. Is that done through a PATCH?

AlexVulaj:
@fahlmant In this situation, there won't be any rolling reboots. When CPMS comes back up, it will already have its spec updated to match the current state of the machines, and thus will not perform any new actions on them.

In the case of CIO updates causing rolling reboots, this is happening because CIO is updating the machines while CPMS is Active, so CPMS triggers itself to bring those machines back into sync with what it believes to be the source of truth.

fahlmant:
I'm not asking about CIO specifically. I'm asking what sequence of events leads to a rolling reboot. If an SRE runs a patch on CPMS incorrectly, can they trigger a rolling reboot?

AlexVulaj:
@fahlmant Any update to the cpms spec will trigger a rolling reboot. If anything updates a machine behind the scenes, CPMS will replace just that machine. CPMS uses immutable infrastructure, so if it detects any drift then it will kill the out of date machine(s) and spin up new ones.

I'll note that CPMS allows a surge to 4 nodes as well, so it will make sure a new instance is ready to go with all workloads moved over before tearing down the old machine.

fahlmant:
Hmm so giving SREP patch is a bit dangerous without requiring elevation

AlexVulaj:
I don't necessarily disagree, my goal was to avoid generating compliance tickets every time an SRE needs to update a control plane. Maybe we do in fact want that, but I'll leave that decision up to you. We can always decline this PR and I can update my [SOP changes](https://github.com/openshift/ops-sop/pull/2371) to require `--as backplane-cluster-admin` for the delete/patch commands.

AlexVulaj:
Though there's still ongoing conversation about if we want to go through with this PR or not, I will note that I have in fact tested it on a staging cluster by pausing the hive syncset and manually adding these updated permissions to the `backplane-srep-admins-cluster` role. If we decide to move forward with this change, it's ready to go.

jewzaam:
With the risk of patch causing a rolling update I think relying on elevation will be better.  It's more paperwork and it will help ensure we don't do this accidentally.. nobody likes paperwork.  Long term we'll have the bug fixed and can have OCM manage this, SRE won't be manually resizing.  Assuming ADR (yet to be drafted) is accepted!

So I think we should ditch this PR and add elevation in the SOP @AlexVulaj

AlexVulaj:
@jewzaam sounds good!  I'll go forward with that.

PR #1413, opened 2023-01-11
Add iam:SimulatePrincipalPolicy to SRE support role

typeid:
### What type of PR is this?
feature

This addition to the SRE support role allows us to simulate API calls against all policies attached to roles. The change will allow us to check if all the needed permissions are available on a STS cluster. 
(Part of https://issues.redhat.com/browse/OSD-11275)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.10/sts_support_permission_policy.json
+++ b/resources/sts/4.10/sts_support_permission_policy.json
@@ -140,6 +140,7 @@
                 "elasticloadbalancing:DescribeTargetHealth",
                 "iam:GetRole",
                 "iam:ListRoles",
+                "iam:SimulatePrincipalPolicy",
                 "kms:CreateGrant",
                 "route53:GetHostedZone",
                 "route53:GetHostedZoneCount",
--- a/resources/sts/4.10/sts_support_permission_policy.json
+++ b/resources/sts/4.10/sts_support_permission_policy.json
@@ -140,7 +140,6 @@
                 "elasticloadbalancing:DescribeTargetHealth",
                 "iam:GetRole",
                 "iam:ListRoles",
-                "iam:SimulatePrincipalPolicy",
                 "kms:CreateGrant",
                 "route53:GetHostedZone",
                 "route53:GetHostedZoneCount",
@@ -156,6 +155,18 @@
                 "tiros:GetQueryAnswer",
                 "tiros:GetQueryExplanation"
             ],
+            "Resource": "*"
+        },
+        {
+            "Action": [
+                "iam:SimulatePrincipalPolicy"
+            ],
+            "Effect": "Allow",
+            "Condition": {
+                "StringEquals": {
+                    "aws:ResourceTag/red-hat-managed": "true"
+                }
+            },
             "Resource": "*"
         },
         {
--- a/resources/sts/4.11/sts_support_permission_policy.json
+++ b/resources/sts/4.11/sts_support_permission_policy.json
@@ -140,6 +140,7 @@
                 "elasticloadbalancing:DescribeTargetHealth",
                 "iam:GetRole",
                 "iam:ListRoles",
+                "iam:SimulatePrincipalPolicy",
                 "kms:CreateGrant",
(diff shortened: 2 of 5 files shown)


jharrington22:
@typeid Does this policy work? I don't see the condition listed for the `SimulatePrincipalPolicy` https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsidentityandaccessmanagement.html

typeid:
@jharrington22 The policy does work, I tested it against a dummy sts cluster. This is where I got it from: https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_testing-policies.html

Note that we might also want iam:GetContextKeysForPrincipalPolicy to make sure the SimulatePrincipalPolicy is as precise as possible:
```
To allow API users to simulate policies attached to IAM users, user groups, roles, or resources
Include the following actions in your policy:
- iam:GetContextKeysForPrincipalPolicy
- iam:SimulatePrincipalPolicy

...

In both cases, the API operations simulate the effect of one or more policies on a list of actions and resources. Each action is paired with each resource and the simulation determines whether the policies allow or deny that action for that resource. You can also provide values for any context keys that your policies reference. You can get the list of context keys that the policies reference by first calling [GetContextKeysForCustomPolicy](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetContextKeysForCustomPolicy.html) or [GetContextKeysForPrincipalPolicy](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetContextKeysForPrincipalPolicy.html). If you don't provide a value for a context key, the simulation still runs. But the results might not be reliable because the policy simulator cannot include that context key in the evaluation.
```

jharrington22 (commented):
We will initiate this API call with the `red-hat-managed=true` tag correct? The condition key only requires the `SimulatePrincipalPolicy` API call to contain the tags? Reason is that we didn't introduce tags on all AWS resources in ROSA until an early minor release of the 4.10 release cycle.

typeid:
This API call will be initiated on roles/policies with the `red-hat-managed:true` tags to limit our permissions to simulate the principal policies for the roles we created ourselves.

If the tag isn't set yet (old cluster), we will skip the check for those clusters.

typeid:
As an update to if we need `iam:GetContextKeysForPrincipalPolicy` as well: we don't. This permission is only needed if we don't know the context keys, but as it's our own policies, we know them. The addition to the policy as it is in this PR is good to go. 

Here's the example response to an API call on a `red-hat-managed: true` role (`arn:aws:iam::<snip>:role/ManagedOpenShift-ControlPlane-Role`) where I check if `ec2:AttachVolume` is allowed on `*`:

```
{
  EvaluationResults: [{
      EvalActionName: "ec2:AttachVolume",
      EvalDecision: "allowed",
      EvalResourceName: "*",
      MatchedStatements: [{
          EndPosition: {
            Column: 1583,
            Line: 1
          },
          SourcePolicyId: "ManagedOpenShift-ControlPlane-Role-Policy",
          SourcePolicyType: "IAM Policy",
          StartPosition: {
            Column: 41,
            Line: 1
          }
        }],
      OrganizationsDecisionDetail: {
        AllowedByOrganizations: true
      }
    }],
  IsTruncated: false
}
```

Removing the red-hat-managed from the role, I get:
```
error: AccessDenied: User: arn:aws:sts::<snip>:assumed-role/testcb/AIDA3Z7BU7KE7WE64YLCX is not authorized to perform: iam:SimulatePrincipalPolicy on resource: arn:aws:iam::<snip>:role/ManagedOpenShift-ControlPlane-Role because no identity-based policy allows the iam:SimulatePrincipalPolicy action
	status code: 403, request id: e34e1dc7-90d5-41ab-a78d-a36c56c14406
```

This will be handled in CAD to provide best effort permission checks on clusters where the tags are set.

typeid:
/close
We are going another route and don't need this policy.

PR #1416, opened 2023-01-12
[WIP][OSD-13564] Add resources to deploy splunk-audit-exporter.

bergmannf:
The splunk-audit-exporter should be deployed to every hosted controlplane to write the audit events to the filesystem.

There the splunk-forward-operator of the management cluster should pick those files up and actually forward them to splunk.

### What type of PR is this?
feature

Add the splunk-audit-exportter to every hosted controlplane as a first step to implement SDE-2217.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/osd-splunk-audit-exporter/deployment.yaml
```diff
@@ -0,0 +1,71 @@
+---
```
lnguyen1401, line 1: I think this `osd-splunk-audit-exporter` directory should be move out of /deploy, because it will be deployed to the hive cluster itself. (see below)

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -18914,6 +19294,307 @@ objects:
       metadata:
         name: ocm-agent-operator
         namespace: openshift-ocm-agent-operator
+- apiVersion: hive.openshift.io/v1
```
lnguyen1401, line 19297: This is what i meant by the comment above. This selectorSyncSet will be synced onto the classic OSD clusters as well.

deploy/osd-splunk-audit-exporter/deployment.yaml
bergmannf: Hm I tried to move it out - but obviously then it won't generate the policy file anymore (even if I try to put in a relative path like `../osd-splunk-audit-exporter`). But if I try to make the python file pick up that path again it will also include it in the regular OSD clusters again.

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
bergmannf: I added a pretty hacky kludge to get the splunk-audit-exporter only installed on the hypershift clusters, but as that includes some changes to the python script, I guess that should not be merged just like that.

bergmannf:
done via app-interface deployments now, as this is not just a static manifest but an application.

PR #1417, opened 2023-01-13
Stop creation of SSS for CMO config in HyperShift

supreeth7:
- This PR intends to delete `deploy/hypershift-cluster-monitoring-config` as we would no longer need MCC to create CMO config for HyperShift  as this configmap is now being created by OSD Fleet Manager, [ref: !222](https://gitlab.cee.redhat.com/service/osd-fleet-manager/-/merge_requests/222)


deploy/hypershift-cluster-monitoring-config/config.yaml
```diff
@@ -2,7 +2,7 @@ deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   matchExpressions:
     - key: ext-hypershift.openshift.io/cluster-type
-      operator: In
+      operator: NotIn
```
mrbarge, line 5: Doesn't this mean it would then land everywhere but management clusters - including standard osd/rosa clusters where it would clash with the other cluster-monitoring-config we push out?

Its content exactly matches our OSD/ROSA fleet config one: https://github.com/openshift/managed-cluster-config/blob/master/deploy/cluster-monitoring-config/config.yaml

wanghaoran1988: @supreeth7 We can remove the whole dir dirrectly since we are using FM to configure this.

supreeth7: That's true 🤔 
Let me purge the entire directory.

supreeth7: @mrbarge mind giving this another look? Thanks!

PR #1418, opened 2023-01-13
[OSD-8342] ignore the node unschedulable alerts during node update

bmeng:
### What type of PR is this?
_bug_

Fix the un-intended KubeNodeUnschedulableSRE alert during the cluster upgrade

_Fixes #[OSD-8342](https://issues.redhat.com//browse/OSD-8342)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
@@ -11,8 +11,8 @@ spec:
   - name: sre-node-unschedulable
     rules:
     - alert: KubeNodeUnschedulableSRE
-      expr: kube_node_spec_unschedulable > 0
-      for: 80m
+      expr: kube_node_spec_unschedulable > 0 unless on(node) (kube_node_role{role!="master"} and on(node) kube_node_spec_taint{key="UpdateInProgress"})
+      for: 60m
       labels:
         severity: critical
         namespace: openshift-monitoring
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
```diff
@@ -11,8 +11,8 @@ spec:
   - name: sre-node-unschedulable
     rules:
     - alert: KubeNodeUnschedulableSRE
-      expr: kube_node_spec_unschedulable > 0
-      for: 80m
+      expr: kube_node_spec_unschedulable > 0 unless on(node) kube_node_spec_taint{key="UpdateInProgress"} > 0
```
mrbarge, line 14: I rreally like this! May I suggest a slight tweak though?
Could we limit the `UpdateInProgress` filter to only apply to `worker` nodes? For example, I think it is still quite valid for an SRE to be paged if a Master or Infra node remains unschedulable for an hour during an upgrade, as we are responsible for those nodes and that would indicate something potentially quite wrong.

I think this should be as simple as joining it on `kube_node_role{role="worker"}` or something like that, what do you think?

That way, it will only be specifically worker nodes during upgrades that we ignore this for.

bmeng:
Matt, thanks for the suggestion, that makes sense to me. I have updated the rule to alert for master nodes regardless the condition.

mrbarge:
Thanks a lot!
/lgtm

PR #1422, opened 2023-01-18
Fix the rule to only return one upgrade result for recent upgrade

weherdh:
### What type of PR is this?
bug

To fix the issue of returning more than one upgrade_results for the SLO recording rule, which will stop firing PrometheusRuleFailures alert for the upgrade_result metric

_Fixes #_[OSD-14614](https://issues.redhat.com/browse/OSD-14614)

### DIFF
--- a/deploy/sre-prometheus/centralized-observability/100-sre-internal-slo-recording-rules.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/centralized-observability/100-sre-internal-slo-recording-rules.PrometheusRule.yaml
@@ -10,5 +10,5 @@ spec:
   groups:
     - name: sre-internal-slo.rules
       rules:
-        - expr: sre:telemetry:managed_labels * on (sre) group_left(version, alerts, upgradeconfig_name) label_replace(upgradeoperator_upgrade_result, "sre", "true", "", "")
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/centralized-observability/100-sre-internal-slo-recording-rules.PrometheusRule.yaml
```diff
@@ -10,5 +10,5 @@ spec:
   groups:
     - name: sre-internal-slo.rules
       rules:
-        - expr: sre:telemetry:managed_labels * on (sre) group_left(version, alerts, upgradeconfig_name) label_replace(upgradeoperator_upgrade_result, "sre", "true", "", "")
+        - expr: sre:telemetry:managed_labels * on (sre) group_left(version, alerts, upgradeconfig_name) label_replace(topk(1, upgradeoperator_upgrade_result), "sre", "true", "", "")
```
mrbarge, line 13: If there are multiple entries being returned, will this return the result for the most recent upgrade?

mrbarge: Based on some analysis I've done of impacted clusters, the answer would seem to be no. I will still proceed with an lgtm here because, either way, this is an improvement on _nothing_ being sent when the condition arises, but I still feel that we should pursue having MUO clear this metric at the start of a new upgrade as the proper solution.

PR #1429, opened 2023-01-20
Update to return latest upgrade result

weherdh:
### What type of PR is this?
bug

To update the query of upgrade_results to return the latest result

_Fixes #_[OSD-14614](https://issues.redhat.com/browse/OSD-14614)


mrbarge:
I've verified this and it seems to be behaving as expected.
/lgtm

PR #1431, opened 2023-01-20
[OSD-8342] update to not alerting for worker node during upgrade

bmeng:
I will try to explain the the PromQL, it consist of two parts and connected with `or`

For the first part `kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}` it will be alerted for master node which is unschedulable all the time, 

For the second part `kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade.*"}) > 0`,  it will ignore the kubeNodeUnschedule alert during the upgrade period.

The `ignoring(node,container,endpoint,job,namespace,service)` is there due to the `kube_node_spec_unschedulable` and `mapi_machine_set_status_replicas` are having different set of labels, and I just want to use the `mapi_machine_set_status_replicas` as an indicator that the upgrade is in progress.

Make it in a table, when a node is unscheduleable:

| Is it a master or worker? | Is in the middle of upgrade? | Alerting? | Caught by |
| ---- | ---- | ---- | ---- |
| master | yes | yes | kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"} |
| master | no | yes | kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"} |
| worker | yes | no | n/a |
| worker | no | yes | kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade.*"}) > 0 |

### What type of PR is this?
_(bug)_

It is another implementation for https://github.com/openshift/managed-cluster-config/pull/1418 

_Fixes #[OSD-8342](https://issues.redhat.com//browse/OSD-8342)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
@@ -11,8 +11,8 @@ spec:
   - name: sre-node-unschedulable
     rules:
     - alert: KubeNodeUnschedulableSRE
-      expr: kube_node_spec_unschedulable > 0
-      for: 80m
+      expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}) or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade$"}) > 0)
+      for: 60m
       labels:
         severity: critical
         namespace: openshift-monitoring
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -30562,8 +30562,10 @@ objects:
         - name: sre-node-unschedulable
           rules:
           - alert: KubeNodeUnschedulableSRE
-            expr: kube_node_spec_unschedulable > 0
-            for: 80m
+            expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"})
+              or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service)
+              sum(mapi_machine_set_status_replicas{name=~".*upgrade$"}) > 0)
+            for: 60m
             labels:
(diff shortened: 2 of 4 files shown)


deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
```diff
@@ -11,8 +11,8 @@ spec:
   - name: sre-node-unschedulable
     rules:
     - alert: KubeNodeUnschedulableSRE
-      expr: kube_node_spec_unschedulable > 0
-      for: 80m
+      expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}) or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade.*"}) > 0)
```
mrbarge, line 14: A minor tweak here - otherwise there is the potential that if a cluster owner decided to create a machine pool with the word `upgrade` in it, it would be excluded from this alert.

The suggestion below will apply if `upgrade` appears at the very end of the machine set name after the availability zone (which is what MUO does). If a cluster owner created a machine pool named `upgrade`, `upgrade` appears before the availability zone.

```suggestion
      expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}) or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade$"}) > 0)
```

rendhalver: Is there a label we apply to the upgrade nodes? It would make sense to use that here if we do.

mrbarge: I don't believe we label upgrade nodes. We only label the machineset.

PR #1432, opened 2023-01-20
Add prune job to recreate velero-operator metrics service

mrbarge:
### What type of PR is this?
Cleanup

- Creates a new pruning cronjob/SA/RBAC `osd-prune-velero-service-osd-14743` that performs a daily deletion of the `managed-velero-operator-metrics` Service and ServiceMonitor, followed by the operator pod itself. 

This is in order to force the recreation of the Service and ServiceMonitor without the presence of a now-unused port `8686` which was removed as part of the operator's SDKv1 migration. Presently, on clusters which had the operator installed prior to the SDKv1 migration, that port still remains in the Service/ServiceMonitor and results in a `TargetDown` alert now that there is nothing listening on that port.

[OSD-14743](https://issues.redhat.com//browse/OSD-14743)

### DIFF
--- a/deploy/sre-pruning/osd-14743/00-pruning.rbac.Role.yaml
+++ b/deploy/sre-pruning/osd-14743/00-pruning.rbac.Role.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: osd-prune-velero-service-osd-14743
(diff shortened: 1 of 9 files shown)


NautiluX (commented):
/lgtm

cblecker:
/hold
I think https://github.com/openshift/managed-velero-operator/pull/164 might be a better fix.

PR #1433, opened 2023-01-23
[OSD-13637] Fix static pod rules for GCP

bergmannf:
While checking the alerts, I noticed they fire on GCP - this is because the pods are named slight different between GCP and AWS.

Filter the guard pods differently, so both GCP and AWS can be matched with a more generic expression.

This way the expression now works on AWS and GCP.

### What type of PR is this?
bug

Fixes the existing alert rules for static pods to work on AWS and GCP.

Fixes [OSD-13637](https://issues.redhat.com//browse/OSD-13637)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #1436, opened 2023-01-24
Adding custom NodeFilesystemAlmostOutOfSpaceSRE alert

ravitri:
### What type of PR is this?
_(bug)_

Based on [OSD-14857](https://issues.redhat.com/browse/OSD-14857) and [OCPBUGS-6577](https://issues.redhat.com/browse/OCPBUGS-6577), the NodeFilesystemAlmostOutOfSpace alert wrongly fires if a tmpfs mount even with only 4K size is 100% used. This is not actionable by SRE. This issue is fixed in 4.13 for cluster-monitoring-operator but yet to be fixed in older versions. Meanwhile, introducing NodeFilesystemAlmostOutOfSpaceSRE alert which would help reduce some noise till the backports and new release is released and used by the customer(s).

4.13+ Fix - [Link](https://github.com/openshift/cluster-monitoring-operator/blob/release-4.13/assets/node-exporter/prometheus-rule.yaml#L68-L82)

_Fixes #_ [OSD-14857](https://issues.redhat.com/browse/OSD-14857)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-NodeFilesystemAlmostOutOfSpaceSRE.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-NodeFilesystemAlmostOutOfSpaceSRE.PrometheusRule.yaml
@@ -1,0 +1,31 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-nodefilesystemalmostoutofspace-alert
+    role: alert-rules
+  name: sre-nodefilesystemalmostoutofspace-alert
+  namespace: openshift-monitoring
+spec:
+  groups:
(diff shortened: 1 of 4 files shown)


mrbarge:
/hold
/lgtm

Seems good to me. Feel free to release the hold when the associated CAMO changes are merged and promoted.

ravitri:
@mrbarge - Hey Matt, thanks a lot for the review! 

Well, based on feedback in PR https://github.com/openshift/configure-alertmanager-operator/pull/259 by @cblecker, there's an alternate solution to ignore the problematic mountpoint in CAMO itself using `MatchRE` so we won't need a separate SRE alert in MCC after all.

Considering minimal change approach to address the bug, I'll close this PR for now. 

/close

PR #1437, opened 2023-01-24
Don't fail osd-disable-cpms if cpms doesn't exist

cblecker:
### What type of PR is this?
bug

Adds the `--ignore-not-found` flag to this command so that the cronjob doesn't fail if the CPMS doesn't exist

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/lgtm

PR #1438, opened 2023-01-24
Related to MCO-457. KubeletConfig workaround.

Makdaam:
Adding CSIMigrationOpenStack:true override to 4.10.51 which is not listed in candidate-4.10 at this time.

### What type of PR is this?
bug

Version not available in candidate-4.10 is 4.10.51 https://github.com/openshift/cincinnati-graph-data/blob/dc0546a679bb101e8977cee75b1fac9ee89ff991/channels/candidate-4.10.yaml 

Currently all upgrades from 4.10 to 4.11 are blocked due to https://issues.redhat.com/browse/MCO-457 . This is the first step of the process proposed by @jewzaam  to:
1. Set the problematic feature flag during an upgrade to the next version which is not available in any channel yet. (MCO will cause a kubelet restart)
2. Add an exception in CIS to allow upgrade from this specific 4.10.51 version to 4.11 which shouldn't trigger the issue.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/kubelet-config/4.10.51/01-kubelet-config.patch.yaml
+++ b/deploy/kubelet-config/4.10.51/01-kubelet-config.patch.yaml
@@ -1,0 +1,6 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+applyMode: AlwaysApply
+name: custom-kubelet
+patch: '{"spec":{"kubeletConfig":{"featureGates":{"CSIMigrationOpenStack": "true"}}}}'
+patchType: merge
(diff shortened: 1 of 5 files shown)


jewzaam:
/hold

We apply KubeletConfig in MCC.  I want explicit validation that the SSS won't clobber the patch.  Alternative is to do something like was done for pre-4.9:
https://github.com/openshift/managed-cluster-config/tree/master/deploy/kubelet-config/pre-4.9

cblecker:
Yeah this implementation will fight

Makdaam:
It probably won't go through.
Per OSD-14865 the suggested workaround is allowlisting the exception without the wait. Closing for now.

As far as I understood from your comments going ahead with a full resource instead of a patch and matchExpression logic is the right way to do it?

PR #1452, opened 2023-02-01
OSD-15020 Adding prune job for RPO service and servicemonitor

ravitri:
### What type of PR is this?
_(cleanup)_

In [OSD-12370](https://issues.redhat.com/browse/OSD-12370) the bug of mismatch in the labels of pod, service and servicemonitor for rbac-permissions-operator(RPO) was fixed, merged and promoted. However, this has caused the service and servicemonitor with old label to still be present on some of the clusters.

Therefore, with the help of this PR the old service and servicemonitor is deleted via a Job. Once it's run, the PR will be reverted after cleanup is done. To track this change, [OSD-15020](https://issues.redhat.com/browse/OSD-15020) jira is created.

_Fixes #_ [OSD-15020](https://issues.redhat.com/browse/OSD-15020) 

- Have created two separate directories with different `config.yaml` since the CronJob group version in OpenShift 4.12 only supports v1 version but in versions older than that, v1beta1 is supported.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-pruning/osd-15020-rpo-412/115-pruning-servicemonitors.CronJob.yaml
+++ b/deploy/sre-pruning/osd-15020-rpo-412/115-pruning-servicemonitors.CronJob.yaml
@@ -4,8 +4,8 @@ metadata:
   name: services-pruner
   namespace: openshift-sre-pruning
 spec:
-  failedJobsHistoryLimit: 5
-  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  successfulJobsHistoryLimit: 1
   concurrencyPolicy: Replace
   schedule: "30 2 * * *"
   jobTemplate:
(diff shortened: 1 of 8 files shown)


deploy/sre-pruning/osd-15020-rpo-412/115-pruning-servicemonitors.CronJob.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: services-pruner
+  namespace: openshift-sre-pruning
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
```
mrbarge, line 8: We probably don't need this many
```suggestion
  failedJobsHistoryLimit: 3
  successfulJobsHistoryLimit: 1
```

mrbarge (commented):
Other than the comments noted, I gave it a test and it seemed to go fine, so in general it looks okay to me

mrbarge:
/lgtm
/hold

We'll unhold when the latest batch of MCC things is promoted.

PR #1457, opened 2023-02-02
OSD-13914 - Fixing the syntax for resource patch for cluster oauth

bdematte:
### What type of PR is this?
Bug

Update syntax for the patching action

[OSD-13914](https://issues.redhat.com//browse/OSD-13914)

Tested the patching on a standard OSD cluster which initially showed similar error when trying to apply the 'patch' 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/rosa-oauth-templates-policies/50-GENERATED-rosa-oauth-templates.Policy.yaml
+++ b/deploy/rosa-oauth-templates-policies/50-GENERATED-rosa-oauth-templates.Policy.yaml
@@ -24,11 +24,15 @@ spec:
                     - complianceType: musthave
                       objectDefinition:
                         apiVersion: config.openshift.io/v1
-                        applyMode: AlwaysApply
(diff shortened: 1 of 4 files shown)


cblecker:
Isn't this a generated resource?
/hold

bdematte:
@cblecker The initial version (to help migrating) was generated with the script, but it has been removed from the generator config (and moved to a dedicated folder) so isn't anymore. 
I will rename it in a dedicated MR to avoid future confusion.

cblecker:
@bdematte Why aren't we generating it? Won't that mean config between stand-alone and HS could drift?

bdematte:
Having a working version now, I think it makes sense. I still have some issues related to oauth (unrelated to this MR) to investigate, but I will ensure we have generators for all resources once I know what is the target for those files.

PR #1461, opened 2023-02-03
OSD-12884 deploy a clusterurlmonitor for each hostedcontrolplane

yithian:
### What type of PR is this?
feature

This PR adds clusterurlmonitors for hostedcontrolplanes so that SRE can monitor API availability of hosted clusters

Fixes [OSD-12884](https://issues.redhat.com//browse/OSD-12884)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


dustman9000:
Is there a reason you did not go through the generated policy path? You will just need to add the proper policy-generator-config.yaml where the original ClusterURLMonitor exists

yithian:
> Is there a reason you did not go through the generated policy path? You will just need to add the proper policy-generator-config.yaml where the original ClusterURLMonitor exists

Mostly because I didn't know about it's existence or how to use it :)

Now that I'm trying it out though, it looks like it generates policies to configure the hosted clusters? However this PR is to monitor hosted clusters' api and so needs to target the management clusters (where the API runs). Is there a way to use the generated policy path to create policies that target the management clusters?

yithian:
Closing this in favor of #1505

PR #1465, opened 2023-02-07
Optimize remote write config and disable metadata traffic

philipgough:
### What type of PR is this?

Enhancement

This PR does two things. Firstly it optimizes the remote write configuration
to send larger batches to the backend which comfortably fall within the ranges
we can support.

This knowledge was gained during our previous load testing https://issues.redhat.com/browse/RHOBS-181
and the change should be favourable from both a client/server perspective

It also disables sending metadata [which Thanos does not currently handle](https://github.com/thanos-io/thanos/blob/main/pkg/receive/handler.go#L514-L523) in remote write and the outcome
is that there is unnecessary traffic on the network which must be processed before being dropped!

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jewzaam:
Monitoring stack for remote writing for HCP is in OSDFM, this is managing CMO on the management cluster and not what needs tuned.  Closing.

PR #1468, opened 2023-02-08
OSD-13913 Update rosa-console-config policies

dakotalongRH:
### What type of PR is this?
Update config policies for rosa-console-branding

https://issues.redhat.com/browse/OSD-13913

https://issues.redhat.com/browse/OSD-13913

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bdematte:
Coupe comments : 
- would be great to rename the file and remove the 'GENERATED' in it + explain in the description the scripting is part of a separated card to secure Milestone delivery
- the 'generate-policy-config.py script needs to be updated to remove 'rosa-console-branding' (as not generated for the time being) 
- you need to run 'make' before submiting the MR to ensure the changes are reflected on `hack/00-osd-managed-cluster-config-<environment>.yaml.tmpl` which are being applied on the different environments

dakotalongRH:
Updated the PR with the suggested changes per @bdematte , thanks!!

jewzaam (changes requested):
/hold

If this works via Policy it should work via SSS.  Has it been tested?  If yes, deploy/rosa-console-branding should be updated from a Patch to the resource as defined in this PR.

deploy/acm-policies/rosa-console-branding.Policy.yaml
```diff
@@ -24,11 +24,19 @@ spec:
                     - complianceType: musthave
                       objectDefinition:
                         apiVersion: operator.openshift.io/v1
-                        applyMode: AlwaysApply
                         kind: Console
-                        name: cluster
-                        patch: '{"spec":{"managementState":"Managed","route":null,"customization":{"brand":null,"documentationBaseURL":"https://docs.openshift.com/rosa/","customProductName":"Red Hat OpenShift Service on AWS","customLogoFile":{"name":"rosa-brand-logo","key":"rosa-brand-logo.svg"}}}}'
-                        patchType: merge
+                        spec:
```
lnguyen1401, line 28: Is this part manually modified after 'make' auto generated the policy? If so i'm afraid the fix would not last, since every time someone runs 'make' in this repo, everything is regenerated again and this fix would be overwritten.
one solution would be that we have a script to generate policies for the 'patch'.  https://issues.redhat.com/browse/OSD-15086

deploy/acm-policies/rosa-console-branding.Policy.yaml
```diff
@@ -24,11 +24,19 @@ spec:
...
+                            route:
+                            customization:
+                                brand:
+                                documentationBaseURL: https://docs.openshift.com/rosa/
+                                customProductName: Red Hat Openshift Service on AWS
+                                customLogoFile:
+                                    name: rosa-brand-logo
+                                    key: rosa-brand-logo.svg
+                        metadata:
+                            name: cluster
```
lnguyen1401, line 39: Is this part manually modified after 'make' auto generated the policy? If so i'm afraid the fix would not last, since every time someone runs 'make' in this repo, everything is regenerated again and this fix would be overwritten.
one solution would be that we have a script to generate policies for the 'patch'. https://issues.redhat.com/browse/OSD-15086

bdematte: The associated folder has been removed from the script hence won't be touched by `make` (the order of action here has been basically generating the files, removing the folder from the script config, and then updating). Or am I missing some details/specific case you are thinking about?

jewzaam:
Prior art now exists in https://github.com/openshift/managed-cluster-config/pull/1474 (thank you!) and we can update to follow that pattern for this configuration?

bdematte:
Submitted replacement MR for this one : https://github.com/openshift/managed-cluster-config/pull/1482

PR #1469, opened 2023-02-08
Update control-plane-operator policy

csrwng:
### What type of PR is this?
feature

The control-plane-opertor needs additional permissions in order to create and destroy a default security group for NodePools where no security group is specified.

_Fixes #_ [HOSTEDCP-789](https://issues.redhat.com//browse/HOSTEDCP-789)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
+++ b/resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
@@ -9,7 +9,15 @@
                 "ec2:ModifyVpcEndpoint",
                 "ec2:DeleteVpcEndpoints",
                 "ec2:CreateTags",
-                "route53:ListHostedZones"
+                "route53:ListHostedZones",
+                "ec2:CreateSecurityGroup",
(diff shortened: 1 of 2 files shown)


fahlmant:
@csrwng Is this specific to hypershift?

csrwng:
@fahlmant yes it is. The entire control plane operator policy is only used in hypershift

jewzaam (changes requested):
This is HS specific, should go in https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json  ?

jharrington22:
@csrwng as @jewzaam suggested could you also add these actions to the [policy](https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json) in the HyperShift directory? The plan is to migrate from these to those newer policies which will contain IAM conditions to ensure we only operate on objects tagged by Red Hat.

jharrington22 (approved):
/lgtm

PR #1471, opened 2023-02-08
[OSD-14739]add acm policy to deploy the webhook patching cronjob to management cluster

bmeng:
### What type of PR is this?
_(feature)_

With deploying the validatingwebhook to the hosted cluster, it needs the cabundle from the management cluster to be injected into the webhook configuration.
As of now, we do not have a mechanism to consume the resources on management cluster for hosted cluster yet.
This will be a temporary solution before we can manage that within the hypershift.

_Fixes #[OSD-14739](https://issues.redhat.com//browse/OSD-14739)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/hypershift-validating-webhook-patching/OWNERS
+++ b/deploy/hypershift-validating-webhook-patching/OWNERS
@@ -1,0 +1,6 @@
+reviewers:
+- bmeng
+- mrbarge
+- ravitri
+- shitaljante
+- a7vicky
--- a/deploy/hypershift-validating-webhook-patching/README.md
+++ b/deploy/hypershift-validating-webhook-patching/README.md
@@ -1,0 +1,38 @@
+# About the validating-webhook-patching-job
+
+This policy used to deploy a cronjob to the hosted control plane namespaces
+on management cluster.
+
+The cronjob is going to patch the validatingWebhookConfiguration custom resources
+which intended to be deployed to the hosted clusters.
+
+As the caBundle is a required field for the validatingwebhook and the validation admission
+happened on the management cluster API server. We need to use the caBundle from the management
+cluster in the hosted cluster validatingwebhook CR.
+
+This is a temporary solution before we can handle the cabundle replacement within the cluster
+deployment by hypershift.
+
+## WebhookConfigurations
+The following webhook configurations will be updated by this job.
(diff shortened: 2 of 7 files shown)


mrbarge:
Couple of minor things before I give this a deeper look:
- Could we add a `hypershift-` prefix to the directory name this is going in, just so it's indicative (`hypershift-validating-webhook-patching`)
- Can we add a brief README.md to the directory describing what this is doing/why, and an OWNERS file (add ourselves and the MCVW approvers as owners)
- Can we link to the corresponding Jira in the PR desc

Some more specific things:
- Execution of once per hour leaves a lot of time for someone to do things on a cluster with no webhook protection. Maybe once every 15 minutes? I don't really know what's a "good" frequency here.
- (nice to have?) I guess it's potentially possible this job could run before the webhooks land. I don't think `patch` has an `--ignore-not-found` option (could be wrong) but it would be nice if it could safely exit without an error if the webhooks it patches do not exist.

bmeng:
Thanks for the review @mrbarge 
regarding to the frequency, yeah I think it is hard to decide, my only concern is that if we run this too frequent, and it may bring extra load to the hypershift mangement cluster (depends on how many hosted control plane being created). I am ok with either 15 minutes or 30 minutes or 1 hour...
for the second part, I added the `|| true` in the end of the oc command, so it should be safe if the patched resource is not there(according to my test, it is just report the webhook does not found and move to the next), see below:

```
$ oc -n ocm-staging-21n1bab01a78f0pgv3g4gh6ctag1f0mh-bmeng-hc logs validating-webhook-patching-job-27929118-lg6qk
Error from server (NotFound): validatingwebhookconfigurations.admissionregistration.k8s.io "bmeng-namespace-validation" not found
Error from server (NotFound): validatingwebhookconfigurations.admissionregistration.k8s.io "bmeng-regular-user-validation" not found
validatingwebhookconfiguration.admissionregistration.k8s.io/bmeng-scc-validation patched
Error from server (NotFound): validatingwebhookconfigurations.admissionregistration.k8s.io "bmeng-techpreviewnoupgrade-validation" not found
```

bmeng:
Updated with the OWNERS and readme. Also the description...

PR #1475, opened 2023-02-09
update the webhook patching cronjob to handle the resource not found issue

bmeng:
### What type of PR is this?
_(bug)_

add the error handle for the cronjob in https://github.com/openshift/managed-cluster-config/pull/1471

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mrbarge:
This is currently failing the PR check. Can you try renaming the yaml file as it's likely hitting this issue: https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/mcc-modify-roleref.md

PR #1476, opened 2023-02-09
OSD-13914 - Migrate patch of oauth for Hypershift support

bdematte:
### What type of PR is this?
Feature

- Changing the applyMode of the rosa-oauth-templates syncset to support behavior similar de 'MustHave' in Policies (ie update of selected fields) to allow to have the resource for Hosted Clusters generated as the other resources.  
- Adding the SSS to the list of generated resources for Policy generate
 
[OSD-13914](https://issues.redhat.com//browse/OSD-13914)

Setting resourceApplyMode to Upsert as in a Patch, we shouldn't delete the resource if we delete the SSS

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/rosa-oauth-templates/config.yaml
+++ b/deploy/rosa-oauth-templates/config.yaml
@@ -4,3 +4,6 @@ selectorSyncSet:
   - key: api.openshift.com/product
     operator: In
     values: ["rosa"]
+  applyBehavior: "CreateOrUpdate"
+  # use Upsert so if this configuration no longer applies it will not delete the OAuth resource in cluster which is a singleton.  We should never delete OAuth.
+  resourceApplyMode: "Upsert"
--- a/deploy/rosa-oauth-templates/rosa-oauth-templates.patch.yaml
+++ b/deploy/rosa-oauth-templates/rosa-oauth-templates.patch.yaml
@@ -1,7 +1,0 @@
-apiVersion: config.openshift.io/v1
-applyMode: AlwaysApply
-kind: OAuth
-name: cluster
-patch: '{"spec":{"templates": {"login": {"name": "rosa-oauth-templates-login"},"providerSelection":
-  {"name": "rosa-oauth-templates-providers"},"error": {"name": "rosa-oauth-templates-errors"}}}}'
-patchType: merge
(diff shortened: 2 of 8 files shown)


deploy/rosa-oauth-templates/rosa-oauth-templates.patch.yaml
```diff
@@ -1,7 +1,12 @@
 apiVersion: config.openshift.io/v1
```
jewzaam, line 1: Rename file: deploy/rosa-oauth-templates/rosa-oauth-templates.OAuth.yaml

deploy/rosa-oauth-templates/config.yaml
```diff
@@ -3,4 +3,6 @@ selectorSyncSet:
   matchExpressions:
   - key: api.openshift.com/product
     operator: In
-    values: ["rosa"]
\ No newline at end of file
+    values: ["rosa"]
+  applyBehavior: "CreateOrUpdate"
+  resourceApplyMode: "Upsert"
```
jewzaam, line 8:
```suggestion
  # use Upsert so if this configuration no longer applies it will not delete the OAuth resource in cluster which is a singleton.  We should never delete OAuth.
  resourceApplyMode: "Upsert"
```

jewzaam (changes requested):
Was this tested on a cluster that has IDP configuration managed through hive?  I want to confirm we know that a resync by hive when one or more SelectorSyncIdentityProviders exists will result in expected configuration, i.e. both the identity provider config and the templates are applied.

bdematte:
Validated on staging the IDP specific setup didn't get overriden when updating oauth with this syntax.

jewzaam (approved):

lnguyen1401:
Tested the policy generated for this patch. working as expected
`21gj0p49puto4v8km55ehm80fkv2tvtc    openshift-acm-policies.rosa-oauth-templates                                           Compliant          3m18s
`

`Event Name:      openshift-acm-policies.rosa-oauth-templates.174259be4aa93d12
      Last Timestamp:  2023-02-10T03:44:45Z
      Message:         Compliant; notification - oauths [cluster] found as specified, therefore this Object template is compliant
    Template Meta:
      Creation Timestamp:  <nil>
      Name:                rosa-oauth-templates
Events:
`

PR #1479, opened 2023-02-10
Refine installer policy for hyperShift hosted clusters

karthikperu7:
### What type of PR is this?
_feature_

This PR refines the installer role for hosted clusters by removing actions that are not required and adding condition keys to the actions that are applicable

_Fixes #_  https://issues.redhat.com/browse/OSD-14611

This is the part of the work to review and refine all hyperShift role policies and make sure they follow the principle of least privilege with condition keys set, where applicable.

Hosted cluster creation, deletion and normal function (like operators working ) has been tested on both prod and stage environments with this installer policy.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.12/hypershift/InstallerPolicy.json
+++ b/resources/sts/4.12/hypershift/InstallerPolicy.json
@@ -4,70 +4,136 @@
         {
             "Effect": "Allow",
             "Action": [
-                "ec2:CreateSubnet",
-                "ec2:CreateVolume",
-                "ec2:CreateVpc",
-                "ec2:DeleteSubnet",
-                "ec2:DeleteVolume",
-                "ec2:DeleteVpc",
-                "ec2:DescribeAddresses",
+                "ec2:CreateTags",
+                "ec2:DeleteTags",
                 "ec2:DescribeAvailabilityZones",
                 "ec2:DescribeInternetGateways",
-                "ec2:DescribeNetworkInterfaces",
                 "ec2:DescribeRegions",
                 "ec2:DescribeReservedInstancesOfferings",
                 "ec2:DescribeRouteTables",
-                "ec2:DescribeSecurityGroups",
                 "ec2:DescribeSecurityGroupRules",
                 "ec2:DescribeSubnets",
                 "ec2:DescribeVpcAttribute",
                 "ec2:DescribeVpcs",
-                "ec2:RunInstances",
-                "elasticloadbalancing:CreateLoadBalancer",
+                "ec2:DescribeInstanceTypeOfferings",
                 "elasticloadbalancing:DeleteLoadBalancer",
-                "iam:AddRoleToInstanceProfile",
-                "iam:AttachRolePolicy",
-                "iam:CreateInstanceProfile",
-                "iam:CreateOpenIDConnectProvider",
-                "iam:CreatePolicy",
-                "iam:CreatePolicyVersion",
-                "iam:CreateRole",
-                "iam:DeleteOpenIDConnectProvider",
-                "iam:DeletePolicyVersion",
-                "iam:DeleteRole",
-                "iam:DeleteRolePermissionsBoundary",
-                "iam:DetachRolePolicy",
-                "iam:GetOpenIDConnectProvider",
-                "iam:GetPolicy",
-                "iam:GetRole",
-                "iam:GetUser",
-                "iam:ListAttachedRolePolicies",
-                "iam:ListPolicyTags",
-                "iam:ListPolicyVersions",
-                "iam:ListRoleTags",
-                "iam:PutRolePermissionsBoundary",
-                "iam:TagInstanceProfile",
-                "iam:TagPolicy",
                 "kms:DescribeKey",
                 "route53:ChangeResourceRecordSets",
                 "route53:DeleteHostedZone",
                 "route53:GetAccountLimit",
                 "route53:GetHostedZone",
                 "route53:ListHostedZones",
-                "route53:ListHostedZonesByName",
                 "route53:ListResourceRecordSets",
-                "s3:CreateBucket",
-                "s3:DeleteBucket",
-                "s3:DeleteObject",
-                "s3:ListObjects",
-                "s3:PutObject",
-                "secretsmanager:GetSecretValue",
                 "servicequotas:GetServiceQuota",
-                "sts:AssumeRole",
-                "sts:AssumeRoleWithWebIdentity",
-                "sts:GetCallerIdentity"
+                "sts:GetCallerIdentity",
+                "iam:TagInstanceProfile"
             ],
-        "Resource": "*"
+            "Resource": "*"
+        },
+        {
+            "Action": [
+                "iam:PassRole"
+            ],
+            "Resource": [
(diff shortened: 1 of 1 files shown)


resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,67 +4,130 @@
...
+              "route53:ChangeResourceRecordSets"
+            ],
+            "Resource": "*",
+            "Condition": {
+              "ForAllValues:StringLike": {
+                "route53:ChangeResourceRecordSetsNormalizedRecordNames": [
+                  "*.openshiftapps.com",
+                  "*.devshift.org",
+                  "*.hypershift.local"
+                ]
```
jharrington22, line 118: @karthikperu7 what about fedRAMP?

resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,67 +4,130 @@
...
+        },
+        {
+            "Action": [
+                "iam:PassRole"
+            ],
+            "Resource": [
+                "arn:*:iam::*:role/*-Worker-Role"
+            ],
+            "Effect": "Allow"
+        },
```
jharrington22, line 36:
```suggestion
        {
            "Action": [
                "iam:PassRole"
            ],
            "Resource": [
                "arn:*:iam::*:role/*-Worker-Role"
            ],
            "Effect": "Allow",
            "Condition": {
                "StringEquals": {
                    "iam:PassedToService": [
                        "ec2.amazonaws.com"
                    ]
                }
            }
        },
```

Further restriction ensuring the role can only passed the AWS EC2 service.

resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,67 +4,130 @@
...
+        {
+            "Sid": "BucketActions",
+            "Effect": "Allow",
+            "Action": [
+                "s3:CreateBucket",
+                "s3:DeleteBucket"
+            ],
+            "Resource": [
+                "arn:aws:s3:::*-oidc"
```
jharrington22, line 72: Does HyperShift create a bucket in the customers account? I think we're moving towards BYO OIDC @arendej ?

resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,67 +4,130 @@
...
+            ]
+        },
+        {
+            "Sid": "CreateAndManageInstanceProfiles",
+            "Effect": "Allow",
+            "Action": [
+                "iam:AddRoleToInstanceProfile",
```
jharrington22, line 79: We should be able to restrict these by tag, did you try it? If not we'll need a card for the HyperShift team

resources/sts/4.12/hypershift/InstallerPolicy.json
arendej: Hypershift is only BYO OIDC.
cc @wgordon17

resources/sts/4.12/hypershift/InstallerPolicy.json
karthikperu7: I think we can restrict the `CreateInstanceProfile` action using condition keys for aws tags, however other actions do not seem to support condition keys as per [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsidentityandaccessmanagement.html) ?

In any case, the instance profiles are not being tagged with `red-hat-managed=true` at the moment :
```
$ aws iam list-instance-profile-tags --instance-profile-name staging-ID-cluster-worker
{
    "Tags": [],
    "IsTruncated": false
}
```

I have created https://issues.redhat.com/browse/SDA-8142 to get this fixed and a follow up for us to edit this policy once the tags are set (OSD-15178) so that it doesnt block this PR.

Thanks, this is a good spot.

resources/sts/4.12/hypershift/InstallerPolicy.json
karthikperu7: I wasnt sure if we should consider fedramp for these roles. I have added the fedramp records now though.

resources/sts/4.12/hypershift/InstallerPolicy.json
karthikperu7: Thanks fixed

resources/sts/4.12/hypershift/InstallerPolicy.json
karthikperu7: if hyperShift installs are going to enforce BYO-OIDC, then this permission would not be needed. I tried an install with BYO-OIDC configuration withouth these permissions and the cluster installed successfully.

I have removed these permission from the installer role now, please let us know if BYO-OIDC will not be enforced, in which case we will need to add these permissions back.

karthikperu7:
I have rebased my branch and made sure the following permissions are present based on the recent PRs on the installer policy file:
```
ec2:describesecuritygrouprules
iam:taginstanceprofile
route53:GetAccountLimit
```

resources/sts/4.12/hypershift/InstallerPolicy.json
wgordon17: @gdbranco could you share the adjusted BYO-OIDC flow? It's still possible that we might need S3 permissions, even with BYO-OIDC. It might not be _required_, since in BYO-OIDC customers can bring _whatever_, but I thought the default still relied on S3 ¯\_(ツ)_/¯

resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,70 +4,146 @@
...
+        },
+        {
+            "Action": [
+                "iam:PassRole"
+            ],
+            "Resource": [
+                "arn:*:iam::*:role/*-Worker-Role"
+            ],
+            "Effect": "Allow",
+            "Condition": {
```
fahlmant, line 39: It is possible to also add a condition here that the role has the red-hat-managed: true tag? I'm not sure if you need a new SID for that

resources/sts/4.12/hypershift/InstallerPolicy.json
```diff
@@ -4,70 +4,146 @@
...
+                "StringEquals": {
+                  "aws:RequestTag/red-hat-managed": "true"
+                }
+              }
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "secretsmanager:GetSecretValue"
```
fahlmant, line 107: What is this for?

gdbranco: This is for BYO OIDC flow, the private key will be stored in secrets manager in client's aws account

resources/sts/4.12/hypershift/InstallerPolicy.json
karthikperu7: The `PassRolw` doesn't support tag based condition keys according to the [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsidentityandaccessmanagement.html)

fahlmant (approved):
/lgtm
/approve

PR #1485, opened 2023-02-13
OSD-15177 switch to mustonlyhave complianceType by default

mrbarge:
### What type of PR is this?
Bug

Switch to a `mustonlyhave` policy compliance type by default.

Using `musthave` policy type can result in lingering data being left in target resources if those resources are changed over time in the Policy. For RBAC, this can lead to leftover permissions left in Roles/ClusterRoles that are no longer present in the policy itself.

[OSD-15177](https://issues.redhat.com//browse/OSD-15177)


mrbarge:
/hold
Just so this gets sanity-checked first. But feel free to unhold if it is sane!

PR #1488, opened 2023-02-13
Fix manifests with syntax errors that cause generated polices to be noncompliant

mprahl:
### What type of PR is this?
cleanup

These commits fix syntax errors in the manifests used to generate policies. This will significantly reduce the churn on the Hub cluster since the current behavior of ACM policies is to send a noncompliant status update every time an invalid object manifest is encountered.

https://issues.redhat.com/browse/OSD-15257

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/elevated-sre/20-cluster-admin.User.yml
+++ b/deploy/backplane/elevated-sre/20-cluster-admin.User.yml
@@ -2,3 +2,4 @@ apiVersion: user.openshift.io/v1
 kind: User
 metadata:
   name: backplane-cluster-admin
+groups: null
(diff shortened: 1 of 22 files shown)


deploy/backplane/elevated-sre/20-cluster-admin.User.yml
```diff
@@ -1,4 +1,5 @@
 apiVersion: user.openshift.io/v1
 kind: User
 metadata:
-  name: backplane-cluster-admin
\ No newline at end of file
+  name: backplane-cluster-admin
+groups: []
```
jewzaam, line 5:
```suggestion
groups: null
```
When a User is created w/o `groups` the value defaults to `null`, suggest we keep consistency with that default.

jewzaam (changes requested):

jewzaam (approved):
/approve

I think the PR check is failing, make sure `make` is run?

PR #1472, opened 2023-02-08
Add iam:TagInstanceProfile permission to Installer role

andreadecorte:
OCM creates InstanceProfile resources but so far they were not tagged. 

### What type of PR is this?
Feature

### What this PR does / why we need it?
This adds the necessary permission to allow tagging and align with all the other AWS resources.

### Which Jira/Github issue(s) this PR fixes?
Related: [SDA-7934](https://issues.redhat.com//browse/SDA-7934)

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1489, opened 2023-02-13
Add tag instance profile to installer and core installer policies

jharrington22:
### What type of PR is this?
cleanup

Adding IAM actions from #1472 to the installer core and installer policy. They should be in all until we have a path to install HyperShift only policies [SDA-8155](https://issues.redhat.com/browse/SDA-8155)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.12/sts_installer_core_permission_policy.json
+++ b/resources/sts/4.12/sts_installer_core_permission_policy.json
@@ -92,6 +92,7 @@
 		    "iam:CreateInstanceProfile",
 		    "iam:DeleteInstanceProfile",
 		    "iam:GetInstanceProfile",
+		    "iam:TagInstanceProfile",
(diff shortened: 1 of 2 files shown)


resources/sts/4.12/sts_installer_core_permission_policy.json
```diff
@@ -92,6 +92,7 @@
 		    "iam:CreateInstanceProfile",
 		    "iam:DeleteInstanceProfile",
 		    "iam:GetInstanceProfile",
+            "iam:TagInstanceProfile",
```
fahlmant, line 95: Formatting looks a little off here

fahlmant (approved):
/lgtm
/approve

PR #1490, opened 2023-02-13
deploy a routemonitor for each hostedcontrolplane

anispate:
As RMO watches all namespaces for routeMonitors. Currently, we're using this to monitor the cluster console route in classic OSD, and therefore will need the same functionality on Hypershift. This is related to [OSD-14588](https://issues.redhat.com/browse/OSD-14588) ticket

### DIFF
--- a/deploy/acm-policies/50-GENERATED-hypershift-route-monitor-operator.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-hypershift-route-monitor-operator.Policy.yaml
@@ -1,0 +1,76 @@
+---
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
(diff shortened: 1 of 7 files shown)


yithian:
The files in `hack/` don't look right to me, they shouldn't have `namespaceSelector`s, for one thing.

Can you try running `make` and seeing if there are any changes to them?

yithian (changes requested):
The files in `hack/` don't look right to me, they shouldn't have `namespaceSelector`s, for one thing.

Can you try running `make` and seeing if there are any changes to them?

yithian (changes requested):
don't forget to add the `hypershift-route-monitor-operator` directory to `scripts/generate-policy-config.py`

PR #1493, opened 2023-02-14
Add route53:GetAccountLimit permission for HS preflight check

tbrisker:
### What type of PR is this?
_feature_

In order to check if user has enough quota for creating hosted zones, we need this permission.

_Required for SDA-7726_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jharrington22 (approved):
/lgtm

PR #1494, opened 2023-02-14
'roleRef' PR check excludes 'namespace'

jewzaam:
### What type of PR is this?
cleanup

The 'namespace' property we have added in many roleRef is not a property in the specification of RoleBinding.  ACM Policy validation fails because of this.  Therefore we need to remove this invalid and ineffectual property (it doesn't get persisted in cluster at all).  This change permits the PR check for roleRef changes to ignore changes to .roleRef.namespace.

Related to https://issues.redhat.com/browse/SDE-2633 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/lgtm
/approve

PR #1496, opened 2023-02-14
OSD-15258 Fix policy non-compliant issues with missing resources

jewzaam:
### What type of PR is this?
bug

Remove some policies that will never be compliant because the RBAC references resources that will never exist.

https://issues.redhat.com/browse/OSD-15258

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


cblecker:
Less lines of code? I'm in.
/lgtm
/approve

PR #1508, opened 2023-02-21
[OSD-15092] Refine IAM policy files for HyperShift specific operators

karthikperu7:
### What type of PR is this?
_feature_

This PR refines the policy files for HyperShift specific operators by removing the actions that are not used by the operator and adding condition keys to restrict them further. The operators files that are being modified are for the following operators:
* Control plane operator
* Node pool management operator
* kube cloud controller operator

_Fixes #_ https://issues.redhat.com/browse/OSD-15092


resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
```diff
@@ -2,32 +2,89 @@
...
-                "route53:ListHostedZones",
-                "ec2:CreateSecurityGroup",
-                "ec2:AuthorizeSecurityGroupIngress",
-                "ec2:AuthorizeSecurityGroupEgress",
-                "ec2:DeleteSecurityGroup",
-                "ec2:RevokeSecurityGroupIngress",
-                "ec2:RevokeSecurityGroupEgress",
-                "ec2:DescribeSecurityGroups",
-                "ec2:DescribeVpcs"
+                "route53:ListHostedZones"
```
fahlmant, line 7: Can the condition for the list of openshiftapps/devshift urls be added here? 
Also, should this be spun out into two different SIDs? That way the VPC Endpoint describe can have the condition of red-hat-managed: true

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
```diff
@@ -2,32 +2,89 @@
...
         },
         {
-            "Effect": "Allow",
             "Action": [
                 "route53:ChangeResourceRecordSets",
                 "route53:ListResourceRecordSets"
             ],
-            "Resource": "arn:aws:route53:::%s"
+            "Effect": "Allow",
+            "Resource": [
```
fahlmant, line 18: Same thought here, can these be limited to the hosted zones with the right URLs?

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
```diff
@@ -2,32 +2,89 @@
...
+                "ec2:CreateVpcEndpoint"
+            ],
+            "Resource": [
+                "arn:aws:ec2:*:*:vpc/*",
+                "arn:aws:ec2:*:*:subnet/*",
+                "arn:aws:ec2:*:*:route-table/*"
+            ]
+        },
+        {
+            "Sid": "ManageVPCEndpoingWithCondition",
```
fahlmant, line 50:
```suggestion
            "Sid": "ManageVPCEndpointWithCondition",
```

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
karthikperu7: It doesnt look like the condition key to list specific regex-ed URLs for `ListHostedZones` is supported as per the [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonroute53.html)

I tried adding the red-hat-managed tag condition key for `DescribeVPCEndpoints` but that throws `Client.UnauthorizedOperation` error for the control plane operator when it tries to perform that action. I tried both ResourceTag/RequestTag condition keys. These condition keys don't appear for `DescribeVPCEndpoints` in the [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html) which makes me think this may not work...

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
karthikperu7: good point, going to test this before committing. Thanks

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
karthikperu7: will fix, thanks

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
karthikperu7: this has been tested and fixed

resources/sts/4.12/hypershift/ControlPlaneOperatorPolicy.json
karthikperu7: fixed

fahlmant (approved):
/lgtm
/approve

PR #1512, opened 2023-02-28
Download PolicyGenerator from upstream

mrbarge:
### What type of PR is this?
Bug

The existing release download location for the `PolicyGenerator` now 404s and doesn't exist.

This moves the download URL back to the upstream (in lieu of any better ideas of where to point it?)


mrbarge:
Reviewers: Feel free to remove the hold if this makes sense, I'm just without any context if there was a particular reason to point at the solostron releases (which no longer exist, for reasons I don't know)

jewzaam:
Confirmed with ACM this PR is updating to the right place for the binary.

/approve

PR #1518, opened 2023-03-03
MTSRE-1031 | fix: addon-operator backplane permissions to span across all clusters with addon-operator

yashvardhan-kukreja:
### What type of PR is this?
_(fix)_

At the moment, the addon-operator oriented backplane permissions don't propagate to all the clusters out there. They only propagate to the clusters with one or more of [these]() addons. 

But that is wrong and we expect the addon-operator permissions to exist in all the clusters with addon-operator. All the clusters with addon-operator are identified by `api.openshift.com/managed: true` in their ClusterDeployments.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/lpsre/config.yaml
+++ b/deploy/backplane/lpsre/config.yaml
@@ -1,6 +1,10 @@
 deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   matchLabels:
+      # the following label selector matches every OSD cluster with addon-operator i.e. every cluster out there
+      api.openshift.com/managed: "true"
+
+      # the following lines are going to be removed after the `api.openshift.com/managed: "true"` is validated
       api.openshift.com/addon-rhmi-operator: "true"
       api.openshift.com/addon-rhmi-operator-internal: "true"
       api.openshift.com/addon-managed-api-service: "true"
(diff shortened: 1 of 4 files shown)


yashvardhan-kukreja:
> This is a broad change from what we do today and would permit lpsre access to all managed openshift clusters. This is not my expectation. Please provide link to JIRA where this is tracked so we can record requirements and decisions around this change.

@jewzaam The change is definitely warranted for because "addon-operator" is essentially a "product" which we are serving in every cluster out there through the means of `api.openshift.com/managed: true` because it's an OSD operator. And it being our product and us being the SREs, we need to have just the right amount of permission in `openshift-addon-operator`  namespace including the backplane-readers-admin permission as well. 

Before this PR, that wasn't the case, and we were getting these permissions on clusters where only certain addons were installed but that was an incorrect behaviour. Addon-operator being served by us a product should be accessible to us wherever and whenever irrespective of whether some addon is installed in that cluster or not.

jewzaam:
Blocked until SD-ADR-0123 is accepted.  Note added to MTSRE-1031

yashvardhan-kukreja:
Hi @jewzaam , I am personally geared away from the idea of getting blocked on the above ADR for getting this PR approved especially considering it's criticality, appropriateness and urgency at the moment. 

I provided more details around its justification in the latest comment in this ticket - https://issues.redhat.com/browse/MTSRE-1031

Please take a look and I'd love to hear out your thoughts.

Thank you!

yashvardhan-kukreja:
Closed in favor of some recent conversations and more clarity around the scope of LP-SRE and SREP.

More details under the comments of the associated Jira ticket.

PR #1533, opened 2023-03-16
Disable CPMS in all 4.12 clusters

karthikperu7:
### What type of PR is this?
cleanup

Disabling CPMS in all 4.12 clusters in the managed fleet until https://issues.redhat.com/browse/OCPBUGS-8349 is resolved and ADR-0112 is accepted

https://issues.redhat.com/browse/OSD-15560

I want to get some more opinions before merging this to make sure this is the right thing to do. 

/hold

### DIFF
--- a/deploy/osd-14634-disable-cpms/config.yaml
+++ b/deploy/osd-14634-disable-cpms/config.yaml
@@ -4,9 +4,6 @@ selectorSyncSet:
   matchExpressions:
   ## This will need to be changed to use version-major-minor-patch when
   ## we move to enable this feature in the future
-  - key: hive.openshift.io/version-major-minor-patch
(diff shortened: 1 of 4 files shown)


iamkirkbater:
I think this should be fine to merge and deploy.  However, I don't think this will catch _all_ of the issues we've been seeing, and could potentially cause other issues.

Potential issue it _could_ create - CPMS is in the middle of rolling a node while MCC config is being pushed down.  In the middle of rolling the node, this cronjob disables CPMS.  The node rollout stops, leaving a 4th master provisioned on that cluster.

karthikperu7:
Going to close this PR as we believe that this is not going to prevent from the bug from occurring as it happens before syncsets land.

PR #1555, opened 2023-03-24
Proof of Concept of split stage/int/prod components

iamkirkbater:
This is a proof of concept that aims to alleviate our current issues with promoting to prod by splitting each component into it's own generated set of syncset files, thereby allowing individual components to promote themselves to prod.

This acts as a way to buy us time to investigate longer-term solutions like ADR-0096 and its relative ADR-0103 without continuing to feel pain every time we want to deploy in the meantime.

This is not _directly_ meant to be a catch-all solution that will fix all of our problems or be any kind of super robust system, this is a step in that direction. This is entirely meant to allow us to not bottleneck important prod promotions because of other components, giving us the ability to maintain velocity while we search for the best and more robust solution, even if that is just another iteration on this.

Because the `on push: make` target runs, you'll want to view just the first commit for a better idea of what the diff here would be as that's "hand crafted" to be more accurate to the expected functionality: https://github.com/openshift/managed-cluster-config/pull/1555/commits/1223927fd41bd7e6bd5c9b3183c42c1d3cd85858

Here's an example of how the flow would work as multiple components are updated.
image

One risk that was called out is that if a component relies on another component it might work in staging but if both components aren't promoted it could cause issues, but I think that risk is negligible because it's not really any different than any other kind of standard architecture where multiple moving pieces need to be pushed at the same time to deploy correctly.


iamkirkbater:
/hold

Please do not merge this.

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -3006,7 +3006,6 @@ objects:
                   metadata:
                     name: api
                   spec:
-                    domainRef: hcp
```
iamkirkbater: not sure why this is here.  ignore this line

hack/00-osd-managed-cluster-config-stage.yaml.tmpl
```diff
@@ -3006,7 +3006,6 @@ objects:
                   metadata:
                     name: api
                   spec:
-                    domainRef: hcp
```
iamkirkbater: not sure why this is here.  Ignore this.

PR #1567, opened 2023-03-30
Remove redundant CreateInstanceProfile action from installer policy

karthikperu7:
### What type of PR is this?
bug

Removing redundant `CreateInstanceProfile` policy from the installer policy file. It already exists in its own SID block with condition keys set.

_Fixes #_

I tested this on a test hosted cluster which installed ok without any issues


karthikperu7:
Will create a different PR with more changes
/close

PR #1549, opened 2023-03-22
Add openshift-customer-monitoring support to HCP

jewzaam:
### What type of PR is this?
bug

### What this PR does / why we need it?
Gap with HCP config, openshift-customer-monitoring is currently  deployed fleet-wide today for OSD/ROSA.

### Which Jira/Github issue(s) this PR fixes?
https://issues.redhat.com/browse/OSD-15636

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1568, opened 2023-03-30
Add openshift-logging namespace support to HCP

janboll:
### What type of PR is this?
bug

The openshift-logging namespace is missing on HCP as of now. Adding it as this is a GAP towards OSD.

_Fixes #_ https://issues.redhat.com/browse/OSD-15783

Follows up on other PRs:

- https://github.com/openshift/managed-cluster-config/pull/1549
- https://github.com/openshift/managed-cluster-config/pull/1558

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/scripts/generate-policy-config.py
+++ b/scripts/generate-policy-config.py
@@ -30,6 +30,7 @@ directories = [
         'osd-delete-backplane-script-resources',
         'osd-delete-backplane-serviceaccounts',
         'osd-backplane-managed-scripts',
+        'osd-logging/unsupported',
         'osd-must-gather-operator',
         'osd-openshift-operators-redhat',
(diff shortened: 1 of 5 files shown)


scripts/generate-policy-config.py
```diff
@@ -30,6 +30,7 @@
         'osd-delete-backplane-script-resources',
         'osd-delete-backplane-serviceaccounts',
         'osd-backplane-managed-scripts',
+        'osd-logging',
```
jewzaam, line 33: The root directory only for extended logging support.  We will never bring this into HCP.
What needs to be included is `deploy/osd-logging/unsupported`

PR #1569, opened 2023-03-30
Remove redundant CreateInstanceProfile action and add missing describeSecurityGroups and describeLoadbalancers action to installer role

karthikperu7:
### What type of PR is this?
bug

Removing redundant `CreateInstanceProfile` action which already exists in it own SID with condition keys set
Adding `DescribeLoadBalancers` and `DescribeSecurityGroups` as it is needed by the installer during installation.

This was tested on a hosted cluster installation and deletion which worked ok.


fahlmant (approved):
/lgtm
/approve

PR #1575, opened 2023-04-03
Fix hypershift support role and update support role for 4.13

karthikperu7:
### What type of PR is this?
bug

There are two parts:
1. The `RunInstances` Permission for HyperShift support role had an invalid condition key: `ResourceTag/red-hat-managed` when it should be `RequestTag/red-hat-managed` . This PR fixes this. It has been tested by running network verifier on a test hosted cluster. Network verifier wasn't working earlier using support role due to the invalid condition key
2. The second change is updating the `4.13` hyperShift support policy to match the updated support role in `4.12` directory.


karthikperu7:
cc:  @fahlmant @jharrington22

fahlmant:
The support role should match this: https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.12/reference_policies/ROSASRESupportPolicy_unabridged.json

jharrington22 (commented):
@karthikperu7 the other change here is that spacing? Can that be a separate commit?

resources/sts/4.12/hypershift/sts_hcp_support_permission_policy.json
```diff
@@ -78,7 +78,7 @@
             ],
             "Condition": {
                 "StringEquals": {
-                    "aws:ResourceTag/red-hat-managed": "true"
+                    "aws:RequestTag/red-hat-managed": "true"
```
jharrington22, line 81: @fahlmant I think we already fixed this somewhere?

karthikperu7:
> @karthikperu7 the other change here is that spacing? Can that be a separate commit?

@jharrington22 I have removed the second change from this PR, I can create a new PR once this is merged to update the support role in `4.13/sts/hypershift` dir

fahlmant (approved):
/lgtm
/approve

PR #1577, opened 2023-04-04
Remove ReleaseAddress from HCP NodePool policy as it's not used

fahlmant:
ReleaseAddress is used for releasing an Elastic IP, which is not applicable to HCP.


jharrington22:
@fahlmant considering there is no assign action I'm just going to merge this while we wait for upstream. https://github.com/openshift/hypershift/pull/2379

PR #1581, opened 2023-04-05
OSD-15881: fix resources causing non-compliant logging policies

jewzaam:
### What type of PR is this?
bug

Remove the `serviceAccount` properties from OperatorGroup as it isn't a valid property and is causing ACM policy to be non-compliant.

https://issues.redhat.com/browse/OSD-15881

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/lgtm
/approve

PR #1583, opened 2023-04-05
Expand SREP read rbac a bit for MC and SC

jewzaam:
### What type of PR is this?
cleanup

Give SRE more read permissions on SC and MC specifically so we do not need to break glass to query for some key resources at a cluster scope (aka with `oc -A`).

None

Born from frustration having to break glass to inspect acm policy, hypershift, and network policy state across a MC or SC.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (commented):
/lgtm

PR #1588, opened 2023-04-06
Adding a missing comma in the HCP installer policy

fahlmant:
Json is fun


jewzaam (changes requested):
Can we spend a moment to add a PR check to validate the JSON is syntatically correct?

jharrington22:
@jewzaam I have some validation for this and to ensure policies in each Y are additive only. It will go in this week, we were just moving fast yesterday.

jharrington22 (approved):
/lgtm

PR #1589, opened 2023-04-06
Remove tagging permissions from the HCP installer policy

jharrington22:
### What type of PR is this?
feature

This PR removes the tagging permissions from the HCP installer that were supporting validation in cluster service. The validation checked and then ensured public subnets passed in from customers were tagged. Such permissions are very privileged so the validation has been removed in [SDA-8759](https://issues.redhat.com/browse/SDA-8759).

A preflight check will take its place to warn customers that their subnet is not tagged.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/LGTM
/Approve

PR #1591, opened 2023-04-07
Adding a config.yaml for every SSS

bdematte:
### What type of PR is this?
Feature preparation

As a preparation for [OSD-15267](https://issues.redhat.com//browse/OSD-15267) (to make it easier to review), adding explicit default config.yaml where it is currently implicit (with the objective of making it mandatory in [OSD-15267](https://issues.redhat.com//browse/OSD-15267)).

[OSD-15553](https://issues.redhat.com//browse/OSD-15553)

Templates aren't modified by the MR which is validating it is doing its expected duty. 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,13 @@ All resources in `deploy/` are bundled into a template that is used by config ma
 
 Direct deployment (#1) supports resources that are not synced down to OSD clusters.  SelectorSyncSet deployment (#2) supports resoures that _are_ synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
 
-* This file will become mandatory in the near future in the scope of OSD-15267 and have been added to all folders as a first step. If not present it's assumed `deploymentMode` is `"SelectorSyncSet"` with no additional configuration.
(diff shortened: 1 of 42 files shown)


jewzaam (changes requested):
Add enforcement in PR check, when `./hack/app_sre_pr_check.sh` is run if a `config.yaml` is missing that should exist it will fail with a user friendly message.

Alternative (which I don't particularlly like) is have `make` generate the "default" config.yaml.  Would want it to be a new make target and be added to the list of default targets.

bdematte:
@jewzaam the idea was to enforce it within [OSD-15267](https://issues.redhat.com//browse/OSD-15267) in order to bundle the changes of expectations/process in 1 step for contributors and make it clearer (with proper mail communication to be sent upon merge of MR for OSD-15267 )

jewzaam:
@bdematte since this is adding the config in every dir where it is needed you're better off putting in the enforcement here.  I guess alternative is to add that plus any drift (new config.yaml needed) later, but I don't see why not just do it now.

jewzaam (approved):

jewzaam:
Failure when config is missing works great for SSS generation.
Verified it fails with a useful message for pr check as well.
And with all the changes, this is a noop for the hive templates.

/approve
/lgtm

PR #1595, opened 2023-04-10
Add HyperShift monitoring label to openshift-monitoring namespace

supreeth7:
### What type of PR is this?
Feature

With HyperShift Management Clusters metrics to be forwarded to RHOBS via the Observability Operator, we need a common label that acts as an identifier that the namespace should be watched by the operator.
This PR adds the `hypershift.openshift.io/monitoring` label to the `openshift-monitoring` namespace on HyperShift Management to achieve this.

_**Note:** Even though this configuration targets Service Clusters, patching the label on the Service Clusters won't make a difference as Service Clusters will not have `rhobs/v1` service monitors in the `openshift-monitoring` namespace. Thus OBO won't scrape any metrics from the namespace._

Resolves: [OSD-15906](https://issues.redhat.com/browse/OSD-15906)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/hypershift-namespace-labels/02-openshift-monitoring.patch.yaml
+++ b/deploy/hypershift-namespace-labels/02-openshift-monitoring.patch.yaml
@@ -1,0 +1,7 @@
+kind: Namespace
+apiVersion: v1
+name: openshift-monitoring
+applyMode: AlwaysApply
+patch: |-
+  { "metadata": {"labels": {"hypershift.openshift.io/monitoring":"true"} } }
+patchType: merge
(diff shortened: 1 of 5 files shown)


wanghaoran1988:
1. Could we only apply this label to Management and Service clusters?
2. Could you also add a Readme?

supreeth7:
> Could we only apply this label to Management and Service clusters?

This is already being done: https://github.com/openshift/managed-cluster-config/blob/master/deploy/hypershift-namespace-labels/config.yaml.

> Could you also add a Readme?

ReadMe added

PR #1596, opened 2023-04-10
Add OSD Fleet Manager backplane access namespace and permissions

tiwillia:
### What type of PR is this?
_feature_

This adds a namespace and subjectPermission for the OSD Fleet Manager backplane access

Related to https://issues.redhat.com/browse/SDA-8247

Followed the backplane team on-boarding documentation from SRE-P.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/backplane/osdfm/20-osdfm.SubjectPermission.yml
```diff
@@ -0,0 +1,14 @@
...
+kind: SubjectPermission
+metadata:
+  name: backplane-osdfm
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+  - backplane-readers-cluster
+  permissions:
+  - clusterRoleName: dedicated-readers
+    namespacesAllowedRegex: "(^kube$|^kube-.*|^openshift$|^openshift-.*|^default$|^redhat-.*)"
```
jewzaam, line 11:
```suggestion
    namespacesAllowedRegex: "(^kube-.*|^openshift$|^openshift-.*|^default$|^redhat-.*)"
```

`kube` NS doesn't actually exist, guessing that was carried from another place.

jewzaam (changes requested):
https://issues.redhat.com/browse/SDA-8247 is closed as "won't do", is there another card tracking this change?

jewzaam:
oh, and need to run `make` to generate stuff

iamkirkbater:
Since this is specific to backplane on Hypershift managment/service clusters, would we also want to limit this role's deployment to hypershift management/service clusters via a config.yaml?  I'd be happy to help create one for that, just want to know if it's necessary?

jewzaam:
Good catch, yes this should be on SC and MC only.  Probably this config.yaml would do it.

```yaml
selectorSyncSet:
  matchExpressions:
  - key: ext-hypershift.openshift.io/cluster-type
    operator: In
    values: ["management-cluster","service-cluster"]
    operator: NotIn
    values: ["true"]
```

deploy/backplane/osdfm/20-osdfm.SubjectPermission.yml
tiwillia: Will make this change.

As for the card, its not closed but the parent was: https://issues.redhat.com/browse/SDA-8133 I re-opened.

tiwillia:
It says the PR build failed but I don't have access to be able to see why :man_shrugging:

PR #1599, opened 2023-04-10
Add Sids to all statements in HCP Control Plane policy

fahlmant:
AWS is requiring all statements in policies to have Sids. This PR adds a Sid to every statement in the HCP control plane policy that doesn't already have one


jharrington22 (approved):
/lgtm

PR #1603, opened 2023-04-12
Remove AssumeRoleWithWebIdentity from the HCP installer policy as it'…

fahlmant:
According to https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html 

`Calling AssumeRoleWithWebIdentity does not require the use of AWS security credentials.`
I installed an HCP cluster successfully without this permission explicitly in the HCP Installer policy, as it's presence in the role is not required for OCM to use it.


jharrington22 (approved):
/lgtm

PR #1604, opened 2023-04-12
Rename Sid for RunInstances restriction based on AWS feedback

fahlmant:
This change clarifies that this Statement restricts RunInstances to Red Hat owned AMIs, rather than a generic "restricted" message


jharrington22 (approved):
/lgtm

PR #1610, opened 2023-04-17
Add request region name to image registry ARN

jharrington22:
### What type of PR is this?
feature

Further restrict image registry ARN with AWS request region name.

The image registry operator [adds](https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/s3/s3.go#L648) the region name when constructing the bucket name so this should be safe.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/approve
/lgtm

PR #1611, opened 2023-04-18
added destination: "generated_deploy/acm-policies" in policy field in config.yaml

lnguyen1401:
### What type of PR is this?
feature

added destination in config.yaml so the new MCC refactor script can generate a policy based on config.yaml and remove the lists we're keeping in https://github.com/openshift/managed-cluster-config/blob/master/scripts/generate-policy-config.py , https://github.com/openshift/managed-cluster-config/blob/master/scripts/generate-subjectpermissions-policy-config.py

https://issues.redhat.com/browse/OSD-16052
_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bdematte:
To properly work, I think we need all the generated resources under the same path (to be able to add all content to the template for hive) so the path value should rather be `acm-policies` than `generated_deploy/acm-policies`. If not, we will likely generate resources which aren't deployed.

In addition, we need to ensure we create a card for the policy which aren't managed through the script right now to ensure they are moved to new script as a follow-up step (and we add the policy destination at that time).

lnguyen1401:
That makes sense. I have updated the destination path, and open a card for the resources not managed by script

lnguyen1401:
cc @dustman9000

jewzaam:
/hold

Want to understand downstream PR https://github.com/openshift/managed-cluster-config/pull/1689 before we accept this change.

rafael-azevedo (approved):
/lgtm

fahlmant (approved):
/lgtm
/approve

PR #1612, opened 2023-04-18
Condition HCP KMS permissions by the red-hat-managed: true tag

fahlmant:
Conditions the HCP KMS permissions to only allow use of KMS keys with the `red-hat-managed: true` flag.

### DIFF
--- a/resources/sts/4.12/hypershift/openshift_hcp_kms_provider_credential_policy.json
+++ b/resources/sts/4.12/hypershift/openshift_hcp_kms_provider_credential_policy.json
@@ -12,8 +12,8 @@
         ],
         "Resource": "*",
         "Condition": {
-          "StringLike": {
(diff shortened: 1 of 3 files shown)


arendej:
We should use a different tag for this such as: 
`rosa-key:true` so we'd see this in the policy:
`"aws:ResourceTag/rosa-key": "true"`

resources/sts/4.12/hypershift/openshift_hcp_kms_provider_credential_policy.json
```diff
@@ -14,6 +14,9 @@
         "Condition": {
           "StringLike": {
             "kms:ViaService": "ec2.*.amazonaws.com"
```
jharrington22, line 16: @fahlmant do we need to remove these as well?

jharrington22 (approved):
/lgtm

PR #1616, opened 2023-04-19
Move to red-hat: true rather than rosa-key to be more generic

fahlmant:
https://issues.redhat.com/browse/OSD-14652


jharrington22 (approved):
/lgtm

PR #1618, opened 2023-04-19
Add kms permission to installer for verifier, and correct tag in NodePool controller

fahlmant:
Adds KMS CreateGrant to the installer role so OCM can create the osd-network-verifier instance. Also fixes the tag expected on the KMS key in the NodePool controller


jharrington22 (approved):
/lgtm

PR #1619, opened 2023-04-19
Remove unused permissions for KMS cluster API provider

jharrington22:
### What type of PR is this?
bug

Removing permissions not required to encrypt EBS volumes in EC2. Scoping to permissions required in AWS documentation [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html)

HyperShift will also make the same change https://github.com/openshift/hypershift/pull/2456

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/LGTM
/approve

PR #1622, opened 2023-04-20
Add tag condition to DeleteLoadBalancer in hcp installer

fahlmant:
Add condition to require `red-hat-managed: true` tag on LB for DeleteLoadBalancer permission in the HCP installer.


fahlmant:
Closing in favor of #1623

PR #1623, opened 2023-04-20
Remove Create and Delete LoadBalancer from HCP Installer Policy

fahlmant:
In HCP, the installer doesn't use the CreateLoadBalancer or DeleteLoadBalancer policies. 

/hold 
Needs verification that the statement above is true.


jharrington22 (approved):
/lgtm

PR #1626, opened 2023-04-21
Update HCP Installer policy to handle verifier runs

fahlmant:
Enables the Installer policy to run the verifier for HCP installs

### DIFF
--- a/resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json
+++ b/resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json
@@ -7,6 +7,8 @@
             "Action": [
                 "ec2:DescribeAvailabilityZones",
                 "ec2:DescribeInternetGateways",
+                "ec2:DescribeInstances",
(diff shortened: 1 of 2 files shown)


resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json
```diff
@@ -176,38 +178,126 @@
...
             }
-          }
         },
         {
+            "Sid": "CreateGrantRestrictedResourceTag",
             "Effect": "Allow",
             "Action": [
-                "kms:CreateGrant"
+                "kms:CreateGrant",
+                "kms:GenerateDataKey*"
```
jharrington22, line 218: Can you update this to `GenerateDataKeyWithoutPlaintext` 

See [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html)

jharrington22 (approved):
/lgtm

PR #1627, opened 2023-04-21
Fix the hcp support to enable verifier runs

fahlmant:
- Merges the unabridged support policy with the managed policy definition.
- Adds permissions necessary for SRE to run the osd-network-verifier
- Adds missing Sids


jharrington22 (approved):
/lgtm

PR #1630, opened 2023-04-21
Remove AuthorizeSecurityGroupIngress from HCP NodePool controller

fahlmant:
NodePool controller doesn't use AuthorizeSecurityGroupIngress permission, removing from policy.


fahlmant:
I'm wrong, closing

jharrington22:
Used [here](https://github.com/openshift/hypershift/blob/00f9920e53175b8c55505562bbfecdf43a1b85d6/control-plane-operator/controllers/hostedcontrolplane/hostedcontrolplane_controller.go#L3884) for hypershift.

PR #1633, opened 2023-04-24
Add workaround job to remediate clusters impacted by OCPBUGS-773

mrbarge:
### What type of PR is this?
Bug

Applies a read-only workaround to 4.10 and 4.11 clusters in the version range impacted by [OCPBUGS-773](https://issues.redhat.com/browse/OCPBUGS-773).

The workaround should be a no-op for OVN clusters as they will have no pods in `openshift-sdn`.

- OCPBUGS-773
- OHSS-21214

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mrbarge:
/hold
for comment and discussion

mjlshen:
+1 to this direction

mrbarge:
Thanks for the review folks. Removing the hold and hoping for the best!
/unhold

PR #1641, opened 2023-04-25
Adds alerts for handling multiple prometheus operators

iamkirkbater:
### What type of PR is this?
Feature

This adds two new alerts - one for detecting multiple prometheus operators during a ClusterOperatorDown event effecting the monitoring operator, the other detects all other Cluster Monitoring down events.

[OSD-16158](https://issues.redhat.com//browse/OSD-16158)

Tested successfully against the following scenarios:

* Monitoring Operator is Healthy - neither should fire
* Monitoring Operator is Healthy but there is a second prometheus - neither should fire
* Monitoring Operator is Down but no second prom - ClusterMonitoringOperatorDownSRE should fire
* Monitoring Operator is Down but second prom - MultiplePrometheusOperatorSRE should fire

To "bring down" the monitoring operator - 
* Pause hive syncing
* `oc patch --as backplane-cluster-admin co monitoring --type merge --subresource status --patch '"status": {"conditions":[{"lastTransitionTime": "2023-04-25T13:58:30Z", "message": "failed", "reason": "RollOutFailed", "status": "False", "type": "Available"}]}'`
* You'll want to update the timestamp to be more recent than what it is in the patch above.

The monitoring operator should then be reporting as Down in prometheus metrics

To deploy a second monitoring stack - 
* spin up a stage cluster
* log into your cluster
* Follow this quick start guide: https://prometheus-operator.dev/docs/prologue/quick-start/
* Add the serviceaccounts to the privileged SCC
https://access.redhat.com/documentation/en-us/openshift_container_platform/3.6/html/cluster_administration/admin-guide-manage-scc#grant-access-to-the-privileged-scc
```oc get serviceaccount -n monitoring -o name | sed s#serviceaccount/#system:serviceaccount:monitoring:# | xargs oc adm policy add-scc-to-user privileged```

Give it a few minutes and you should have a few extra prometheus-k8s-* pods running in the `monitoring` namespace

### DIFF
--- a/deploy/sre-prometheus/100-monitoring-operatordown.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-monitoring-operatordown.PrometheusRule.yaml
@@ -1,0 +1,39 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-monitoring-operatordown
+    role: alert-rules
+  name: sre-monitoring-operatordown
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-monitoring-operatordown
+    rules:
+    - alert: ClusterMonitoringOperatorDownSRE
+      # Specifically for the Cluster Monitoring Operator, we want to see if the sum cluster_operator_up (which will turn to 0 when
+      # the operator is DOWN) minus the count of prometheus pods outside of the openshift-* namespaces equals -1. This means that if
+      # the cluster operator is down and there are no rogue prometheus instances this will fire.  If the monitoring operator goes
+      # down but there also is rogue prometheus instances, this would return 0 because of the `absent` and therefore will not fire.
+      expr: |
+        (sum(cluster_operator_up{name="monitoring"}) - absent(count(kube_pod_labels{label_app_kubernetes_io_component="prometheus",namespace!~"openshift-.*"}))) == -1
+      for: 10m
+      labels:
+        severity: critical
+        namespace: openshift-cluster-version
+      annotations:
+        message: The monitoring operator may be down or disabled, and the components it manages may be unavailable or degraded. Cluster upgrades may not complete. For more information refer to 'oc get -o yaml clusteroperator monitoring'.
+    - alert: MultiplePrometheusOperatorsSRE
+      # This alert will fire when there are multiple prometheus operators on the cluster unless the cluster monitoring operator is not down.
+      # We first get the labels from the pod with the namespace, adding a custom label to compare on against the count of the cluster monitoring
+      # operator pods available. If the cluster monitoring operator is up, this won't fire for just having additional prometheus instances. If
+      # the customer has multiple prometheus instances, and the cluster monitoring operator is down, this will fire.
+      expr: |
+         count(label_replace(kube_pod_labels{label_app_kubernetes_io_component="prometheus",namespace!~"openshift-.*"}, "sre_metric", "openshift-sre", "", "(.*)")) by (namespace, sre_metric) unless ignoring (namespace) count(label_replace(cluster_operator_up{name="monitoring"}, "sre_metric", "openshift-sre", "", "(.*)") > 0) by (sre_metric)
+      for: 10m
+      labels:
+        severity: info
+        namespace: openshift-monitoring
+      annotations:
+        message: "There are multiple prometheus operators that are impacting the cluster monitoring operator, in customer namespaces."
+
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -32833,6 +32833,46 @@ objects:
       kind: PrometheusRule
       metadata:
         labels:
+          prometheus: sre-monitoring-operatordown
+          role: alert-rules
+        name: sre-monitoring-operatordown
+        namespace: openshift-monitoring
+      spec:
+        groups:
+        - name: sre-monitoring-operatordown
+          rules:
+          - alert: ClusterMonitoringOperatorDownSRE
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: deploy/sre-prometheus/100-monitoring-operatordown.PrometheusRule.yaml [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+        message: The monitoring operator may be down or disabled, and the components it manages may be unavailable or degraded. Cluster upgrades may not complete. For more information refer to 'oc get -o yaml clusteroperator monitoring'.
+    - alert: MultiplePrometheusOperatorsSRE
+      # This alert will fire when there are multiple prometheus operators on the cluster unless the cluster monitoring operator is not down.
+      # We first get the labels from the pod with the namespace, adding a custom label to compare on against the count of the cluster monitoring
+      # operator pods available. If the cluster monitoring operator is up, this won't fire for just having additional prometheus instances. If
+      # the customer has multiple prometheus instances, and the cluster monitoring operator is down, this will fire.
+      expr: |
+         count(label_replace(kube_pod_labels{label_app_kubernetes_io_component="prometheus",namespace!~"openshift-.*"}, "sre_metric", "openshift-sre", "", "(.*)")) by (namespace, sre_metric) unless ignoring (namespace) count(label_replace(cluster_operator_up{name="monitoring"}, "sre_metric", "openshift-sre", "", "(.*)") > 0) by (sre_metric)
```
mrbarge, line 32: Open question, because I really don't know the answer - how confident are we that this identifies the presence of a prometheus __operator__ and not just another Prometheus? In the past I've spotted customer prometheuses running (due to running istio) that don't have an operator installed, which seems okay. Admittedly I can't remember any of those cluster names to check if they would match this metric.

Ultimately I just harbor a little concern about the potential of accidentally masking genuine cluster-monitoring-operator issues due to the presence of an innocuous Prometheus that's not causing any trouble.

iamkirkbater:
>> How is this possible?

> Have we discussed with OCP about changing the label on the CMO deployment?

Even if it's not possible, it doesn't hurt to have the contingency. Especially since we would "never know" if a customer's workload is deployed alongside our workload and doesn't interfere, since we'd  never have been alerted in the first place.

> Do we even want to send a service log via OAO compared to denying the actions via managed-cluster-validating-webhooks? Customers generally don't action service logs/there's a scalability problem with them vs just preventing the action with our guardrails. Though I suppose the two approaches aren't mutually exclusive.

Again this comes down to the possibility that there could be hundreds of non-interfering prometheus workloads that are out there that do not compete with our workloads that we just don't know about, and I would hate to sweeping block all of them just because a small portion of them are causing issues with our montioring stack.

deploy/sre-prometheus/100-monitoring-operatordown.PrometheusRule.yaml
iamkirkbater: This is a good callout, but the main reason as to why I wanted to specifically target only clusterOperatorDown with it. I wish there was a way to see how many of these in the past we've seen the ClusterMonitoringOperator go down but it's not due to the 2nd prometheus operator/instance when it's installed.  

This, though, is one of the main reasons that I am not also creating the OAO PR at the moment.  One of the goals of this is to capture this through a separate alert, so that we can start tracking some data through the TOIL reviews of "ok, this is firing but it had nothing to do with the second prometheus instance" which we would bake into the SOP itself.

iamkirkbater: As for your original question, though, this doesn't detect a prometheus operator at all, just the prometheus pod, which is what we're detecting for at the bottom of the SOP.

In my testing I didn't even see a second prometheus operator interfering directly, as I was able to run a second one for about 30-45 minutes before being able to circle back to my test cluster and everything was still running along fine.

So part of this alert is data collection, part of the alert is 🤞🏼 that we can just send a SL and never be alerted for this again.

mjlshen:
Agreed with your points, thanks Kirk! Want fo focus on this one:

> Again this comes down to the possibility that there could be hundreds of non-interfering prometheus workloads that are out there that do not compete with our workloads that we just don't know about, and I would hate to sweeping block all of them just because a small portion of them are causing issues with our montioring stack.

I think we should figure this out then, what interferes and why so we can document how customers can be successful and put guardrails around what we don't want customers doing. My cursory investigation which led to the SOP update concluded that it was kube-controller-manager wiping out these pods due to matching labels that the monitoring stack's `ReplicaSets` were fighting over.

For example, on a dev cluster I have the monitoring stack has the labelSelectors:

```yaml
spec:
  selector:
    matchLabels:
      app.kubernetes.io/instance: k8s 
      app.kubernetes.io/managed-by: prometheus-operator
      app.kubernetes.io/name: prometheus
      operator.prometheus.io/name: k8s 
      operator.prometheus.io/shard: "0" 
      prometheus: k8s 
```

So I would expect that this problem of a "second monitoring stack" only occurs when there are multiple pods with all of those labels as the requirements are ANDed together.

iamkirkbater:
I agree that we should figure this out as well.  By breaking these alerts apart, we do two things - 1 we gather more data on how often this is happening and 2 - from that data we can determine if this is a blanket cause or if there's something more to look into.

I've gotten a must-gather from a cluster where this is happening for when we open a bug with the Monitoring team. Ultimately, though, we would expect the default on-cluster operator to not clash with customer-installed workloads. However, even if we get an update from the upstream monitoring team, we'll still have to ask for that to be backported, and we'd also need customers to upgrade to that supported version.

To be explicitly clear here - we don't have any OAO automation in place yet to automatically handle these alerts, this is step 1 which is to break the alerts apart by cause.  Step 2 is collect some data on how often this is happening through telemeter, and step 3 is going to be deciding using the data from step 2 to determine what the next course of action is for SREP, at least. All of this can go on, in parallel with whatever the upstream monitoring team is doing and investigating.

Are you suggesting with your last comment that we should be alerting on ALL of those labels?

mjlshen:
> Are you suggesting with your last comment that we should be alerting on ALL of those labels?

Yeah, I think that the problem only occurs when all of the labels match - but haven't fully investigated this

PR #1642, opened 2023-04-26
OCPBUGS-773 split out seccomp and no-seccomp variants

mrbarge:
### What type of PR is this?
Bug

Fixes an issue in #1633 to ensure that it can run on clusters with seccomp and without.

OCPBUGS-773


NautiluX (commented):
/lgtm

PR #1656, opened 2023-05-03
Update the Template code to better support the following three scenar…

jnpacker:
…ios:

1. If ConfigMap is missing, set scope to External
2. If the Key is missing/invalid in the ConfigMap, set the scope to External
3. In both these cases, make sure the Config Policy evaluates as compliant so the Ingress gets configured.

### What type of PR is this?
Enhancement of feature
This adds better support for scenario 1 listed above
https://issues.redhat.com/browse/SDA-9047
_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jnpacker:
Rebasing in ocm-190 MR.

PR #1666, opened 2023-05-05
Add NLB support to the kube controller manager HCP policy

fahlmant:
Adds permissions related to NLB management by the kube-controller-manager HCP policy

### DIFF
--- a/resources/sts/4.12/hypershift/openshift_hcp_kube_controller_manager_credentials_policy.json
+++ b/resources/sts/4.12/hypershift/openshift_hcp_kube_controller_manager_credentials_policy.json
@@ -14,7 +14,8 @@
         "elasticloadbalancing:DescribeLoadBalancerAttributes",
         "elasticloadbalancing:DescribeListeners",
         "elasticloadbalancing:DescribeTargetGroups",
-        "elasticloadbalancing:DescribeTargetHealth"
+        "elasticloadbalancing:DescribeTargetHealth",
+        "elasticloadbalancing:DescribeLoadBalancerPolicies"
       ],
       "Resource": [
         "*"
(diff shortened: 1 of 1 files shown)


resources/sts/4.12/hypershift/openshift_hcp_kube_controller_manager_credentials_policy.json
```diff
@@ -49,6 +52,39 @@
...
+        "StringEquals": {
+          "aws:RequestTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "NetworkLoadBalanacerManagement",
+      "Effect": "Allow",
+      "Action": [
+        "elasticloadbalancing:CreateListener",
```
mjlshen, line 74: Any reason not to add `elasticloadbalancing:CreateTargetGroup` into this Sid?

I also think that:

`elasticloadbalancing:ModifyListener` and `elasticloadbalancing:DeleteListener` are likely to be needed in some workflows. Are we sure the rest in https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.12/openshift_kube_controller_manager_credentials_policy.json#L27-L55 are not needed?

fahlmant: CreateTargetGroup might be able to go into this block as well, but it needs to be in the `RequestTag` since it's a Create action. Any actions that creates a resource that is tagged with `red-hat-managed` true and can be restricted in AWS by RequestTag should be. 

For the other permissions, I'm working on testing NLB is hypershift to try and trigger them. They'll each need to be validated they they're used in an HCP context and checked if they can be restricted by one of our conditions.

fahlmant:
I have verified all the permissions are used through various ways with Service LBs on HCP. 

/hold cancel

PR #1669, opened 2023-05-08
Set Load Balancer type for ingressController to Classic

jnpacker:
### What type of PR is this?
BUG, to conform with AWS policies

Uses a Classic Load Balancer

https://issues.redhat.com/browse/SDA-9047

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/approve

PR #1672, opened 2023-05-10
OCM-190: Import ConfigMap support for ingress configuration.

jnpacker:
## Scenarios:

1. If ConfigMap is missing:
   * Do not include endpointPublishgStrategy in the IngressController policy
   * Do not add an annotation to the router-default ingress service
2. If the ConfigMap is present
   * Do not add an annotation to the service if aws-load-balancer-subnets if missing, otherwise add it.
   * If endpoint-publishing-strategy = internal, set the scope to Internal, otherwise its set to External
   * If loadbalancer-aws-type is missing, use Classic, otherwise use the value in loadbalancer-aws-type.
3. Add an annotation to the router-default service when the configMap key is present
   * If aws-load-balancer-subnets is present, apply the value to the annotation (comma separated list). If missing, skip modification of annotations on the service

INPUTs:
* ConfigMap in the openshift-acm-policies namespace, its name is the same as the HCP. It support for the following keys
  * aws-load-balancer-subnets: Comma separated list of subnet ID's or names
  * loadbalancer-aws-type: NLB / Classic
  * endpoint-publishing-strategy: internal / external  (resulting values are Internal or External)

### What type of PR is this?
feature

Adds further configuration to Ingress

https://issues.redhat.com/browse/OCM-190

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
@@ -15,6 +15,69 @@ spec:
             apiVersion: policy.open-cluster-management.io/v1
             kind: ConfigurationPolicy
             metadata:
+                name: rosa-ingress-controller-policies
+            spec:
+                evaluationInterval:
+                    compliant: 2h
+                    noncompliant: 45s
+                object-templates-raw: |-
+                    - complianceType: musthave
+                      metadataComplianceType: musthave
+                      objectDefinition:
+                        apiVersion: operator.openshift.io/v1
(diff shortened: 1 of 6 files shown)


jnpacker:
@cdoan1 can you give this a review. Thx.

cdoan1:
@jnpacker do we need to make changes to the config.yam ? 

deploymentMode: Policy ? is this still the deploymentMode ?

jnpacker:
I think the setting is fine because it created the appropriate yaml with make

fahlmant (approved):
/lgtm
/approve

PR #1676, opened 2023-05-18
WIP: HOLD - Back out the Service Annotation.

jnpacker:
* Backing out the service annotation just to be safe.
* OCM will still put a value in the ConfigMap, but by removing this policy entry, that value will not be applied inside the HCP
* This is a smaller known change, then modifying ClusterService
* If this change is NOT made, it should NOT affect the HCP, but it is safer and makes sure there are no unused artifacts in the env

### What type of PR is this?
BUG 🐞 

It will stop placing a list of subnets as an annotation on the ingress service. The list comes from a ClusterService provisioned ConfigMap.

[OCM-190](https://issues.redhat.com//browse/OCM-190)

_Fixes #_

Will validate that INT/STAGE do not show a large number of NonCompliant once this change is in place.


jnpacker:
@cdoan1 can you do a quick review!

jnpacker:
It was determined that the Annotation as is was working correctly and Cluster Service would adjust the values it is supplying.

PR #1680, opened 2023-05-23
OSD-16095 - Moving the rosa-oauth-template elements to the hosted control plane

bdematte:
### What type of PR is this?
Bug

Moving the targets for the oauth customization as this is not handled the same way on hosted clusters (different secret names, not possible to customize oauth object). 
As changes need to be applied to management clusters, it wasn't possible to test it on a non-intrusive way. However, this is only moving/deleting some resources which were non-effective and creating some secrets so should be safe to merge.  

[OSD-16095](https://issues.redhat.com/browse/OSD-16095)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


rafael-azevedo (approved):
/lgtm

PR #1683, opened 2023-05-23
Change node manager  PassRole resource ARN to be service specific

fahlmant:
### What type of PR is this?
cleanup 

This changes the PassRole action resource identifier to be more service specific to ROSA and match the installer role, ref #1682 

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jharrington22 (approved):
/lgtm

PR #1684, opened 2023-05-24
OSD-16520 - adding NodeFilesystemAlmostOutOfSpaceSRE alert

Makdaam:
, split for cp+infra and worker nodes

### What type of PR is this?
_feature_

This PR adds SRE versions of the upstream alert NodeFilesystemAlmostOutOfSpace, separate for the control plane + infra, which will be routed to the on-call SRE and worker nodes which will be automatically handled by ocm-agent.

_[OSD-16520](https://issues.redhat.com//browse/OSD-16520)_

Similar to https://github.com/openshift/managed-cluster-config/pull/1651 and https://github.com/openshift/managed-cluster-config/pull/1653.

There will be a separate CAMO pull request to disable the upstream alert.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-sre-node-spacefillingup.yaml
+++ b/deploy/sre-prometheus/100-sre-node-spacefillingup.yaml
@@ -1,0 +1,34 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-node-filedescriptor-limit
+    role: alert-rules
+  name: sre-node-filedescriptor-limit
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-controlplane-node-almostoutofspace
+    rules:
+    - alert: ControlPlaneNodeFilesystemAlmostOutOfSpaceSRE
+      # This is the same as the upstream alert, but groups by instance id (which is which node is affected)
+      # and then only fires on the control-plane, infra, or master node roles
+      # See the ocm-agent folder for the WorkerNode version of this alert
+      annotations:
+        description: Filesystem on {{ $labels.device }} at {{ $labels.instance }}
+          has only {{ printf "%.2f" $value }}% available space left.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/NodeFilesystemAlmostOutOfSpace.md
+        summary: Filesystem has less than 3% space left.
+      expr: |-
+        (
+          node_filesystem_avail_bytes{job="node-exporter",fstype!="",mountpoint!=""} / node_filesystem_size_bytes{job="node-exporter",fstype!="",mountpoint!=""} * 100 < 3
+        and
+          node_filesystem_readonly{job="node-exporter",fstype!="",mountpoint!=""} == 0
+        )
+        * on (instance) group_left () (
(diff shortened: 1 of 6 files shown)


deploy/sre-prometheus/100-sre-node-spacefillingup.yaml
```diff
@@ -0,0 +1,35 @@
...
+      # This is the same as the upstream alert, but groups by instance id (which is which node is affected)
+      # and then only fires on the control-plane, infra, or master node roles
+      # See the ocm-agent folder for the WorkerNode version of this alert
+      annotations:
+        description: Filesystem on {{ $labels.device }} at {{ $labels.instance }}
+          has only {{ printf "%.2f" $value }}% available space left.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/NodeFilesystemAlmostOutOfSpace.md
+        summary: Filesystem has less than 3% space left.
+      expr: |-
+        group by (instance)
```
iamkirkbater, line 23: I think if we remove this `group by (instance)` line it should still work, I tested this with the other NodeFilesystemSpaceFillingUp alert, and that ends up preserving the labels on the alert itself while only showing the ones firing on whatever nodetype is defined under `* on (instance)` and below

deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
```diff
@@ -73,3 +73,32 @@ spec:
...
+      # metric it splits on the roles, so infra,worker then gives two results, one with just the infra role
+      # and one with just the worker role.  So we get just the infra/controlplane nodes and then exclude
+      # them from the worker node result, which gives us just the customer worker nodes.
+      annotations:
+        description: Filesystem on {{ $labels.device }} at {{ $labels.instance }}
+          has only {{ printf "%.2f" $value }}% available space left.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/NodeFilesystemAlmostOutOfSpace.md
+        summary: Filesystem has less than 3% space left.
+      expr: |-
+        group by (instance)
```
iamkirkbater, line 88: Same here, we might want to not `group by (instance)` because then it preserves the value and original labels of the original metric.

iamkirkbater (changes requested):

deploy/sre-prometheus/100-sre-node-spacefillingup.yaml
Makdaam: Good idea to keep the details that we can get from the metrics.

simonpasquier:
OCP monitoring team member here, is there something that we can address "upstream"? In particular, would it be useful if the existing NodeFilesystem* alerts carry a `role` label?

cc @jan--f

iamkirkbater:
Hey @simonpasquier - that would be great! Though, we'd still need to put this workaround in place so that we can route our alerts until all managed clusters have upgraded to the OCP version where that is added. Is there a specific process we need to follow to request something like that? (I could have sworn that I created a JIRA or started a discussion about this at one point but I can't for the life of me find it)

simonpasquier:
We can continue the conversation on the `#forum-openshift-monitoring` Slack channel. This could even be something relevant for https://github.com/kubernetes-monitoring/kubernetes-mixin/.

PR #1687, opened 2023-05-25
OSD-16223 - Add alert for detecting multiple version of EFS pods

Tafhim:
### What type of PR is this?
feature

This PR adds Prometheus Rule to detect if multiple version of EFS operator exists

[OSD-16223](https://issues.redhat.com//browse/OSD-16223)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: aws-efs-alerts
+    role: alert-rules
+  name: aws-efs-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: aws-efs-alerts 
+    rules:
+    - alert: MultipleVersionsOfEFSCSIDriverInstalled 
+      expr: count(kube_pod_info{namespace="openshift-operators", pod=~"efs-csi-node-.*"}) > 0 and count(kube_pod_info{namespace="openshift-cluster-csi-drivers", pod=~"aws-efs-csi-driver-operator-.*"}) > 0
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+        source: "https://issues.redhat.com/browse/OSD-16223"
+        link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/MultipleVersionsOfEFSCSIDriverInstalled.md"
+      annotations:
+        message: Multiple versions of EFS CSI Driver installed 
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    prometheus: aws-efs-alerts
+    role: alert-rules
+  name: aws-efs-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: aws-efs-alerts 
+    rules:
+    - alert: MultipleVersionsOfEFSOperatorInstalled
```
bmeng, line 13: Better name it with multiple versions of EFS CSI driver installed. Since the real issue here is the csi driver but not the operator

deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
```diff
@@ -0,0 +1,19 @@
...
+  groups:
+  - name: aws-efs-alerts 
+    rules:
+    - alert: MultipleVersionsOfEFSOperatorInstalled 
+      expr: count(kube_pod_info{namespace="openshift-operators", pod=~"efs-csi-node-.*"}) > 0 and count(kube_pod_info{namespace="openshift-cluster-csi-drivers", pod=~"aws-efs-csi-driver-operator-.*"}) > 0
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+      annotations:
+        message: Multiple versions of EFS Operator installed
```
bmeng, line 19: Same as above

deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
```diff
@@ -0,0 +1,21 @@
...
+  groups:
+  - name: aws-efs-alerts 
+    rules:
+    - alert: MultipleVersionsOfEFSCSIDriverInstalled 
+      expr: count(kube_pod_info{namespace="openshift-operators", pod=~"efs-csi-node-.*"}) > 0 and count(kube_pod_info{namespace="openshift-cluster-csi-drivers", pod=~"aws-efs-csi-driver-operator-.*"}) > 0
+      labels:
+        severity: critical
+        namespace: openshift-monitoring
+        source: "https://issues.redhat.com/browse/OSD-16223"
+        link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/MultipleVersionsOfEFSCSIDriverInstalled.md"
```
boranx, line 19: the link throws 404 for me. For a better primary experience, it might be nice to implement this beforehand

mrbarge:
If the only action on this alert is to send a service log, could we onboard it into OCM Agent instead, so that SRE has to do nothing at all?

deploy/sre-prometheus/100-multiple-efs-operator-PrometheusRule.yaml
bmeng: FYI, I think this should be populated automatically https://github.com/openshift/configure-alertmanager-operator/blob/master/controllers/secret_controller.go#L504

PR #1691, opened 2023-05-26
[OSD-16075] Add new SL for audit-events failing delivery.

bergmannf:
This SL will be send, when a hypershift-cluster fails to send audit-events to the customer's configured CloudWatch, because the event was too large (>256kb) *and* the delivery of a smaller event to indicate this failure also failed.

### What type of PR is this?

feature

Add a new notification to automatically send an SL, when this alert is observed.

Closes [OSD-16075](https://issues.redhat.com//browse/OSD-16075)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocm-agent-operator-managedfleetnotifications/config.yaml
+++ b/deploy/ocm-agent-operator-managedfleetnotifications/config.yaml
@@ -1,0 +1,8 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+  - key: api.openshift.com/fedramp
+    operator: NotIn
+    values:
+      - "true"
--- a/deploy/ocm-agent-operator-managedfleetnotifications/OWNERS
+++ b/deploy/ocm-agent-operator-managedfleetnotifications/OWNERS
@@ -1,0 +1,4 @@
+reviewers:
+- mrbarge
+- bmeng
+- ravitri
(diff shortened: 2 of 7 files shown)


mrbarge:
Thanks for doing this @bergmannf ! It looks perfect as far as I can tell. I'll put a hold on, feel free to unhold when you're all good.

/hold
/lgtm

deploy/ocm-agent-operator-managedfleetnotifications/config.yaml
```diff
@@ -0,0 +1,12 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+  - key: api.openshift.com/name
+    operator: In
+    values:
+      - "rhobsp02ue1"
```
mrbarge, line 8: There's nothing specifically _wrong_ with this, but I will just note this is a bit of an anti-pattern from an MCC perspective, we do not normally want to encourage targeting specific clusters (and names are not unique).

It would be fine for this resource to land on the entire fleet, as the 'fleet mode' operation of OCM Agent is not operational on anything but RHOBS clusters.

bergmannf: As you already pointed out - I only added this, because it was explicitly called out in the old version of the SOP.
I just removed that selector now.

bergmannf:
RHOBS and fleet-manager PRs are now merged, so I am unholding this now.
/unhold

PR #1695, opened 2023-05-31
Stop processing the annotation for the Ingress's Service resource

jnpacker:
### What type of PR is this?
* OCM-923, we want to make sure we remove the annotation from the Ingress Service resource.  There is a corresponding change in ClusterService to stop placing these details in a ConfigMap

* OCM-923 asks to remove the annotation. This makes sure the annotation is not created, even if ClusterService accidentally creates the configMap Key/Value (like it does today)

OCM-923

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


cdoan1:
looks good to me

cblecker (approved):

PR #1699, opened 2023-06-03
OSD-16737 make OCPBUGS-773 work on OVN clusters too

mrbarge:
### What type of PR is this?
Bug

Switches the OCPBUGS-773 workaround (#1633) to use the `openshift-cluster-node-tuning-operator` pods for carrying out the workaround, rather than the SDN pods, so that OVN clusters can also benefit from the fix.

The CronJob to remediate clusters impacted by OCPBUGS-773 (#1633) previously only operated on SDN clusters, because that was what the workaround was geared around. However, we have observed OVN clusters experiencing the same issue and they're subsequently not benefiting from this 

However, the main reason for using the SDN pods seems to be in order to run a `chroot /host && podman version` on the underlying node. We can do that from a workload which is universal to both SDN and OVN clusters, the `openshift-cluster-node-tuning-operator` `tuned` pods.

[OSD-16737](https://issues.redhat.com//browse/OSD-16737)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocpbugs-773/no-seccomp/10-OCPBUGS-773.CronJob.yaml
+++ b/deploy/ocpbugs-773/no-seccomp/10-OCPBUGS-773.CronJob.yaml
@@ -47,16 +47,16 @@ spec:
             args:
             - -c
             - |
-              pods=$(oc get pod -n openshift-sdn -l app=sdn -oname)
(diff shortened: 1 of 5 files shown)


mjlshen (approved):
/lgtm

PR #1705, opened 2023-06-09
* Remove ELB type, this will cause the system to default to Classic.

jnpacker:
* OCP-HyperShift will start to provide the type value
* Make it so the Policy will only apply if the IngressController default resource is present
* If the policy is removed, we DO NOT remove the resource Testing
* Validated the policy on ACM 2.7.4 and ACM 2.8.0

### What type of PR is this?
bug

* It allows for the HostedCluster to eventually control the ELB type.
* For now the ELB will default to Classic

https://issues.redhat.com/browse/ACM-5858
_Fixes #_

NA

- [x] Tested latest changes against a cluster. (ACM 2.7.4 & ACM 2.8)


jnpacker:
@cdoan1 looking for a 👍🏼

cdoan1:
👍 
/lgtm

PR #1709, opened 2023-06-14
CreateVPCEndpoint now specifies a security group 

fahlmant:
As part of the fix in https://issues.redhat.com/browse/OCPBUGS-11894, CreateVPCEndpoint now specifies a security group, specifically one that's created for the cluster. In order to allow the Control Plane operator to use the security group, a new SID is added to allow CreateVPCEndpoint with any security group with a `red-hat-managed: true` tag.


jharrington22 (approved):
/lgtm

PR #1710, opened 2023-06-14
Create RBAC For SD-CICD Role

MrSantamaria:
Feature PR 

Follows https://source.redhat.com/groups/public/sre/wiki/backplane_new_team_on_boarding
SDCICD is adding backplane access to check arbitrary clusters post deployment for an acceptance test in the Tekton environment for Progressive Delivery of Operators. 
We need to rewrite the logic of the acceptance test we wrote over here:
https://github.com/openshift/aws-vpce-operator/blob/main/hack/tests/avo-acceptance-test.yaml#L59-L128
To not pull Kubeconfigs directly from Hives and instead use RBACs to authenticate into clusters. 

Jira:
[SDCICD-993](https://issues.redhat.com//browse/SDCICD-993)

This is my first time creating RBACs in this style. My thought process was to look at SRE's permissions and pull the access to be mostly read only since we don't want to alter clusters. 
Our goal is to use the RBACs in this case to do zero touch testing in higher level environments. Meaning we will not be modifying clusters but validating values. 
IE: A SSS upgrade can be found in a cluster managed by hive through regions.

### DIFF
--- a/deploy/backplane/sdcicd/01-sdcicd-read-only-cluster.ClusterRole.yml
+++ b/deploy/backplane/sdcicd/01-sdcicd-read-only-cluster.ClusterRole.yml
@@ -1,0 +1,29 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: backplane-sdcicd-readers-cluster
+rules:
+# SDCICD can read namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+# SDCICD can read OLM objects
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - catalogsources
+  - clusterserviceversions
+  - installplans
+  - subscriptions
+  - operatorconditions
+  - operatorgroups
+  - operators
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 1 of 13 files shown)


deploy/backplane/sdcicd/20-sdcicd.SubjectPermission.yml
```diff
@@ -0,0 +1,18 @@
...
+kind: SubjectPermission
+metadata:
+  name: backplane-sdcicd
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+  - backplane-sdcicd-readers-cluster
+  - backplane-readers-cluster
+  permissions:
+  - clusterRoleName: backplane-sdcicd-admins-project
```
dustman9000, line 11: this should not be needed

deploy/backplane/sdcicd/01-sdcicd-read-only-cluster.ClusterRole.yml
```diff
@@ -0,0 +1,29 @@
...
+# SDCICD can read namespaces
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+# SDCICD can read CSR objects
```
dustman9000, line 15:
```suggestion
# SDCICD can read OLM objects
```

PR #1720, opened 2023-06-28
Update ebs IOPS prometheus rule

hbhushan3:
### What type of PR is this?
deploy/sre-prometheus/aws/ebs-iops-burstbalance update - remove from FedRAMP as gp3 is used in FedRAMP - OCP >=4.10

EBS IOPS PrometheusRule is not needed in FedRAMP and the [workload](https://github.com/openshift/managed-prometheus-exporter-ebs-iops-reporter/blob/master/hack/00-osd-managed-prometheus-exporter-ebs-iops-reporter.selectorsyncset.yaml.tmpl#L15-L23) generating the metric is not deployed on STS clusters. 

[OSD-15402](https://issues.redhat.com/browse/OSD-15402)


hbhushan3:
Created another PR - 1722. Closing this due to some permission issue with make command on localhost

PR #1725, opened 2023-06-29
Restrict the SDCICD role to int and stage

MrSantamaria:
Restricts the SDCICD role to be available only in Integration and Stage.

### DIFF
--- a/deploy/backplane/sdcicd/config.yaml
+++ b/deploy/backplane/sdcicd/config.yaml
@@ -1,3 +1,7 @@
-deploymentMode: "SelectorSyncSet" 
-selectorSyncSet: 
-  resourceApplyMode: "Sync"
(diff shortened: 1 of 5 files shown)


deploy/backplane/sdcicd/config.yaml
```diff
@@ -1,3 +1,3 @@
-deploymentMode: "SelectorSyncSet" 
-selectorSyncSet: 
-  resourceApplyMode: "Sync"
+deploymentMode: "Direct"
```
dustman9000, line 1:
```suggestion
deploymentMode: SelectorSyncSet
selectorSyncSet:
  matchExpressions:
    - key: api.openshift.com/fedramp
      operator: In
      values:
        - "true"
    - key: api.openshift.com/environment
      operator: NotIn
      values:
        - "production"
```

PR #1732, opened 2023-07-10
OSD-17329 - adjusting 'CannotRetrieveUpdatesSRE' alert to provide generic summary

theautoroboto:
### What type of PR is this?
bug

Our alerting framework in FedRAMP uses summary is the dedupe key. With each change in the dynamic value we receive a new alert. This PR will set the summary as a static value and provide the time details in the description

_Fixes #_
Dedupe issue in GoAlert

- [ X ] Tested latest changes against a cluster
- [ N/A ] Included documentation changes with PR
- [ X ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (commented):
/lgtm

PR #1743, opened 2023-07-14
[OSD-17347] Make token refresher FIPS compliant

karthikperu7:
### What type of PR is this?
Security fix

Updating the token refresher image to use FIPS compliant version

https://issues.redhat.com/browse/OSD-17347

Tested on a stage cluster and it worked ok. metrics were still passing through to observatorium

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


karthikperu7:
Going to put a hold so that I can unhold when I am back on keys and monitor it going to stage. lgtm welcome though!

/hold

karthikperu7:
Should probably have added, the new image was built with FIPS after our fix here: https://github.com/observatorium/token-refresher/pull/30

/unhold

PR #1745, opened 2023-07-16
Allow DNSErrors10MinSRE to reset in the 10 minutes interval

typeid:
### What type of PR is this?
This should get rid of most unactionable DNSErrors10MinSRE alerts, as it stops the alert from triggering for short window failures. 

Checking if there's over 3 failures over 10 minutes **for 10 minutes** is redundant. Instead, with this change, we check if we had at least one failure every 3 minutes over a 10 minute window. 

Primary noise reduction

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
@@ -11,9 +11,9 @@ spec:
   - name: sre-dns-alerts
     rules:
     - alert: DNSErrors10MinSRE
-      # Alerts if we had any DNS failures in a 3 minutes interval three times in a row
-      expr: increase(dns_failure_failure_total[3m]) >= 1
-      for: 10m
+      # Alerts if we had DNS failures in multiple 4 minute windows for 15 minutes.
+      expr: increase(dns_failure_failure_total[4m]) >= 1
+      for: 15m
       labels:
         severity: critical
         namespace: openshift-monitoring
--- a/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
@@ -11,8 +11,8 @@ spec:
   - name: sre-dns-alerts
     rules:
     - alert: DNSErrors10MinSRE
-      # If the amount of DNS failures increases by 3 over the span of 10 minutes
-      expr: increase(dns_failure_failure_total[10m]) >= 3
+      # Alerts if we had any DNS failures in a 3 minutes interval three times in a row
+      expr: increase(dns_failure_failure_total[3m]) >= 1
       for: 10m
       labels:
         severity: critical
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
mjlshen, line 1: I need some help clarifying how this makes the alert less flappy - this is what I believe the alerts mean:

* Original: If (in the last 10 minutes, the total DNS failures increased by 3) for 10 minutes, fire. So roughly first alert comes in 10 minutes after the third failure and it doesn't flap as long as 3 failures happen every 10 minutes
* New: If (in the last 3 minutes, the total DNS failures increased by 1) for 10 minutes, fire. So roughly first alert comes in 10 minutes after the first failure and it doesn't flap as long as 1 failure happens every 3 minutes

So the original alert should be flappy if many failures happen within 10 minutes, stop for 10 minutes, then happen again, etc. The new alert attempts to address this by saying, only fire if we're consistently getting at least one failure every 3 minutes and avoids alerting when there are a cluster of failures that don't result in at least one failure per 3 minutes?

typeid: In essence:

This is the view of the `dns_failure_failure_total` counter for a cluster that had DNSErrors10MinSRE triggering and auto resolving a few times over the past day.
Screenshot from 2023-07-17 08-23-54

The increases always have the same scheme. Here, between 2:28 PM and 2:36 PM, we have had slight increases in errors over about 8 minutes. 
image

Our SOP mentions that these errors self-resolve after about 10 minutes. This is what I want to address in this PR: I don't want the alerts that would self resolve to even trigger in the first place. 

The reason they self-resolve after about 10 minutes is that we use the increase(...[10m]) sliding window. If an increase happens at minute 0, but then never increases again, it will still be relevant for the whole 10 minutes. 

In our case, we have constant increases over 8 minutes, which means that our sliding window is 18 minutes:
````
sliding window = last_increase_time - first_increase_time + 10 min
```
Our alert then self resolves after last increase + 10 minutes. 

**I think there's a few things to consider here:**

Do we get any benefit from the short alerts we currently have? 
- If yes, it's still probably not something that our primary wants on their board. Our SOP mentioning the alerts `self-resolve after about 10 minutes` is a red flag IMO. 
- If no, we should change the alert to not trigger for the usual self-resolving noise. We could change the sliding window time to be lower, and slightly increase the `for` of the expression to make such we filter out all noise:

```
- alert: DNSErrors10MinSRE
           # 4 Minutes seem to be needed to not reset the increase inbetween, I am not sure yet why, as the increases happen in 2 minutes intervals, so 3 minutes should have been enough.
            expr: increase(dns_failure_failure_total[4m]) >= 1 
           # Increase this to 15, so that constant increases over 8 minutes don't alert
            for: 15m 
            labels:
              severity: critical
```

Sliding window change differences: 
![image](https://github.com/openshift/managed-cluster-config/assets/29230116/3a3f8a99-cd92-4748-ac52-6c7ffc4e4eb6)
vs 
![image](https://github.com/openshift/managed-cluster-config/assets/29230116/64f11c24-8300-449e-9b52-54e9a2059d35)

PR #1747, opened 2023-07-19
Fix missing when SCPs enforce S3 versioning

rafael-azevedo:
### What type of PR is this?
Bugfix

There is a Permissions issue when ROSA SCPs enforces S3 versioning

https://issues.redhat.com/browse/OSD-17500


fahlmant (approved):
For clarity: @rafael-azevedo and I confirmed the s3 versioning enforcement is not from an SCP config, but instead some AWS Config rules and remediation actions. To emulate this, we enabled bucket versioning manually on the bootstrap s3 bucket immediately after creation. Without these permissions, the bucket was not able to be deleted after boostrapping was completed. With these permissions, the s3 bucket was deleted.

/lgtm
/apprve

PR #1750, opened 2023-07-20
SRE-P permissions to manage Dynatrace / Otel resources

mrbarge:
### What type of PR is this?
Feature

Allows members of the SRE-P group to monitor Dynatrace and OpenTelemetry related resources on HyperShift Management/Service clusters without elevation.

- get/list/watch on `dynakubes`
- get/list/watch on `opentelemetrycollectors`

These permissions are tightly bound to specific namespaces so are not needed to be ClusterRoles and/or managed by rbac-permissions-operator.

[OSD-17552](https://issues.redhat.com//browse/OSD-17552)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/srep/dynatrace/10-srep-dynatrace.Role.yml
+++ b/deploy/backplane/srep/dynatrace/10-srep-dynatrace.Role.yml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: backplane-srep-dynatrace
+  namespace: dynatrace
+rules:
+# can view dynakubes
+- apiGroups:
+  - dynatrace.com
+  resources:
+  - dynakubes
+  verbs:
+  - list
+  - get
+### END
--- a/deploy/backplane/srep/dynatrace/10-srep-dynatrace.Role.yml
+++ b/deploy/backplane/srep/dynatrace/10-srep-dynatrace.Role.yml
@@ -12,4 +12,5 @@ rules:
   verbs:
   - list
   - get
+  - watch
 ### END
--- a/deploy/backplane/srep/dynatrace/10-srep-opentelemetry.Role.yml
+++ b/deploy/backplane/srep/dynatrace/10-srep-opentelemetry.Role.yml
@@ -1,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: backplane-srep-opentelemetry
+  namespace: openshift-opentelemetry-operator
+rules:
+# can view opentelemetry collectors
+- apiGroups:
+  - opentelemetry.io
+  resources:
+  - opentelemetrycollectors
+  verbs:
+  - list
+  - get
+  - watch
+### END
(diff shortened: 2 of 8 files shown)


ravitri:
I only had one query regarding `instrumentation` resource but that's taken care of already. LGTM from my end.
/lgtm

mrbarge:
@wanghaoran1988 @jewzaam could you please review this from a backplane approver perspective?

deploy/backplane/srep/dynatrace/10-srep-dynatrace.Role.yml
```diff
@@ -0,0 +1,15 @@
...
+  namespace: dynatrace
+rules:
+# can view dynakubes
+- apiGroups:
+  - dynatrace.com
+  resources:
+  - dynakubes
+  verbs:
+  - list
+  - get
```
jewzaam, line 14:
```suggestion
  - get
  - watch
```

deploy/backplane/srep/dynatrace/10-srep-opentelemetry.Role.yml
```diff
@@ -0,0 +1,16 @@
...
+kind: Role
+metadata:
+  name: backplane-srep-opentelemetry
+  namespace: openshift-opentelemetry-operator
+rules:
+# can view opentelemetry collectors
+- apiGroups:
+  - opentelemetry.io
+  resources:
+  - opentelemetrycollectors
```
jewzaam, line 11:
```suggestion
  - opentelemetrycollectors
  - instrumentations
```

deploy/backplane/srep/dynatrace/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: ext-hypershift.openshift.io/cluster-type
+    operator: In
+    values: ["management-cluster","service-cluster"]
```
jewzaam, line 6: There isn't some dynatrace specific label we can use?  This will work but will require changes when this stack is used beyond HCP ROSA.

jewzaam (changes requested):

deploy/backplane/srep/dynatrace/10-srep-opentelemetry.Role.yml
mrbarge: I originally had this present, but removed it because we don't expect to have any instrumentations or be interacting with them. No problem with it being included though.

deploy/backplane/srep/dynatrace/config.yaml
mrbarge: There's not, no - this is the only way we have at the moment of targeting where Dynatrace is landing, and is consistent with the same targeting we use in the SelectorSyncSets for the DT Operator and configs.

Granted, we do know that this will also ideally need to land on some AppSRE clusters too, but we do presently lack any clusterdeployment labelling that helps us better target everything.

jewzaam (approved):

PR #1758, opened 2023-07-21
[OCM-1016] Update HCP installer role permission policy

davidleerh:
### What type of PR is this?
_(feature)_

When a user tries to configure "audit log forwarding" on their HCP clusters, it requires the user to create the audit log forwarding role. Some of the prerequisites include the setup of trust policy, permissions policy (based on documentation: https://access.redhat.com/solutions/7002219) as well as having an OIDC provider.

This change allows OCM (Clusters Service) to assume role into the user's installer role and run validation on whether the user properly set up the needed OIDC provider.

_Fixes #_
https://issues.redhat.com/browse/OCM-1016

We are only adding the "iam:GetOpenIDConnectProvider" permission to HCP cluster installer roles at the moment, since "Cloudwatch audit log forwarding" feature is currently available for HCP clusters ONLY.

NOTE: changes here need to be requested and made to AWS managed policy as well (ROSA HCP installer role)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json
+++ b/resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json
@@ -20,6 +20,7 @@
                 "ec2:DescribeInstanceTypeOfferings",
                 "elasticloadbalancing:DescribeAccountLimits",
                 "elasticloadbalancing:DescribeLoadBalancers",
+                "iam:GetOpenIDConnectProvider",
                 "iam:GetRole",
                 "route53:ListHostedZones",
                 "route53:ListResourceRecordSets",
--- a/resources/sts/4.13/hypershift/sts_hcp_installer_permission_policy.json
+++ b/resources/sts/4.13/hypershift/sts_hcp_installer_permission_policy.json
@@ -20,6 +20,7 @@
                 "ec2:DescribeInstanceTypeOfferings",
                 "elasticloadbalancing:DescribeAccountLimits",
                 "elasticloadbalancing:DescribeLoadBalancers",
+                "iam:GetOpenIDConnectProvider",
                 "iam:GetRole",
                 "route53:ListHostedZones",
                 "route53:ListResourceRecordSets",


fahlmant:
@davidleerh I think I misunderstood. This only applies to hypershift but not Classic? Do you need this permission immediately, or just for managed policies (which will take a few weeks)?

davidleerh:
@fahlmant 

At the moment this change is required only for hypershift clusters. 
When a user tries to enable "audit log forwarding" for a cluster (both Day 1 and Day 2), CS first checks whether the cluster is hypershift or not. If it is hypershift, it then runs the OIDC provider check and ultimately proceeds with creating a secret used by the audit exporter for log forwarding. 

Once (if) ROSA allows for users to set up audit log forwarding for classic clusters, I believe we can add the "iam:GetOpenIDConnectProvider" into the classic cluster installer roles. Please advise if this is incorrect.

The change to both managed and unmanaged installer role policies is not urgent at the moment; however, I wanted to address the issue raised by QE to improve the error message being returned to the user; to better inform the user of any  missing prerequisites needed for log forwarding. 

In the meantime, we are pursuing to alert users the possibilities of configuration failure through a generic error returned
(e.g. "Audit log forwarding cannot be set up due to misconfigured trust policy or missing OIDC provider) instead of pinpointing the exact cause (e.g. "need to setup OIDC provider") - until the installer permissions are updated.

fahlmant:
@davidleerh thanks for the context! The "unmanaged" policies we're currently using for HCP are just the classic roles, which are in `resources/sts/4.12` and `resources/sts/4.13`, which the managed policies are in the `hypershift/` directories within those. So currently the PR wouldn't affect unmanaged policies in HCP.

davidleerh:
@fahlmant 
Thank you for the clarification!
In that case, since there is currently no differentiation between the installer permission policies (in unmanaged) between HCP and Classic, would it be okay to add the IAM:GetOpenIDConnectProvider permission onto the unmanaged installer policy being referenced?

davidleerh:
Adding the IAM:GetOpenIDConnectProvider permission to HCP installer managed policies only.

fahlmant (approved):
/approve
/lgtm

PR #1759, opened 2023-07-24
Ignoring resize alerts for upgrade

bdematte:
### What type of PR is this?
Bug

The *NeedResizing alerts are based on node count only, and as such, are requiring check and are most of the time silenced (as the cluster CPU/Mem consumption on masters do not require a scale up). As a consequence, managed-upgrade-operator should ignore those alerts are we are expecting those to remain 'Firing' in valid cases. 

https://issues.redhat.com/browse/OHSS-24871

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
+++ b/deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
@@ -51,6 +51,7 @@ data:
       - ThanosNoRuleEvaluations
       # OSD-13649
       - etcdGRPCRequestsSlow
+      # OHSS-24871
       - MasterNodesNeedResizingSRE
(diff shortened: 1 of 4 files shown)


deploy/managed-upgrade-operator-config/10-managed-upgrade-operator-configmap.yaml
```diff
@@ -51,6 +51,8 @@ data:
       - ThanosNoRuleEvaluations
       # OSD-13649
       - etcdGRPCRequestsSlow
+      - MasterNodesNeedResizingSRE
```
sam-nguyen7, line 54:
```suggestion
      # OHSS-24871
      - MasterNodesNeedResizingSRE
```

PR #1761, opened 2023-07-24
Change deployment of EBS related alerts to only non-CCS clusters

fahlmant:
### What type of PR is this?
Change

Changes EBS volume related alerts to only fire on non-CCS clusters. 
The EbsVolumeStuckAttaching10MinSRE alert usually resolves itself shortly after firing. The SOP itself can be simplified to "Check to see what namespace the volume is. If it's not a managed namespace, ignore the alert"

Additionally, the only EBS volumes that the core platform uses in AWS Managed OpenShift products are the prometheus and alertmanager volumes. If either of these two volumes get stuck attaching, this would present as a ClusterOperatorDown, alerting us to the issue, thus making this alert redundant. 

https://issues.redhat.com/browse/OSD-16627

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/aws/config.yaml
+++ b/deploy/sre-prometheus/aws/config.yaml
@@ -2,3 +2,7 @@ deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   matchLabels:
       hive.openshift.io/cluster-platform: aws
+  matchExpressions:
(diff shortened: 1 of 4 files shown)


mjlshen (approved):
/lgtm

PR #1762, opened 2023-07-24
Don't include remoteWrite for non-uwm CMO config

jewzaam:
### What type of PR is this?
bug

Fixes generation of CMO config for "non-uwm".  It is now consistent with the config that previously existed, does not include remoteWrite.

N/A

Untested as of yet...

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


jewzaam:
this configuration was why things were refactored, missed the context in OSD-12572

PR #1779, opened 2023-08-02
Splitting osd-serviceaccounts nonsts into private-link and not private-link

Makdaam:
### What type of PR is this?
_cleanup_

It stops syncing the cluster-ingress-operator SA to private-link clusters, which don't have the operator installed by default.
We need it to get ClusterSyncs healthy on Hive clusters, the sync fails due to missing operator namespace.

This gets MCC in sync with https://github.com/openshift/cloud-ingress-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L42-L53

_Fixes OSD-17775_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #1781, opened 2023-08-03
[OSD-16014] split the haproxyDown alert for default and non-default ingresscontroller

bmeng:
### What type of PR is this?
_(feature)_

As the ingresscontroller will be managed by the customer in the upcoming released feature, we should not receive the alert for the customer owned ingresscontrollers.
Split the existing HAProxyDown alert into default IC and non-default IC and deploy them on cluster with different annotations

_Fixes #[OSD-16014](https://issues.redhat.com//browse/OSD-16014)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-sre-haproxydown.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-haproxydown.PrometheusRule.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-ingresscontroller-alerts
+    role: alert-rules
+  name: sre-haproxydown-default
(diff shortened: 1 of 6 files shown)


bmeng:
should be promoted to production before https://github.com/openshift/configure-alertmanager-operator/pull/291

hectorakemp (approved):
So from what I can understand:
If the new managed ingress is enabled: 
- We don't apply `deploy/sre-prometheus/legacy-ingress/100-sre-haproxydown.PrometheusRule.yaml` (since it's in the managed ingress folder). 
If it is not enabled:
We apply both 100-sre-haproxydown rules. 
Then we suppress the default https://github.com/openshift/configure-alertmanager-operator/pull/291 for everyone. 
This makes sense to me. 
Great work @bmeng

PR #1740, opened 2023-07-13
OSD-16361: Deprecate custom-domains-operator if legacy-ingress-support label is available 

samanthajayasinghe:
### What type of PR is this?
feature

### What this PR does / why we need it?
Remove  custom-domains-operator resources if legacy-ingress-support label is available.

4.13+ clusters will include `legacy-ingress-support: "false" ` label and CDO, and its resources may not be deployed to new clusters. 

### Which Jira/Github issue(s) this PR fixes?

https://issues.redhat.com/browse/OSD-16361

### Pre-checks (if applicable):
- [x] Tested latest changes against a cluster
- [] Included documentation changes with PR
- [x] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1782, opened 2023-08-03
Revert "OSD-16361: Deprecate custom-domains-operator if legacy-ingress-support label is available "

hectorakemp:
Reverts openshift/managed-cluster-config#1740

After discussion with @dustman9000 and @fahlmant - we are not going to deprecate the operator when `legacy-ingress-support: false`. Instead, we are going to 'deregister' existing custom domains using the operator itself (https://github.com/openshift/custom-domains-operator/pull/120). 

This means that this pull request will need to be reverted.


hectorakemp:
cc @samanthajayasinghe @bmeng

PR #1784, opened 2023-08-04
update infra resize alert

clcollins:
This PR renames the infra resizing expression names to be more in line with what they're actually reporting, and renames the alerts triggered for the same reason.  These alerts will not make it to Primary on-call, because they're warnings, but will be (via a follow-up PR) used to trigger a single InfraNodeNeedsResizingSRE alert that will roll up to Primary.
    
This also extends the CPU `for` threshold from 8 hours to 16 hours, to account for increased "daytime" usage that may be causing some flapping alerts.  The longer threshold will ensure excessive CPU utilization continues for more than just the workday before triggering a "resize required" alert.
    
Finally, this adds an average of the values of the CPU excessive resource consumption, so we receive just one alert, rather than one for every node in the cluster that is currently firing.

Signed-off-by: Chris Collins <collins.christopher@gmail.com>

### DIFF
--- a/deploy/sre-prometheus/100-infra-resizing.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-infra-resizing.PrometheusRule.yaml
@@ -8,26 +8,29 @@ metadata:
   namespace: openshift-monitoring
 spec:
   groups:
-  - name: sre-infra-resizing-recording.rules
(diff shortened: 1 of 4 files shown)


mjlshen (approved):
/lgtm

PR #1806, opened 2023-08-23
Add ListHostedZonesByName and GetHostedZone to HCP managed installer policy

fahlmant:
Classic install has these 2 permission here https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.13/sts_installer_permission_policy.json#L142-L144

These permissions are needed on order to guarantee idempotency for the provisioning flow, in particular the DNS component. 
Currently during provisioning, a hosted zone is created with a CreateHostedZone call. If the provisioning fails later and tries again, this call will fail with an error HostedZoneAlreadyExists.
Unfortunately, this is not enough because in a later step, the NameServers of the new Hosted Zone in the correspondent Hosted Zone on the RH Management account is used to resolve ingress.NameServers are available as part of the DelegationSet property of an HostedZone which is only returned by CreatedHostedZone or GetHostedZone call. As the CreateHostedZone error doesn't return the ID of the existing zone, the ID is retrieved by ListHostedZone call. Then this ID is used to call GetHostedZone and get the name servers. 

ListHostedZonesByName  uses a requested dnsname as a parameter and places matching zones at the top of the list, while ListHostedZones simply returns an unordered list. ListHostedZonesByName allows for a quicker flow, and is functionally the same level of permissions. 

Ref: https://issues.redhat.com/browse/OSD-17774

### DIFF
--- a/resources/sts/4.13/hypershift/sts_hcp_installer_permission_policy.json
+++ b/resources/sts/4.13/hypershift/sts_hcp_installer_permission_policy.json
@@ -22,7 +22,9 @@
                 "elasticloadbalancing:DescribeLoadBalancers",
                 "iam:GetOpenIDConnectProvider",
                 "iam:GetRole",
+                "route53:GetHostedZone",
                 "route53:ListHostedZones",
+                "route53:ListHostedZonesByName",
                 "route53:ListResourceRecordSets",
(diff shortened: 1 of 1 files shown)


rafael-azevedo (approved):
/lgtm

PR #1814, opened 2023-08-28
Remove disable-CPMS CronJob from Management Clusters

anispate:
[OSD-17901](https://issues.redhat.com/browse/OSD-17901)


mjlshen (approved):
Can we change the title to something like "Remove disable-CPMS CronJob from Management Clusters" since strictly speaking this likely will not enable CPMS?

/lgtm
/hold

Feel free to unhold after the PR title change

PR #1815, opened 2023-08-29
Update CPMS state to Active in mgmt cluster

anispate:
[OSD-17901](https://issues.redhat.com/browse/OSD-17901)

As part of the PR, we are activating the control plane machine set by setting the .spec.state field to Active on the mgmt cluster.


mjlshen (approved):
Could you clarify in the PR title/description that we are only doing this for management clusters?

/lgtm
/hold

Feel free to unhold afterwards

PR #1816, opened 2023-08-30
Ensuring that make fails when sss name is uppercase

anispate:
Fixing bug #1815:  Ensuring that the Make fails when the SelectorSyncSet name is uppercase.

Part of [OSD-17901](https://issues.redhat.com/browse/OSD-17901) ticket:


mjlshen (approved):
/lgtm

PR #1819, opened 2023-09-05
OSD-15563 - Update generate script to move generate config to config.yaml

bdematte:
### What type of PR is this?
Refactor

Goal is to improve (by unifying) the flow to submit configuration changes, may it be for HCP (through Policy) or standard OSD (through syncset) 

[OSD-15563](https://issues.redhat.com//browse/OSD-15563)

The goal of the MR is to let the generated templates for each environment untouched as this is a refactor. Especially, this is the best way to ensure the change isn't breaking the existing (as only those templates really matter as for the environment stability). 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bdematte:
/retest
Checking there are no leftovers files (shouldn't as I added an extra cleanup step for the testing)

PR #1823, opened 2023-09-07
fix: custom ingresscontroller OAO alerting

typeid:
### What type of PR is this?
bug

This PR fixes two bugs:

- a bug preventing the alert for a custom IC to trigger.
**Explanation:** the alerting expression is already written to only alert for 5 minutes, with the `for: 30m`, this will never trigger. 

- a bug resulting in faulty resolve notifications.
**Explanation:**  as the alert only lasts for 5 minutes, we would (if the above is fixed) always send the resolution SL after 5 minutes. Even if the alert lasted forever, if the audit exporter pods restart, the cache is lost and the alert resets. Therefore, we cannot have a resolution SL here.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid:
/hold

There's another bug that would get introduced here. Fixing it and pushing.

mjlshen (approved):
/lgtm

PR #1825, opened 2023-09-08
Support enableDebuggingHandlers per default MCP

jewzaam:
### What type of PR is this?
cleanup

Original implementation was for all machines.  The updates to support managing for 'masters' separate from 'workers' and creates new KubeletConfig for each.

Related to https://issues.redhat.com/browse/OSD-16901

Did not open a new JIRA for this work, just piggybacking on existing closed JIRA.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/kubelet-config/disable-debugging-handlers-masters/01-disable-debugging-handlers-masters.KubeletConfig.yaml
+++ b/deploy/kubelet-config/disable-debugging-handlers-masters/01-disable-debugging-handlers-masters.KubeletConfig.yaml
@@ -1,0 +1,11 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: disable-debugging-handlers-masters
+spec:
+  kubeletConfig:
+    enableDebuggingHandlers: false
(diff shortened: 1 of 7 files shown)


deploy/kubelet-config/config.yaml
```diff
@@ -6,6 +6,3 @@ selectorSyncSet:
     - key: hive.openshift.io/version-major-minor
       operator: NotIn
       values: ["4.1", "4.2", "4.3", "4.4", "4.5", "4.6", "4.7", "4.8"]
-    - key: ext-managed.openshift.io/disable-debugging-handlers
```
mjlshen, line 9: What's the significance of removing this label selector? I feel like there's low impact because of the machineConfigPoolSelector, but just want to make sure.

jewzaam: This will always apply to clusters that match, 4.9+, and applies to all nodes.
The config for debugging handlers is in new KubeletConfig unique to the MCP.
MCO renders machineconfig taking all these into account.

mjlshen (approved):
/lgtm

PR #1827, opened 2023-09-10
[WIP] OSD-18411 automate SL notification for KubeNodeUnschedulable alert if customer workloads prevent drain

maorfr:
### What type of PR is this?
feature

this is an attempt to automate SL notification for the KubeNodeUnschedulable alert, in case the alert is firing while customer workloads are preventing the node from draining.

the implementation comes from this SOP: https://github.com/openshift/ops-sop/blob/master/v4/alerts/KubeNodeUnschedulableSRE.md

this discrimination is done by using new metrics:
- `customer_pod_preventing_drain`
- `customer_pdb_preventing_drain`

these metrics do not exist yet, and the plan is to introduce them in https://github.com/openshift/osd-metrics-exporter
(something along the lines of https://github.com/openshift/osd-metrics-exporter/pull/112)

https://issues.redhat.com/browse/OSD-18411

how do you like this idea?

### DIFF
--- a/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
@@ -10,8 +10,28 @@ spec:
   groups:
   - name: sre-node-unschedulable
     rules:
+    - alert: KubeNodeUnschedulableNotificationSRE
+      expr: customer_pod_preventing_drain > 0
+      for: 60m
+      labels:
+        severity: Warning
+        namespace: openshift-monitoring
+        managed_notification_template: "KubeNodeUnschedulableCustomerPodPreventingDrain"
+        send_managed_notification: "true"
+      annotations:
+        message: "The node {{ $labels.node }} has been unschedulable for more than an hour."
+    - alert: KubeNodeUnschedulableNotificationSRE
+      expr: customer_pdb_preventing_drain > 0
+      for: 60m
+      labels:
+        severity: Warning
+        namespace: openshift-monitoring
+        managed_notification_template: "KubeNodeUnschedulableCustomerPDBPreventingDrain"
+        send_managed_notification: "true"
+      annotations:
+        message: "The node {{ $labels.node }} has been unschedulable for more than an hour."
     - alert: KubeNodeUnschedulableSRE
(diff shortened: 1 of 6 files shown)


maorfr:
this needs a RoleBinding for the osd metrics SA to view machine info in openshift-machine-api

deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
```diff
@@ -10,8 +10,28 @@ spec:
   groups:
   - name: sre-node-unschedulable
     rules:
+    - alert: KubeNodeUnschedulableNotificationSRE
```
mrbarge, line 13: One minor nit: we are currently bundling all the OCM-Agent related alerts under `deploy/sre-prometheus/ocm-agent`, could this file please be moved there?

maorfr: i have split up the ocm-agent alerts in https://github.com/openshift/managed-cluster-config/pull/1827/commits/78fad79a21fbec6d3a5573766c225c5a669a1516
does this make sense?

mrbarge: Thanks, that works! 👍  (in reality, there is no difference where they sit, but all the other OCM-Agent related PrometheusRules sit under there so it'll be easier for us to find it in future)

maorfr: understood. thanks!

mrbarge:
From the perspective of the `PrometheusRule` and `ManagedNotification`, this PR is good and would have the desired effect. If/when the `customer_*_preventing_drain` metrics are introduced, this would be fine to onboard into OCM Agent.

One thing to follow on from here is that `KubeNodeUnschedulableSRE` would/should be adapted to only fire for control plane and infra nodes. (assuming we still want to get pinged by the alert in these cases). Wouldn't necessarily need to be part of this PR, but should be the eventual goal.

Secondly, one missing aspect is the inability of the Managednotification to inform the cluster owner of exactly which node & workloads are the problematic ones. If we implement OSD-14130 to introduce some form of templating to OCM Agent `ManagedNotifications` that could populate values based on alert metadata, that would be perfect here.

But consider this a general endorsement for this approach. :+1:

jewzaam:
+1 looks very reasonable.  is lack of templating a blocker?

PR #1833, opened 2023-09-12
[OSD-18400] Add suricata deployment for hypershift.

bergmannf:
This will deploy suricata as a daemonset on every management cluster. Right now only SSH connections will generate alerts.

### What type of PR is this?

feature

Adds suricata as IDS for hypershift management clusters.

Fixes [OSD-18400](https://issues.redhat.com//browse/OSD-18400)

- [X] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-suricata/hypershift-management-cluster/20-osd-suricata-Daemonset.yaml
+++ b/deploy/osd-suricata/hypershift-management-cluster/20-osd-suricata-Daemonset.yaml
@@ -49,7 +49,12 @@ spec:
         kubernetes.io/os: linux
       serviceAccountName: suricata-sa
       tolerations:
-        - operator: Exists
+        - key: "node-role.kubernetes.io/master"
+          operator: "Exists"
+          effect: "NoSchedule"
+        - key: "node-role.kubernetes.io/infra"
+          operator: Exists
+          effect: "NoSchedule"
       volumes:
         - hostPath:
             path: /
--- a/deploy/osd-suricata/hypershift-management-cluster/20-osd-suricata-Daemonset.yaml
+++ b/deploy/osd-suricata/hypershift-management-cluster/20-osd-suricata-Daemonset.yaml
@@ -27,8 +27,8 @@ spec:
               cpu: 100m
               memory: 50Mi
           securityContext:
-            privileged: true
-            runAsUser: 0
+            capabilities:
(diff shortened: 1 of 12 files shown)


bergmannf:
Putting this on hold as the whole IDS scenario still needs discussions, but I am opening up all related PRs so it's easier to discuss the changes.

rhdedgar:
Once this is rebased, it can pull in the new changes from:

https://github.com/openshift/managed-cluster-config/pull/1875
https://github.com/openshift/managed-cluster-config/pull/1876

bergmannf:
@rhdedgar Thanks for the review - I integrated the changes from the other PRs.

Once this has been deployed and works, I was thinking of merging the two deployments into one, so we only have to maintain a single once (assuming the changes made here would be ok for your deployment to fedramp).

deploy/osd-suricata/hypershift-management-cluster/config.yaml
```diff
@@ -0,0 +1,9 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet:
+  matchExpressions:
+    - key: ext-hypershift.openshiftio/cluster-type
+      operator: In
+      values: ["management-cluster"]
+    - key: api.openshift.com/fedramp
+      operator: NotIn
+      values: ["true"]
```
dustman9000, line 9: Can you also add the integration environment?

deploy/osd-suricata/hypershift-management-cluster/20-osd-suricata-Daemonset.yaml
```diff
@@ -0,0 +1,60 @@
...
+          volumeMounts:
+            - mountPath: /host/
+              name: host-filesystem
+      dnsPolicy: ClusterFirst
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/os: linux
+      serviceAccountName: suricata-sa
+      tolerations:
+        - operator: Exists
```
dustman9000, line 52: Can you please add tolerations to avoid the following nodes?
```
node-role.kubernetes.io/master
node-role.kubernetes.io/infra
```
This way, this can be deployed to default worker pools and future request serving nodes.

PR #1836, opened 2023-09-13
Add customer notification on node conditions

jewzaam:
### What type of PR is this?
feature

Adds notifications for node conditions.

https://issues.redhat.com/browse/OSD-18445

Details for alert criteria and notification text are tracked in a doc linked from the card.  Ideally we do _not_ discuss those here but will update the PR once it is clear.  The criteria and text are what keep this WIP at this time.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/sre-prometheus/ocm-agent/100-ocm-agent-node-condition.PrometheusRule.yaml
```diff
@@ -0,0 +1,80 @@
...
+      record: sre:node:condition
+
+  # available conditions: https://kubernetes.io/docs/reference/node/node-status/#condition 
+  # TODO convert to single alert when OCM Agent supports templatized messages: https://issues.redhat.com//browse/OSD-14130
+  - name: sre-node-condition-managed-notification-alerts.alert
+    rules:
+    - alert: NodeConditionDiskPressureNotificationSRE
+      expr: count by (machineset, condition, status)
+            (
+              count by(node) (cluster:cpu_core_node_labels{label_node_role_kubernetes_io_infra="",label_node_role_kubernetes_io_master=""})
```
iamkirkbater, line 32: Will this actually exclude infra nodes?  In my testing for other alerts I had to explicitly exclude infra nodes from the list of worker nodes since they will appear with both labels, however I was accomplishing this a different way so I'm not sure if that applies here. See: https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml#L43-L60

jewzaam: Yea, tested and verified it does exclude infra.  The label `label_node_role_kubernetes_io_infra` is set to `"true"` when that node has the role `infra`.

iamkirkbater: Awesome, thanks for confirming :)

PR #1840, opened 2023-09-14
reduce UpgradeConfigValidationFailedSRE severity from critical to warning

maorfr:
### What type of PR is this?
feature

reduce alert noise. from looking at the alert history, it seems that the alert is either:
- auto resolving
- not actionable
- not up-to-date

is this alert still really needed? i suggest we reduce the severity, at least for now.

https://issues.redhat.com/browse/OSD-18386


cblecker:
/hold
/lgtm cancel

This is a valid alert describing an error condition. The attached card is to fix the logic in MUO to be more forgiving in eventual consistency situations, but this alert can describe an actual error condition too.

mjlshen:
Captured in #1925

/close

PR #1842, opened 2023-09-14
reduce DNSErrors10MinSRE severity from critical to warning

maorfr:
### What type of PR is this?
feature

reduce noise. this alert severity may be reduced because:
- it seems the sentiment is that it is noise
- it is usually accompanied by other alerts
- the metric comes from https://github.com/openshift/managed-prometheus-exporter-dns, which is being [deprecated](https://docs.google.com/document/d/1ypPZ_pd74vV8FUqU86F7CtC3c66ukuQ6p7CgF3P1wNk)

https://issues.redhat.com/browse/SDE-2864

### DIFF
--- a/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-dns-latency.PrometheusRule.yaml
@@ -15,7 +15,7 @@ spec:
       expr: increase(dns_failure_failure_total[10m]) >= 3
       for: 10m
       labels:
-        severity: critical
+        severity: warning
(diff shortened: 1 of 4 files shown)


mjlshen (commented):
/hold

This epic's acceptance criteria is not complete and downgrading the severity was just the initial suggestion, however the discussion in the epic is around defining an SLO around DNS performance to adjust the alert. The alert is not purely noise as it has caught bugs in the past, e.g. OCPBUGS-7000

maorfr:
related to #1872

PR #1854, opened 2023-09-19
Add more OWNERS

maorfr:
With great power comes great responsibility.
With responsibility comes accountability.
With accountability comes ownership.
With ownership comes (oncall) relief.

SRE Principles:
- It’s okay to fail
- Assume positive intent
- Start with trust; extend trust
- Disagree and commit
- Communicate

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,8 +1,10 @@
 reviewers:
+- srep-engineers
 - srep-functional-leads
 - srep-team-leads
 - sre-group-leads
 approvers:
+- srep-engineers
 - srep-functional-leads
 - srep-team-leads
 - sre-group-leads
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,6 +1,41 @@
 # See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
 
 aliases:
+  srep-engineers:
+    - a7vicky
+    - aliceh
+    - anispate
+    - bdematte
+    - bergmannf
+    - boranx
+    - clcollins
+    - feichashao
+    - geowa4
+    - gsleeman
+    - hectorakemp
+    - jwai7
+    - katherinelc321
+    - lnguyen1401
+    - luis-falcon
+    - Makdaam
+    - mbarnes
+    - MitaliBhalla
+    - mrWinston
+    - Nikokolas3270
+    - reedcort
+    - rendhalver
+    - rhdedgar
+    - robotmaxtron
+    - sam-nguyen7
+    - shitaljante
+    - superfunhappytime
+    - Tafhim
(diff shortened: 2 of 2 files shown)


cattias:
lgtm

as a long term solution I think we should consider splitting MCC in smaller components with more precise ownership and combine it with the managed openshift versioning work, getting us to a place where our releasing and testing processes are more efficient and do not require so much scrutiny

ravitri:
+1 Maor and Christophe :+1: Thanks a lot Maor! Will wait for review from others too who would like to comment.

bmeng:
+1 on this

bng0y:
+1

maorfr:
note: this addition only allows merging to this repository, which means - not production.

production promotions are still controlled via app-interface, and this PR is not a proposal to change that gate.

maorfr:
another note: most of this list was collected from https://github.com/openshift/managed-cluster-config/graphs/contributors

wanghaoran1988:
LGTM from my perspective, as said, it's not for production.

sajeelirkal:
+1 lgtm

jmilhau:
I have to concede that opening ownership to more people might introduce errors / mistakes / defects / inefficiencies. I consider this as VERY positive: these situations will be growing moments and learning opportunities that must be seized by any SRE.

Big +1 and I would love to have a retrospective on this change in 6 months.

karthikperu7:
+1  from me

bdematte:
I think we should look into splitting it to components before opening this. Currently, MCC is not testable in staging, and in its current form, revert are often leading to longer production promotion (source of frustration and higher stability risk in general). 
I think once split into component, it would make it easier to support building good understanding of the specific scope to support people growing expectise on the area to be efficient reviewers (and easier to build tests as well). 
Definitively agree on the final goal but I think we could prevent unneeded suffering with some preparation work.

geowa4:
Probably should remove mbarge while we’re in here. 😞

rafael-azevedo:
@maorfr I may be missing something , In reference to  the last point (With ownership comes (oncall) relief). but what specifically about adding this ownership will relieve oncall ? What will this enable towards that effort.

jewzaam:
context on hold: my opinions on this is decision needs to be made within group leads and team leads before merging
once that decision is made, hold can be removed.  I am not specifically blocking this for any other reason and defer to the decision made, just want to be sure we don't merge without that discussion and decision

mjlshen:
I'm a light -1 based on how we currently work, which is reaching out to @managed-cluster-config on Slack, which pings all the OWNERS. This PR would drastically increase the scope of that Slack handle.

If we expect to use the Slack handle less frequently or otherwise change it - e.g. have an understanding that people assigned to the PR are expected to review, then I don't mind doing this.

maorfr:
split region leads to #1858

PR #1857, opened 2023-09-19
Fix SecurityGroup describe actions in HCP support policy

fahlmant:
According to AWS docs [0], all ec2:Describe permissions require a Resoruce of * and are not allowed to use resource specification. Without *, this permission block provided no actual permissions, so changing to * will allow the permissions intended. The permissions should still only affect security groups since the Describe actions are specific to Security Groups and related resources.

[0] https://docs.aws.amazon.com/IAM/latest/UserGuide/troubleshoot_policies.html?icmpid=docs_iam_console#mismatch_action-no-resource

### What type of PR is this?
bug fix

To enable HCP Support managed policy to use SecurityGroup related read actions

https://issues.redhat.com/browse/OSD-18534

- [] Tested latest changes against a cluster

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.13/hypershift/sts_hcp_support_permission_policy.json
+++ b/resources/sts/4.13/hypershift/sts_hcp_support_permission_policy.json
@@ -141,7 +141,7 @@
         "ec2:DescribeSecurityGroups",
         "ec2:DescribeStaleSecurityGroups"
       ],
-      "Resource": "arn:aws:ec2:*:*:security-group*/*"
+      "Resource": "*"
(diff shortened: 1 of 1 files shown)


mjlshen (approved):
Not sure if this was waiting for @jharrington22 's review so I'll put a hold. Feel free to unhold whenever

/lgtm
/hold

jharrington22 (approved):
/lgtm

PR #1858, opened 2023-09-19
add region leads as owners

maorfr:
An addition of SREP region leads as reviewers and approvers, which seems to be well within immediate consensus.

split from #1854

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,8 +1,10 @@
 reviewers:
+- srep-region-leads
 - srep-functional-leads
 - srep-team-leads
 - sre-group-leads
 approvers:
+- srep-region-leads
 - srep-functional-leads
 - srep-team-leads
 - sre-group-leads
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,6 +1,14 @@
 # See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
 
 aliases:
+  srep-region-leads:
+    - geowa4
+    - lnguyen1401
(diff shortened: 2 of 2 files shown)


fahlmant:
As I understand it, the goal is to enable more folks to tune alerts specifically. There are a few places we can alter alerts:
1. configure-alertmanager-operator, where certain things like filtering, changing severity and routing can be done
2. Individual operators sometimes ship their own alerts
3. A few scattered in MCC
For the ones in MCC, I propose an alternative path to achieve the goal of allowing quicker iteration on alerts:
- Move all MCC alerts to a single directory within the `deploy/` directory
- Create an OWNERS file within that new directory that specifies the desired group of reviewers and approvers. 
There are multiple places within MCC that already use this pattern. An example that will allow for a full override of all parent approvers can be found here: https://github.com/openshift/managed-cluster-config/blob/c622e77b958c75290062b6e0d4bed3b54d6d5985/resources/sts/OWNERS#L7C1-L8C25

jewzaam:
+1 to what Taylor said.  this specific broad permission has been discussed in the past and the right answer is the permissions scoped to a directory.  may need to also include `hack/` as all PRs will generate templates.  putting hold for finalizing approach

/hold

rogbas:
+1 to @fahlmant suggestion. Considering the wide scope of MCC and the fact that it touches (in public and fedramp) security, compliance, alerts, SLA and others, giving RLs approval access to a folder containing Alerting/Metrics objects is a good approach to improve RLs autonomy.

cc @dustman9000 @wanghaoran1988

maorfr:
implemented the above in #1867

PR #1863, opened 2023-09-20
deprecate direct deployment method

maorfr:
### What type of PR is this?
cleanup

remove no longer used deployment method - `Direct`

https://issues.redhat.com/browse/OSD-18588

related to:
- https://github.com/openshift/managed-cluster-config/pull/1861
- https://github.com/openshift/managed-cluster-config/pull/1862

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ add the policy as a new SelectorySyncSet.
 
 All resources in `generated_deploy/` are bundled into a template that is used by config management to apply to target "hive" clusters.  The configuration is deployed to the "hive" cluster inside a SelectorSyncSet.
 
-SelectorSyncSet deployment supports resoures that are synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
+SelectorSyncSet deployment supports resources that are synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
 
 * This file is now mandatory in the scope of OSD-15267 and have been added to all folders. In case it is not define, `make` will fail
 ```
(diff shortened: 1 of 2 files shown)


bdematte:
I am generally +1 on this as this is not needed.
I would just suggest to take this opportunity to document in the README how to do the same with standard SSS so that it is clear and easy for future contributors.

maorfr:
the way to do direct deployments is via app-interface. do you think that needs to be here?

bdematte:
> the way to do direct deployments is via app-interface. do you think that needs to be here?

I would say yes (as I am not sure this is widely known and the first idea which would come to my mind would be a SSS in MCC). I think this would ease contribution and reduce risk of having such SSS in MCC in the future if we expect it to be in app-interface.

README.md
```diff
@@ -23,12 +23,9 @@ add the policy as a new SelectorySyncSet.
...
 # Configuration
 
-All resources in `generated_deploy/` are bundled into a template that is used by config management to apply to target "hive" clusters.  The configuration for this supports two options for deployment.  They can be deployed in the template so they are:
+All resources in `generated_deploy/` are bundled into a template that is used by config management to apply to target "hive" clusters.  The configuration is deployed to the "hive" cluster inside a SelectorSyncSet.
 
-1. deployed directly to the "hive" cluster
-2. deployed to the "hive" cluster inside a SelectorSyncSet
-
-Direct deployment (#1) supports resources that are not synced down to OSD clusters.  SelectorSyncSet deployment (#2) supports resoures that _are_ synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
+SelectorSyncSet deployment supports resoures that are synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
```
typeid, line 28:
```suggestion
SelectorSyncSet deployment supports resources that are synced down to OSD clusters.  Each are explained in detail here.  The general configuration is managed in a `config.yaml` file in each deploy directory.  Key things of note:
```

typeid (approved):
Makes sense to me, and I see Benjamin had already +1d this as well.

PR #1867, opened 2023-09-21
add region leads as OWNERS for alerts

maorfr:
An addition of SREP region leads as reviewers and approvers for alerts, which seems to be within immediate consensus.

split from #1858


fahlmant:
I think you'll need
```
options: 
   no_parent_owners: true 
```
in order to make this effective

maorfr:
if i'm reading https://www.kubernetes.dev/docs/guide/owners/#owners correctly, this will exclude top level OWNERS. is that the desired outcome?

PR #1872, opened 2023-09-25
[OSD-18636] Remove python-based operator deployment script and dns prometheusrule

bergmannf:
### What type of PR is this?
Cleanup - remove no longer desired exporter.

Removes the prometheus rule for the dns-exporter.

Also removes the deploy_assets.py python script: that seems to no longer be used, but looks like it was used to deploy the python-based operators in the past.

Fixes [OSD-18636](https://issues.redhat.com//browse/OSD-18636).

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/approve
/lgtm

PR #1877, opened 2023-09-26
Create a specific MHC for management clusters

mjlshen:
### What type of PR is this?
bug

The HyperShift Operator deletes nodes when uninstalling hostedclusters and depends on a MachineHealthCheck to be able to replace an arbitrary number of worker nodes at a time. This PR creates `srep-management-cluster-healthcheck`, which is a copy of `srep-worker-healthcheck` except it changes `.spec.maxUnhealthy: 3` to `.spec.maxUnhealthy: 100%`

[OSD-18661](https://issues.redhat.com//browse/OSD-18661)


mjlshen:
/close

We want to make sure that HyperShift can't just delete machines first

PR #1882, opened 2023-09-28
.spec.unhealthyConditions must have at least one condition

mjlshen:
### What type of PR is this?
bug

This PR resolves syntax errors from #1879 it adds back this unhealthyCondition:

```
  - type:    "Ready"
    timeout: "300s"
    status: "Unknown"
```

as a stopgap - I believe this is not as impactful as we first anticipated https://docs.openshift.com/container-platform/4.13/machine_management/deploying-machine-health-checks.html#machine-health-checks-about_deploying-machine-health-checks

> To limit disruptive impact of the machine deletion, the controller drains and deletes only one node at a time. 

So I think this is still safe...

[OSD-18661](https://issues.redhat.com//browse/OSD-18661)

### DIFF
--- a/deploy/osd-machine-api/management-clusters/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
+++ b/deploy/osd-machine-api/management-clusters/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
@@ -12,10 +12,12 @@ spec:
       - "infra"
       - "master"
     - key: hypershift.openshift.io/request-serving-component
+      operator: In
       values:
       - "true"
-  # An empty unhealthyConditions causes this MHC to execute only if a machine's underlying node is deleted
-  # https://github.com/openshift/machine-api-operator/blob/c11d6227cb4640ce979edd4e9469342274e88910/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L789-L793
-  unhealthyConditions: []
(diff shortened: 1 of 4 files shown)


deploy/osd-machine-api/management-clusters/011-machine-api.srep-worker-healthcheck.MachineHealthCheck.yaml
```diff
@@ -12,10 +12,12 @@ spec:
...
     - key: hypershift.openshift.io/request-serving-component
+      operator: In
       values:
       - "true"
-  # An empty unhealthyConditions causes this MHC to execute only if a machine's underlying node is deleted
-  # https://github.com/openshift/machine-api-operator/blob/c11d6227cb4640ce979edd4e9469342274e88910/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L789-L793
-  unhealthyConditions: []
+  unhealthyConditions:
+  - type:    "Ready"
+    timeout: "300s"
```
jewzaam, line 20: This is when the MHC will consider the node unhealth, so after 5 minutes.  Other non-metal worker MHC is 480s, so this is faster.  What is the rational behind 300s?

mjlshen: No rational - I can change this to 480s

PR #1890, opened 2023-10-04
adds subject permission for fedramp for services we manage in FR only

tonytheleg:
### What type of PR is this?

feature

This PR is to grant the basic permissions SREP should have on any namespaces we manage in OSD/ROSA clusters. Specifically this adds 2 namespaces in FedRAMP only (keycloak and goalert) so that we can manage those services and not trigger a bunch of compliance alerts. It leverages the existing backplane-srep-admins-project clusterole so there is no extra scope creep outside of standard namespaces

For [OSD-18759](https://issues.redhat.com//browse/OSD-18759) and [OSD-18758](https://issues.redhat.com//browse/OSD-18758)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/srep/fedramp/30-srep-fedramp.SubjectPermission.yml
+++ b/deploy/backplane/srep/fedramp/30-srep-fedramp.SubjectPermission.yml
@@ -1,0 +1,15 @@
+apiVersion: managed.openshift.io/v1alpha1
+kind: SubjectPermission
+metadata:
+  name: backplane-srep-fedramp
+  namespace: openshift-rbac-permissions
+spec:
+  permissions:
+  - clusterRoleName: backplane-srep-admins-project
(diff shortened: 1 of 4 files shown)


deploy/backplane/srep/fedramp/30-srep-fedramp.SubjectPermission.yml
mjlshen, line 1: Should we scope these permissions to FedRAMP Hives only? Thinking of https://github.com/openshift/managed-cluster-config/tree/master/deploy/backplane/srep/hive

mjlshen: No, these namespaces are on arbitrary clusters, not on Hive

PR #1895, opened 2023-10-06
feat(OSD-19000): directly use KubePersistentVolumeFillingUp expression

typeid:
### What type of PR is this?
feature

We currently are already automating service logs with OAO for this, but still route it to primary. This PR uses the direct expression(s) from https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/control-plane/prometheus-rule.yaml#L317 and routes them to OAO.

Ref: https://github.com/openshift/configure-alertmanager-operator/pull/298

https://issues.redhat.com/browse/OSD-19000

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
@@ -10,10 +10,45 @@ spec:
   groups:
   - name: sre-managed-notification-alerts
     rules:
-    - alert: LoggingVolumeFillingUpNotificationSRE
+    - alert: LoggingVolumeFillingUpNotificationSRE1m
     # KubePersistentVolumeFillingUp alert firing in openshift-logging Namespace.
-      expr: count(ALERTS{alertname="KubePersistentVolumeFillingUp", alertstate="firing", namespace="openshift-logging"}) >= 1
-      for: 30m
+    # See https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/control-plane/prometheus-rule.yaml#L299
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
```diff
@@ -10,10 +10,45 @@ spec:
...
+        unless on(namespace, persistentvolumeclaim)
+        kube_persistentvolumeclaim_access_mode{namespace="openshift-logging", access_mode="ReadOnlyMany"} == 1
+        unless on(namespace, persistentvolumeclaim)
+        kube_persistentvolumeclaim_labels{namespace="openshift-logging",label_alerts_k8s_io_kube_persistent_volume_filling_up="disabled"} == 1
+      for: 1m
+      labels:
+        severity: Info
+        namespace: openshift-logging
+        send_managed_notification: "true"
+        managed_notification_template: "LoggingVolumeFillingUp"
```
iamkirkbater, line 33: Do we want both of these to fire the same notification?

One being an Info and the other being an Immediate concern leads me to think we should potentially have separate notifications for this.

typeid: As discussed if we want to split it into two notifications, we would have to make sure OAO doesn't send both notifications. The proposed changes should keep the same behavior as before, so it should be fine like this for now.

typeid:
Closed PR as the alert should still fire after the CAMO change, just without the pagerduty receiver.

PR #1898, opened 2023-10-06
Ignore KubeletDown during control plane phase of upgrades

mjlshen:
### What type of PR is this?
bug/feature

KubeletDown intermittently fires and self-resolves during upgrades - we've filed OCPBUGS-20201, but in the meantime we can silence it.

[OSD-19002](https://issues.redhat.com//browse/OSD-19002)


cblecker:
/lgtm

cc @sdodson

PR #1900, opened 2023-10-09
[OSD-18400] Disable more eve types

bergmannf:
### What type of PR is this?
Bug

Disable more types for the eve output of suricata.
It seems that by default it will put far too much information in the eve.json output, we don't care about in hypershift, we only want alerts.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #1907, opened 2023-10-11
Ignore 'MachineOutOfComplianceSRE' alerts during upgrade healthchecks

tnierman:
### What type of PR is this?
_bug_

Adds `MachineOutOfComplianceSRE` to the list of alerts we can safely ignore during upgrade. This alert only indicates that machines older than 28 days are present on the cluster, which doesn't necessarily indicate anything functionally wrong with the platform.


mjlshen (approved):
/unhold
/lgtm

PR #1910, opened 2023-10-14
Allow ocpbugs-20184 cronjob to schedule anywhere

cblecker:
### What type of PR is this?
bug

Updates the cronjob for ocpbugs-20184 as follows:
- Runs every 10 minutes instead of every hour. If things are working properly, the impact will be negligible, but it will fix stalled upgrades quicker.
- Allow it to schedule anywhere, but still prefer infra nodes. This will allow it to find any functional node to schedule to.
- Set resources to help with scheduling.
- Adds pod anti-affinity to prefer that if one pod gets wedged, it'll try to select a different node.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #1916, opened 2023-10-27
Bump infra machinehealthcheck timeout to 10m

tnierman:
### What type of PR is this?
_bug_

Fixes: https://issues.redhat.com/browse/OSD-19449

Increases the `srep-infra-machinehealthcheck` timeout to 600s, to avoid infra node startup delays.

### DIFF
--- a/deploy/osd-machine-api/010-machine-api.srep-infra-healthcheck.MachineHealthCheck.yaml
+++ b/deploy/osd-machine-api/010-machine-api.srep-infra-healthcheck.MachineHealthCheck.yaml
@@ -17,6 +17,6 @@ spec:
     timeout: "480s"
     status: "False"
   - type:    "Ready"
-    timeout: "480s"
+    timeout: "600s"
     status: "Unknown"
   maxUnhealthy: 2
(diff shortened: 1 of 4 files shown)


deploy/osd-machine-api/010-machine-api.srep-infra-healthcheck.MachineHealthCheck.yaml
```diff
@@ -14,9 +14,9 @@ spec:
       operator: Exists
   unhealthyConditions:
   - type:    "Ready"
-    timeout: "480s"
+    timeout: "600s"
```
mjlshen, line 17:
```suggestion
    timeout: "480s"
```
We can keep this at 480s - this controls how fast the MHC controller effectively notices that a node is explicitly NotReady `type: Ready, status: False` and considers it for replacement.

The second one is essentially a timeout how long a machine exists, but is not a node yet.

mjlshen (changes requested):

mjlshen (approved):
/lgtm

PR #1921, opened 2023-11-06
Disable second ingress service log on v4.13+

hectorakemp:
### What type of PR is this?
_(feature)_

There is a request from the BU side to _disable_ the 2nd IC service log automation on version 4.13+, instead of when the new 'managed ingress' feature is enabled.
This pull request makes the SL's `PrometheusRule` align with that idea.

[Discussion](https://redhat-internal.slack.com/archives/CCX9DB894/p1698966021083539) 

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid (approved):
/lgtm

This makes sense to me based on the discussion with BU.

PR #1923, opened 2023-11-06
No more customer kube-node-unscheduleables

iamkirkbater:
Send Customer KubeNodeUnscheduleable Alerts to OCM Agent Operator and ensure infra/control plane alerts are handled.

I've kept the original KubeNodeUnscheduleable alert around for now, but it should fire 1m after the ControlPlane version of this alert, so that we get the Control Plane version first and the KubeNodeUnscheduleable gets grouped into that PD alert - as we gain confidence that this alert is not missing anything then we can remove the original KubeNodeUnscheuldeable.

### What type of PR is this?

Feature

Send Customer KubeNodeUnscheduleable Alerts to OCM Agent Operator and ensure infra/control plane alerts are handled.

_Fixes [OSD-18411](https://issues.redhat.com/browse/OSD-18411)_

I've kept the original KubeNodeUnscheduleable alert around for now, but it should fire 1m after the ControlPlane version of this alert, so that we get the Control Plane version first and the KubeNodeUnscheduleable gets grouped into that PD alert - as we gain confidence that this alert is not missing anything then we can remove the original KubeNodeUnscheuldeable.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -60,7 +60,7 @@ spec:
       summary: Worker node is predicted to run out of inodes in 4 hours
     - activeBody: |-
         Your cluster requires you to take action. A workload is been preventing a machine from deleting. The name of the workload can be found by looking at the 'pods_preventing_node_drain' metric in Prometheus. Not taking action can result in higher compute costs as the instance won't terminate until the workload has been moved.
-      name: CustomerWorkloadPrecventingDrain
+      name: CustomerWorkloadPreventingDrain
       resendWait: 1
       severity: Error
       summary: Workload preventing machine deletion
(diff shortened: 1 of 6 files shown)


deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
```diff
@@ -58,3 +58,9 @@ spec:
       resendWait: 24
       severity: Info
       summary: Worker node is predicted to run out of inodes in 4 hours
+    - activeBody: |-
+        Your cluster requires you to take action. A workload is been preventing a machine from deleting. The name of the workload can be found by looking at the 'pods_preventing_node_drain' metric in Prometheus. Not taking action can result in higher compute costs as the instance won't terminate until the workload has been moved.
+      name: CustomerWorkloadPrecventingDrain
```
mjlshen, line 63:
```suggestion
      name: CustomerWorkloadPreventingDrain
```

PR #1930, opened 2023-11-15
Restrict iam:PassRole for classic 4.14

mjlshen:
### What type of PR is this?
feature

Minimizes the scope of how the installer and MAPI are allowed to use `iam:PassRole`:

* Installer: allow passing the `ControlPlane-Role` and `Worker-Role` to `ec2.amazonaws.com`
* MAPI: allow passing roles to`ec2.amazonaws.com`

[OSD-19684](https://issues.redhat.com//browse/OSD-19684)

### DIFF
--- a/resources/sts/4.14/openshift_machine_api_aws_cloud_credentials_policy.json
+++ b/resources/sts/4.14/openshift_machine_api_aws_cloud_credentials_policy.json
@@ -2,6 +2,21 @@
   "Version": "2012-10-17",
   "Statement": [
     {
+      "Sid": "PassRoleToNodes",
+      "Action": [
+        "iam:PassRole"
+      ],
+      "Effect": "Allow",
+      "Resource": "*",
+      "Condition": {
+        "StringEquals": {
+          "iam:PassedToService": [
+            "ec2.amazonaws.com"
+          ]
+        }
+      }
+    },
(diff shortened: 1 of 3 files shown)


resources/sts/4.14/openshift_machine_api_aws_cloud_credentials_policy.json
```diff
@@ -1,6 +1,21 @@
 {
   "Version": "2012-10-17",
   "Statement": [
+    {
+      "Sid": "PassRoleToNodes",
+      "Action": [
+        "iam:PassRole"
+      ],
+      "Effect": "Allow",
+      "Resource": "*",
```
mjlshen, line 10: I updated this PR to relax the resource constraint on MAPI's managed policy due to existing clusters where the `Worker` and `Control-Plane` account roles do not follow the expected naming scheme (see card for more details).

The installer managed policies can continue to be restricted in this way as the permissions aren't relevant during an uninstall of an existing cluster and can be safely updated.

jharrington22:
@mjlshen have you tested this? I believe the installer will need PassRole for the bootstrap and control plane nodes and MAPI will need it for the workers.

mjlshen:
There's currently no path forward

/close

PR #1931, opened 2023-11-15
update to account for AWS ap regions

mjlshen:
### What type of PR is this?
bug

https://github.com/openshift/cluster-image-registry-operator/blob/release-4.14/pkg/storage/util/util.go#L87-L134

This PR accounts for how the image-registry operator names buckets in
AWS, which is:

`${cluster_id}-image-registry-${aws_region}-${suffix}`

* `${cluster_id}` is 32 characters
* -image-registry- is 16 characters
* ap-northeast-1 is 14 characters
* The image-registry operator truncates the name to 62 characters

which leaves us with no suffix for these kinds of regions.

[OSD-19694](https://issues.redhat.com//browse/OSD-19694)


rafael-azevedo (approved):
/lgtm

PR #1944, opened 2023-11-23
OSD-19183: add sre-cluster-version-operator PrometheusRule

Tof1973:
### What type of PR is this?
feature

It will create new  ClusterOperatorDegraded alerts for kube-* operators when the events last 1day, with special need_attention label to specify that this warning alert should not be deleted as it rquires atten

[OSD-19183](https://issues.redhat.com//browse/OSD-19183)

As mentioned in [OSD-18336 study](https://issues.redhat.com/browse/OSD-18336), activating those alerts when issue last more than 24h should raise in average 1 alert per week, that we'd better look at.

### DIFF
--- a/deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
@@ -1,0 +1,31 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-cluster-version-operator
+    role: alert-rules
+  name: sre-cluster-version-operator
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-cluster-version-operator
+    rules:
+    - alert: ClusterOperatorDegradedSRE
+      annotations:
+        summary: Cluster operator has been degraded for 1 day.
+        description: The {{ "{{ $labels.name }}" }} operator is degraded because {{ "{{ $labels.reason }}" }}, and the components it manages may have reduced quality of service.  Cluster upgrades may not complete. For more information refer to 'oc get -o yaml clusteroperator {{ "{{ $labels.name }}" }}'{{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} or {{ label \"url\" (first $console_url ) }}/settings/cluster/{{ end }}{{ end }}" }}.
+      expr: |
+        max by (namespace, name, reason)
+        (
+          (
+            cluster_operator_conditions{job="cluster-version-operator", condition="Degraded", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"}
+            or on (namespace, name)
+            group by (namespace, name) (cluster_operator_up{job="cluster-version-operator", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"})
+          ) == 1
+        )
+      for: 1d
+      labels:
+        link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/ClusterOperatorDegraded.md"
+        namespace: openshift-monitoring
+        needs_attention: "true"
+        severity: warning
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,30 @@
...
+  labels:
+    prometheus: sre-cluster-version-operator
+    role: alert-rules
+  name: sre-cluster-version-operator
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-cluster-version-operator
+    rules:
+    - alert: ClusterOperatorDegraded
```
typeid, line 13: Will this alert name cause a collision? Can we have multiple alerts defined with the same alert name?

Generally, when SRE re-create an alert we use the pattern `AlertNameSRE` to make it clear it's not an upstream alert. I would suggest renaming it to `ClusterOperatorDegradedSRE` to align with the rest.

deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,30 @@
...
+          (
+            cluster_operator_conditions{job="cluster-version-operator", condition="Degraded", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"}
+            or on (namespace, name)
+            group by (namespace, name) (cluster_operator_up{job="cluster-version-operator", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"})
+          ) == 1
+        )
+      for: 1d
+      labels:
+        severity: warning
+        needs_attention: "true"
```
typeid, line 29: We should definitely document this label in ops-sop, in case this is not something you're tracking in your epic yet.

typeid (changes requested):

typeid:
@Tof1973 Could you re-clarify in this PR how many new primary paging alerts we expect to be getting off this (e.g. 1-2 a day?)?

deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
Tof1973: https://github.com/openshift/ops-sop/pull/2947

deploy/sre-prometheus/100-cluster-version-operator.PrometheusRule.yaml
```diff
@@ -0,0 +1,31 @@
...
+            cluster_operator_conditions{job="cluster-version-operator", condition="Degraded", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"}
+            or on (namespace, name)
+            group by (namespace, name) (cluster_operator_up{job="cluster-version-operator", name =~ "kube-apiserver|kube-controller-manager|kube-scheduler"})
+          ) == 1
+        )
+      for: 1d
+      labels:
+        link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/ClusterOperatorDegraded.md"
+        namespace: openshift-monitoring
+        needs_attention: "true"
```
mjlshen, line 30: What does `needs_attention: true` mean in practice?

Tof1973: In practice, it means that this is a warning alert, that should not be deleted and should raise an incident on PD.
This has to be used for alerts that we delete but that we want to look at after some time without considering them as critical, as it will put the cluster in danger in the future...
The none deletion, will be handled by https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/90712, and the documentation by https://github.com/openshift/ops-sop/pull/2947

mjlshen: Thanks for the info. I'm not really seeing the value though - why not just make it critical?

I don't know if we want Primary having to make decisions on how important a page is. Either we should respond to it ASAP or it's something that we can ignore until later?

mjlshen: I'm pushing back because in the past, warning alerts fired and were visible to primary, but we felt that primary was overloaded by the existing set of critical alerts. If primary is expected to look at something it should be critical.

We could have a warning alert out there to collect data around the impact of this alert getting promoted to a critical one for example. Otherwise, I think this kind of "more important warning" alert should go through wider discussion among SREs. E.g. if I'm busy can I silence it? Does it count against the 3.5 alerts/hour SLO?

PR #1945, opened 2023-11-23
Make podman compatible with MacOS

andreadecorte:
Seting the SELinux mount label when running on podman for macOS make the run fails with an error. So we don't set it.

Similar to https://github.com/openshift/release/pull/35997

### What type of PR is this?
bug

Fixes error
```
Error: preparing container 44309ffd66e127508322ff6634ec3acd17be9bcd735210b54bd81b5e7e139ba2 for attach: lsetxattr /Users/adecorte/code/managed-cluster-config/.git: operation not supported
```
when running on MacOS

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,9 +36,16 @@ ifndef GEN_CMO_CONFIG
 $(error GEN_CMO_CONFIG is not set; check project.mk file)
 endif
 
-
+VOLUME_MOUNT_FLAGS = :z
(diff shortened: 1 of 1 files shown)


typeid (approved):
Had looked at this in the past and didn't understand how it could work for mac. I guess the answer is it didn't yet :) 

Thanks Andrea!

/lgtm
/approve

PR #1946, opened 2023-11-27
formatting conditions less permissive as requested

rafael-azevedo:
Changes registry operator permissions in AWS to be more restrictive regex as requested by AWS .


mjlshen (approved):
/lgtm

PR #1961, opened 2023-12-16
The Policy will be non-compliant until the default IngressController resource appears

jnpacker:
### What type of PR is this?
bug (regression) due to performance improvements.

* Allows the Policy to avoid a 2hr reconcile when the default IngressController resource is missing
* This was accomplished, by trying to create an invalid resource when the default IngressController is missing
* Even if there was a Policy looking for the default IngressController, it would show as non-compliant
* Reconcile is 45s when non-compliant, so there should not be a dramatic affect on provisioning

_Fixes #_ https://issues.redhat.com/browse/OCM-5315

Make sure TLS is online and the rosa-ingress-certificate-policies and rosa-ingress-certificate-check policies are not return a large number of messages when non-compliant.

- [ ] ~Included documentation changes with PR~
- [ ] ~If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:~

### DIFF
--- a/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
@@ -43,6 +43,14 @@ spec:
                               scope: '{{hub- if eq (fromConfigMap "openshift-acm-policies" .ManagedClusterName "endpoint-publishing-strategy") "internal" -hub}} Internal {{hub- else -hub}} External {{hub- end -hub}}'
                           {{hub- end hub}}
                     {{- end }}
+                    {{- if ne (lookup "operator.openshift.io/v1" "IngressController" "openshift-ingress-operator" "default").metadata.name "default" }}
+                    - complianceType: musthave
+                      metadataComplianceType: musthave
+                      objectDefinition:
+                        kind: Namespace
+                        metadata:
+                          name: donotcreateme?
+                    {{- end }}
                 pruneObjectBehavior: None
                 remediationAction: enforce
                 severity: low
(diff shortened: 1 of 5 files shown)


deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
```diff
@@ -43,6 +43,14 @@ spec:
                               scope: '{{hub- if eq (fromConfigMap "openshift-acm-policies" .ManagedClusterName "endpoint-publishing-strategy") "internal" -hub}} Internal {{hub- else -hub}} External {{hub- end -hub}}'
                           {{hub- end hub}}
                     {{- end }}
+                    {{- if ne (lookup "operator.openshift.io/v1" "IngressController" "openshift-ingress-operator" "default").metadata.name "default" }}
+                    - complianceType: musthave
+                      metadataComplianceType: musthave
+                      objectDefinition:
+                        kind: Namespace
+                        metadata:
+                          name: donotcreateme?
```
andreadecorte, line 52: '?' is not a valid character for the name

jnpacker: That was the point. We want something that is guaranteed to fail to be applied, which marks the Policy as non-compliant, then when the default IngressController resource is created, this section is NO longer applied be of the IF on line 46.

This seemed safer, then trying to engage a webHook that would for example try to block creating a new ClusterVersion kind resource.

jnpacker: If your okay with that, I can ask it to be reviewed and pushed to integration to test.

jnpacker:
GRC posted a new PolicyGenerator binary 1.12.4, this correctly generates the extraDependencies in both places.

The following PR supersedes this one.
https://github.com/openshift/managed-cluster-config/pull/1964

PR #1962, opened 2023-12-18
[OSD-19769] Add ClusterMonitoringErrorBudgetBurnSRE alert

zmird-r:
### What type of PR is this?
feature

This PR aims to replace the ClusterOperatorDown alert for the cluster monitoring operator. More information on the ticket https://issues.redhat.com/browse/OSD-19769

The silence for the ClusterOperatorDown is being added on CAMO with this PR: https://github.com/openshift/configure-alertmanager-operator/pull/312. To merge after this one.

### DIFF
--- a/deploy/sre-prometheus/monitoring/100-cluster-monitoring-error-budget-burn.yaml
+++ b/deploy/sre-prometheus/monitoring/100-cluster-monitoring-error-budget-burn.yaml
@@ -1,0 +1,73 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-cluster-monitoring-error-budget-burn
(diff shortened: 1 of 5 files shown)


typeid (changes requested):
If we set the SLO at 99% we will see no benefit in a budget burn alert. It will likely even cause more alerts.
Please correct me if I'm wrong. I think the intention here is to allow flappyness, thus we should only alert if the operator is down for a high percentage of the time.

zmird-r:
We can always adjust the SLO later; the idea is to start high and maybe have it too noisy rather than start too low and delay it too much. Do you have a suggestion for a starting point?

typeid:
After rethinking, this should be fine. Let's start with 0.99 SLO and move from there. 

/lgtm

typeid (approved):

PR #1963, opened 2023-12-18
Update LPRSE permissions for ACS

robshelly:
### What type of PR is this?
_feature_

Update of LPSRE permissions for ACS. We have reviewed the permissions we will need using for ACS once the addon is rolled out to prodcution clusters and updated accordingly. 

https://issues.redhat.com/browse/MTSRE-1564

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/lpsre/acs/01-acs-lpsre-admins.SubjectPermission.yaml
+++ b/deploy/backplane/lpsre/acs/01-acs-lpsre-admins.SubjectPermission.yaml
@@ -8,7 +8,7 @@ spec:
   - clusterRoleName: view
     namespacesAllowedRegex: (^redhat-acs-fleetshard$|^rhacs$|^rhacs-.*)
     namespacesDeniedRegex: openshift-backplane-cluster-admin
-  - clusterRoleName: backplane-lpsre-acm-admins-project
+  - clusterRoleName: backplane-acs-lpsre-admins-project
     namespacesAllowedRegex: (^redhat-acs-fleetshard$|^rhacs$|^rhacs-.*)
     namespacesDeniedRegex: openshift-backplane-cluster-admin
   subjectKind: Group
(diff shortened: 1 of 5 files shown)


Nanyte25:
@jewzaam and @NautiluX should we create a new PR of can you review?

jewzaam:
@Nanyte25 easiest to not rename the clusterrole..

Nanyte25:
@jewzaam Can you confirm leaving the previous cluster-role name will not impact functionality going forward?

Nanyte25:
@jewzaam There is no issue as the current backplane access is functioning as expected, just wanted confirmation @robshelly please updated PR.

Nanyte25:
@jewzaam does that mean we can leave the change in the SubjectPermissions and except the @robshelly PR or should we submit a new PR?

jewzaam:
@Nanyte25 my goal is to make sure we don't have clusterrolebindings or rolebindings orphaned when changes to subject permisisons are made.  The easiest way to check is to just deploy the change to a test cluster and observe if there are any clusterrolebindings or rolebindings that are not expected.  I've linked the bug for context in a prior comment.

Ajpantuso:
> @Nanyte25 my goal is to make sure we don't have clusterrolebindings or rolebindings orphaned when changes to subject permisisons are made. The easiest way to check is to just deploy the change to a test cluster and observe if there are any clusterrolebindings or rolebindings that are not expected. I've linked the bug for context in a prior comment.

@jewzaam The ACS addon is not currently installed on any clusters hence the [config](https://github.com/openshift/managed-cluster-config/blob/8492c71b09a779266ab39bc3d5cb28ad7db3e0a2/deploy/backplane/lpsre/acs/config.yaml) prevents the existing `SubjectPermissions` from being deployed anywhere. So while a test using a cluster with the ACS addon installed would orphan bindings post change this scenario will not occur in reality. As long as we deploy this prior to use of the ACS addon there should be no impact.

jewzaam:
Fair enough, confirmed addon isn't in use so this configuration isn't applied anywhere at this time and changing the config in this PR is a noop.

/approve

Nanyte25:
Can we get an `LGTM` added here thanks @jewzaam and @Ajpantuso

PR #1964, opened 2023-12-18
OCM-5315 & ACM-9127 Policy will be non-compliant until the default IngressController resource appears

jnpacker:
### What type of PR is this?
**bug** (regression) due to performance improvements.

* Allows the Policy to avoid a 2hr reconcile when the default IngressController resource is missing
* This was accomplished, by creating a new dependency that looks for the default IngressController resource
* We then put an extraDependencies check on the rosa-ingress-certificate-policies Policy so that it doesn't run until the resource exists
* Reconcile is 45s when non-compliant (default ingressController resource is missing), so there should not be a dramatic affect on provisioning

Fixes # https://issues.redhat.com/browse/OCM-5315

Make sure TLS is online and the rosa-ingress-certificate-policies and rosa-ingress-certificate-check policies are not return a large number of messages when non-compliant.

- [ ] ~Included documentation changes with PR~
- [ ] ~If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ ifndef GEN_CMO_CONFIG
 $(error GEN_CMO_CONFIG is not set; check project.mk file)
 endif
 
+POLICYGEN_VERSION=v1.12.3
+
 VOLUME_MOUNT_FLAGS = :z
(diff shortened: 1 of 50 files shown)


mjlshen (approved):
/approve

PR #1965, opened 2023-12-19
remove unneeded kms:DescribeKey permission from 4.14+

yithian:
### What type of PR is this?
_feature_

The _Creating a ROSA cluster with STS using customizations_ doc [0] instructs customers to grant the necessary permissions (including kms:DescribeKey) in the key policy for the key they want to use. Since this permission is already granted in the key policy, we don't need it granted in the control plane policy as well.

This also allows us to limit the privileges we have, as we don't get access to "*" (all kms keys) and don't require the customer to do any additional tagging of their keys.

AWS-managed keys will not be affected by this, as all principals in the account that are authorized to use EC2 have permission to use AWS-managed keys.

[0]
https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa-sts-creating-a-cluster-with-customizations.html#rosa-sts-creating-cluster-customizations-cli_rosa-sts-creating-a-cluster-with-customizations

[OSD-19640](https://issues.redhat.com/browse/OSD-19640)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant:
/hold
Needs further discussion. The permission is listed by both the old and new AWS cloud controller code bases that OCP uses, and we need to match that. Before we remove it, we need to consult with the OCP teams to see if this really is safe to remove, and document that.

yithian:
> Note: For a role to get be able to get a key, both the role and the key policy need to grant each other. Just relying on the key policy won't work.

The AWS docs seem to suggest otherwise. From the [Authentication and access control doc](https://docs.aws.amazon.com/kms/latest/developerguide/control-access.html):

> Key policy – Every KMS key has a key policy. It is the primary mechanism for controlling access to a KMS key. **You can use the key policy alone to control access**, which means the full scope of access to the KMS key is defined in a single document (the key policy). For more information about using key policies, see [Key policies](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html).

And from the [Key Policies doc](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html):

> No AWS principal, including the account root user or key creator, has any permissions to a KMS key unless they are explicitly allowed, and never denied, in a key policy, IAM policy, or grant.

Am I misinterpreting these statements?

fahlmant:
The key policy controls access, BUT the Role needs to still have the actions defined. If a role doesn't have DescribeKey, then it can't Describe any keys.
```
An error occurred (AccessDeniedException) when calling the DescribeKey operation: User: arn:aws:sts::<replaced>:assumed-role/test-role/tfahlman is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:us-east-1:<replaced>:key/<replaced> because no identity-based policy allows the kms:DescribeKey action
```
This key gives access to this role to use it, but the role doesn't have a policy with kms:DescribeKey attached.

yithian:
That has not been my experience...

I created a test user in my dev account which has no permissions policy at all
image

I created a new test key which grants the test user the permissions to use the key
image

`
$ aws kms describe-key --key-id 90a7c589-8590-4f4c-9ef2-adf5c33d42f3
{
    "KeyMetadata": {
        "AWSAccountId": "714359388950",
        "KeyId": "90a7c589-8590-4f4c-9ef2-adf5c33d42f3",
        "Arn": "arn:aws:kms:us-east-1:714359388950:key/90a7c589-8590-4f4c-9ef2-adf5c33d42f3",
        "CreationDate": 1702995498.685,
        "Enabled": true,
        "Description": "",
        "KeyUsage": "ENCRYPT_DECRYPT",
        "KeyState": "Enabled",
        "Origin": "AWS_KMS",
        "KeyManager": "CUSTOMER",
        "CustomerMasterKeySpec": "SYMMETRIC_DEFAULT",
        "KeySpec": "SYMMETRIC_DEFAULT",
        "EncryptionAlgorithms": [
            "SYMMETRIC_DEFAULT"
        ],
        "MultiRegion": false
    }
}
$ aws sts get-caller-identity 
{
    "UserId": "AIDA2MUZDHMLLHR6RQFYC",
    "Account": "714359388950",
    "Arn": "arn:aws:iam::714359388950:user/testuser"
}
`

Is requiring a policy that grants permission to use `kms:DescribeKey` in addition to requiring the key policy to grant that permission unique to roles? Because it doesn't seem to be a requirement for users, at least in my testing. Am I completely misunderstanding something?

fahlmant:
I haven't tested with users, just roles, as that's how this policy is consumed and used in OCP.

yithian:
I created a test role with no attached or inline policies and gave my test user the ability to assume the role
image
image

I then updated the key policy to grant use permissions to the test role
image

I was still able to describe the key using just a key policy

```
$ aws sts assume-role --role-arn arn:aws:iam::<redacted>:role/testrole --role-session-name test-session
{
    "Credentials": {
        "AccessKeyId": "<redacted>",
        "SecretAccessKey": "<redacted>",
        "SessionToken": "<redacted>",
        "Expiration": "2023-12-19T15:40:52Z"
    },
    "AssumedRoleUser": {
        "AssumedRoleId": "<redacted>:test-session",
        "Arn": "arn:aws:sts::<redacted>:assumed-role/testrole/test-session"
    }
}

$ aws sts get-caller-identity 
{
    "UserId": "<redacted>:test-session",
    "Account": "<redacted>",
    "Arn": "arn:aws:sts::<redacted>:assumed-role/testrole/test-session"
}

$ aws kms describe-key --key-id 90a7c589-8590-4f4c-9ef2-adf5c33d42f3
{
    "KeyMetadata": {
        "AWSAccountId": "<redacted>",
        "KeyId": "90a7c589-8590-4f4c-9ef2-adf5c33d42f3",
        "Arn": "arn:aws:kms:us-east-1:<redacted>:key/90a7c589-8590-4f4c-9ef2-adf5c33d42f3",
        "CreationDate": 1702995498.685,
        "Enabled": true,
        "Description": "",
        "KeyUsage": "ENCRYPT_DECRYPT",
        "KeyState": "Enabled",
        "Origin": "AWS_KMS",
        "KeyManager": "CUSTOMER",
        "CustomerMasterKeySpec": "SYMMETRIC_DEFAULT",
        "KeySpec": "SYMMETRIC_DEFAULT",
        "EncryptionAlgorithms": [
            "SYMMETRIC_DEFAULT"
        ],
        "MultiRegion": false
    }
}
```

Can you confirm that the user/role you tested with was granted use rights to use the key you tested against? I'm not able to replicate that error

fahlmant:
I can't seem to get it to work on my side, but it looks like it works for you. In that case, then next step is to validate with the OCP cloud team if this permission can be removed from the AWS cloud controllers

yithian:
That seems like it's a different question though, right? OSD-19640 is about reducing the scope of kms:DescribeKey permissions required (and this PR does that by relying on kms:DescribeKey being granted via the key policy of the customer-managed KMS key)

Is getting the AWS cloud controllers not to need kms:DescribeKey at all an above-and-beyond thing?

fahlmant:
We depend on the list of permissions from upstream, usually via Credentials Requests. Scoping permissions is okay, but removing permissions that upstream lists is a larger risk. If we're different from Credentials Requests because a permission is seemingly not used, then we need to validate that with upstream.

resources/sts/4.14/sts_instance_controlplane_permission_policy.json
```diff
@@ -44,6 +44,17 @@
                 "kms:DescribeKey"
```
mjlshen, line 44: I think we can remove the permission from here

mjlshen (changes requested):

resources/sts/4.14/sts_instance_controlplane_permission_policy.json
yithian: Ugh, force-pushed a fix

mjlshen (commented):
/lgtm

resources/sts/4.14/sts_instance_controlplane_permission_policy.json
mjlshen, line 1: Could we move these changes to the 4.15 folder? Changes still look good to me, but the 4.15 policies exist now :D

yithian:
@fahlmant does this need more discussion or any other changes before removing the hold?

yithian:
@dustman9000 since Taylor is out until mid-February, do you have any thoughts on if/how we can proceed?

mjlshen:
/unhold

The original hold was in place because the permission was being removed. This is no longer the case

PR #1967, opened 2023-12-20
ocm-agent: Fix wrong alert label

RaphaelBut:
### What type of PR is this?
bug

This alert has never fired because of the wrong label. We need the alert to work as ocm-agent is supporting templates now.

Context: Found this when doing doc updates for ocm-agent for https://issues.redhat.com/browse/OSD-19692 
Corresponding updates for troubleshooting in https://github.com/openshift/ops-sop/pull/2968 

_Fixes #_
https://issues.redhat.com/browse/OSD-20080 

- [x] Included documentation changes with PR: Added corresponding changes in ops-sop

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
@@ -12,7 +12,7 @@ spec:
     rules:
     - alert: OCMAgentResponseFailureServiceLogsSRE
       # Alert if the OCM agent response failure metric has been set for an hour average window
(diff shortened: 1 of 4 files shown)


Nikokolas3270:
/lgtm
Thanks for the fix

RaphaelBut:
/hold 
@ravitri fyi: there is a bit of a risk here of firing lots of alerts to primary as we are basically enabling the alert for the first time. While I dont think that the new changes in ocm-agent will cause alerts, I am not sure how many unrelated issues this alert may have found in the past. Those may now surface all at once, causing many alerts.

RaphaelBut:
/unhold 
Good to merge.

PR #1971, opened 2023-12-25
[OSD-20168] remove red-hat-managed tag condition from permission to create ec2 SecurityGroup in HCP

maorfr:
### What type of PR is this?
bug

without the removal of this condition, customer can only create `Service` of type `LoadBalancer` if it has this tag (`red-hat-managed: "true"`).

IIUC, this eventually lands in the `ROSAKubeControllerPolicy` managed policy.

https://issues.redhat.com/browse/OSD-20168


fahlmant:
/hold
This cannot be changed. AWS requires this condition for Load Balancers in their managed policies.

PR #1990, opened 2024-01-15
OCM-2303 | [HCP] Disable ingress certificate in ACM policy if configmap parameter is true

andreadecorte:
Ingress in HCP are configured by an ACM policy and configuration are passed in a configmap written by OCM.

In local environment, certificates are not generated, so specifying them in the ingress config make the operator degraded, as the relevant secret does not exist. To fix this, we added a new parameter in the CM to skip certificates.

With this change, the policy will read the parameter and exclude any certificate related config/object if it is true. If it is false or not present, we will behave as before (so reference the certificates). So backward compatibility is guaranteed. No impact expected in standard integration/staging/production clusters.

Tested the changes in an integration Service Cluster.

CC @cdoan1 for awareness

### What type of PR is this?
Bug fixing for local development environment

Allows having compliant policies and proper setup when certificate are not available

[OCM-2303](https://issues.redhat.com//browse/OCM-2303)

To be monitored that default ingress controller gets certificates in integration/staging.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
@@ -38,8 +38,10 @@ spec:
                           annotations:
                             ingress.operator.openshift.io/auto-delete-load-balancer: 'true'
                         spec:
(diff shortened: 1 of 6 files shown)


mjlshen (approved):
/approve

PR #1992, opened 2024-01-16
OSD-18436 : Workaround for bug 16655: CreateContainerError due to seccomp error

mrWinston:

- Deploys a CronJob to clusters affected by https://issues.redhat.com/browse/OCPBUGS-16655
- That cronjob checks if nodes are affected by the bug
- If so, the machines are deleted and the node is recreated
- Also, an SL is sent to the CU

However, some restrictions apply:
- Only infra nodes are recreated
- SL is only sent if there hasn't been an SL sent in the last 24 hours (by this cronjob)

Here's a rough outline over the steps in the script:
1. Check namespaces in `$CHECKED_NAMESPACES` for failing pods
2. Check if those pods fail with a seccomp error and save the nodes those pods are running on
3. For every node identified, Check if it's an infra node.
4. If so, get the machine CR and delete it
5. wait for the new machine to be ready and the new node to become ready
6. After all nodes have been handled, a servicelog is sent and the current timestamp is saved in a configmap annotation. This annotation is then used to check if we already sent a servicelog today.

_Fixes [OSD-18436](https://issues.redhat.com//browse/OSD-18436)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocpbugs-15043/10-ocpbugs-15043.CronJob.yaml
+++ b/deploy/ocpbugs-15043/10-ocpbugs-15043.CronJob.yaml
@@ -1,0 +1,59 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: ocpbugs-15043
+  namespace: openshift-config
+spec:
+  failedJobsHistoryLimit: 2
+  successfulJobsHistoryLimit: 2
+  concurrencyPolicy: Forbid
+  schedule: "*/5 * * * *"
+  jobTemplate:
+    spec:
+      ttlSecondsAfterFinished: 3600
+      template:
+        metadata:
+          labels:
+            app: ocpbugs-15043
+        spec:
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - preference:
+                  matchExpressions:
+                  - key: node-role.kubernetes.io/master
+                    operator: Exists
+                weight: 1
+          tolerations:
+            - effect: NoSchedule
+              operator: Exists
+          serviceAccountName: ocpbugs-15043
+          restartPolicy: Never
+          volumes:
+            - name: scriptconfigmap
+              configMap:
+                name: ocpbugs-15043-script
+                items:
+                  - key: ocpbugs-15043.sh
+                    path: script
+            - name: pull-secret
+              secret:
+                secretName: pull-secret
+          containers:
+          - name: ocpbugs-15043
+            image: image-registry.openshift-image-registry.svc:5000/openshift/tools:latest
+            imagePullPolicy: Always
+            resources:
+              requests:
+                cpu: 100m
+                memory: 100Mi
+              limits:
+                cpu: 100m
+                memory: 100Mi
+            command:
+            - /bin/bash
+            - -c
+            - |
+                  chmod +x /etc/config/script/ocpbugs-15043.sh
+                  /etc/config/script/ocpbugs-15043.sh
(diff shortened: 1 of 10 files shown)


deploy/ocpbugs-15043/script.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo "Found node '$node'. Checking associated machine cr"
+
+    MACHINE="$(jq -r '.metadata.annotations."machine.openshift.io/machine"' <<< "$nodeJson" | cut -d '/' -f2)"
+    if [[ $MACHINE = "" ]]; then
+      echo "Couldn't find Machine for node $node. Skipping"
+      continue
+    fi
+
+    echo "Found machine '$MACHINE' for node '$node'. Removing it..."
+    oc -n "$MACHINE_API_NAMESPACE" delete machine "$MACHINE" --wait
```
bng0y, line 63: should this not delete a machine already in a delete state?

deploy/ocpbugs-15043/script.sh
```diff
@@ -0,0 +1,136 @@
...
+    echo "Machine removed successfully. Waiting for all machines to be in Running state again"
+    oc -n "$MACHINE_API_NAMESPACE" wait --all --for=jsonpath='{.status.phase}'=Running --timeout=5m machines
```
bng0y, line 65: is there a risk that this deletes all infra-nodes with one run?

deploy/ocpbugs-15043/script.sh
```diff
@@ -0,0 +1,136 @@
...
+    exit 0
+fi
+
+# print affected nodes and 
+
+for NODE in "${NODES[@]}"; do
+    nodeJson="$(oc get node "$NODE" -ojson)"
+
+    # make sure we only restart infra nodes
+    if ! jq -e '.metadata.labels."node-role.kubernetes.io/infra"' <<< "$nodeJson" > /dev/null ; then
```
bng0y, line 49: what is the rational for only sticking to infra nodes?

deploy/ocpbugs-15043/10-ocpbugs-15043.CronJob.yaml
```diff
@@ -0,0 +1,60 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: ocpbugs-15043
+  namespace: openshift-config
+spec:
+  failedJobsHistoryLimit: 2
+  successfulJobsHistoryLimit: 2
+  concurrencyPolicy: Forbid
+  schedule: "*/5 * * * *"
```
bng0y, line 11: what's motivating the 5-minute schedule?

deploy/ocpbugs-15043/script.sh
mrWinston: Mostly this is here to be 100% sure we don't delete a control plane node accidentally and that we don't interfere with CU workloads. The pods that are monitored by this script should only be running on infra nodes anyway, so in practice, this might not even have any effect.

deploy/ocpbugs-15043/10-ocpbugs-15043.CronJob.yaml
mrWinston: Just as a starting point for a discussion :) I wasn't sure about the interval so i just picket 5 minutes to begin with. Ideally, this cronjob is faster at bouncing faulty machines than our alerting is at detecting the failed pods.

deploy/ocpbugs-15043/script.sh
mrWinston: What do you mean exactly? This command will wait until the targeted machine has been completely deleted.

PR #1948, opened 2023-11-28
Use internal OCM endpoint proxy for MUO on non-management clusters

hectorakemp:
### What type of PR is this?
_(feature)_

### What this PR does / why we need it?
Updates the Managed Upgrade Operator configuration to use the internal OCM endpoint proxy instead of the raw `api.openshift.com` endpoint. Not relevant (yet) to Hypershift clusters. 

Depends on https://issues.redhat.com/browse/OSD-19745, https://issues.redhat.com/browse/OSD-19232 being in stage, as well as https://github.com/openshift/ocm-agent/pull/71 to be merged. 
### Which Jira/Github issue(s) this PR fixes?

_Fixes #_
https://issues.redhat.com/browse/OSD-19764

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #1993, opened 2024-01-18
Revert "Use internal OCM endpoint proxy for MUO on non-management clusters"

hectorakemp:
Reverts openshift/managed-cluster-config#1948

This pull request caused an incident, due to it assuming that production changes in Managed upgrade operator had propagated, resulting in a flood of `CannotRetrieveUpdatesSRE`. See incident channel here: https://redhat.enterprise.slack.com/archives/C06EEJJPUTU


NautiluX (commented):
/lgtm

PR #1996, opened 2024-01-19
[OSD-20478] Import csv_succeeded into observatorium

zmird-r:
Importing csv_succeeded to monitor SRE operator versions and status.

Ticket: [OSD-20478](https://issues.redhat.com/browse/OSD-20478)


mjlshen (approved):
/lgtm

PR #1997, opened 2024-01-23
OSD-19863: Lower retention for management cluster CMO prometheus data

typeid:
### What type of PR is this?
bug

This PR lowers the retention of prometheus data for management clusters to avoid running into full PVs. 

https://issues.redhat.com/browse/OSD-19863

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/cluster-monitoring-config-non-uwm/config.yaml
mjlshen, line 1: I think https://github.com/openshift/managed-cluster-config/blob/master/deploy/cluster-monitoring-config/config.yaml also needs to be excluded from landing on MC's

deploy/cluster-monitoring-config-non-uwm/config.yaml
```diff
@@ -1,10 +1,12 @@
 deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
-  matchLabelsApplyMode: "OR"
```
typeid, line 3: I moved the 4.5 specific out of here so we don't have the `matchLabelsApplyMode: "OR"`

NautiluX (commented):
/lgtm

PR #1999, opened 2024-01-24
AUTH-478: osd-suricata - add proper pod-security labels

ibihim:
### What type of PR is this?

cleanup

The [pod security](https://cloud.redhat.com/blog/pod-security-admission-in-openshift-4.11) level is not set for this namespace. It entered the top 10 of violating openshift namespaces.

You can look it up on thanos when searching for `alerts{alertname="PodSecurityViolation"}`.

In a future update, we will enforce the pod security level and this namespace won't work anymore, without the labels set as the global config will enforce `restricted`.

_Fixes #_ https://issues.redhat.com/browse/AUTH-478

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-ids-tester/100-ids-tester.Namespace.yaml
+++ b/deploy/osd-ids-tester/100-ids-tester.Namespace.yaml
@@ -2,3 +2,10 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-suricata
+  labels:
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/audit: privileged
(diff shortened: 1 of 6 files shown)


clcollins:
/lgtm

Looks like just some minor cleanup/ordering needs to be done for the PR build test to pass.

ibihim:
@clcollins,

I ran `make` according to @hectorakemp (ty!), but it didn't cover all possible places.

I grep'ed and added labels everywhere, WDYT: https://github.com/openshift/managed-cluster-config/pull/1999/commits/14db8a275b8bfe2ad1688926876336a581315149?

Are those not important and I should just drop that commit or is the `make` target missing files?

PR #2000, opened 2024-01-25
change cmo to remoteWrite directly, using oauth2

RaphaelBut:
### What type of PR is this?
cleanup

We are removing the token-refresher by using prometheus remoteWrite directly.
The secret, previously used by token-refresher, is read and used by prometheus now to do oauth2.
Removal of the tokenrefresher resources will be done in a follow up PR, once this has been validated in stage.

https://issues.redhat.com/browse/OSD-18526 

_Fixes #_

- [X ] Tested latest changes against a cluster

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/cluster-monitoring-config/config.yaml
+++ b/resources/cluster-monitoring-config/config.yaml
@@ -5,7 +5,16 @@ prometheusK8s:
 # make sure to adjust https://github.com/openshift/ops-sop/blob/master/v4/howto/observability/o11m_mst_killswitch.md
 # so that killswitch instructions point to the right file!
   remoteWrite:
-    - url: http://token-refresher.openshift-monitoring.svc.cluster.local
+    - url: ${OBSERVATORIUM_URL}
+      oauth2:
+        clientId:
+          secret:
+            key: client-id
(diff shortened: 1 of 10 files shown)


resources/cluster-monitoring-config/config.yaml
mjlshen, line 1: You'll also need to add `OBSERVATORIUM_URL` as a parameter to scripts/template.yml

RaphaelBut: I added it via the app interface deployment to be env specific here https://gitlab.cee.redhat.com/service/app-interface/-/commit/421fdf8af0ae7db768fb6340dedbc9a0305e599e 

This was previously set via the secret as well, but the cmo lacks the ability to read it .

RaphaelBut: Ah it needs to be added here as well https://github.com/openshift/managed-cluster-config/blob/master/scripts/templates/template.yaml#L4-L30

mjlshen (approved):
/lgtm

PR #2007, opened 2024-01-31
Remove token-refresher resources

RaphaelBut:
### What type of PR is this?
cleanup, removing the deployment of token-refresher

remove all token-refresher resources after #2000 is merged 
We now directly remoteWrite metrics from the prometheus pod of cmo instead of going through token-refresher 

_Fixes #_
https://issues.redhat.com/browse/OSD-18526

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid (approved):
/lgtm
/approve

Looks good for staging.

PR #2010, opened 2024-02-02
[OSD-20478] Filter only for SRE Operators in the sre operators record…

zmird-r:
Filtering out metrics for operators not owned by SREP.

Ticket: [OSD-20478](https://issues.redhat.com/browse/OSD-20478)


apahim:
Title says "SRE Operators" and description says "operators not owned by SREP". Which one is it? If it's the former, we are missing `addon-operator` and `package-operator`. If it's the latter, we are good.

mjlshen (approved):
/lgtm

PR #2018, opened 2024-02-08
Remove log linking workaround in 4.15

cblecker:
### What type of PR is this?
feature

This removes the specific log linking workaround implemented here from 4.15+, as this feature will GA and not require this configuration

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #2020, opened 2024-02-15
OSD-20677 schedule monitoring-plugin pods onto infra nodes

yithian:
### What type of PR is this?
feature

schedule monitoring-plugin pods onto infra nodes on 4.14 clusters

_Fixes #_ [OSD-20677](https://issues.redhat.com//browse/OSD-20677)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-monitoring-plugin/config.yaml
+++ b/deploy/osd-monitoring-plugin/config.yaml
@@ -1,0 +1,7 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+    - key: hive.openshift.io/version-major-minor
+      operator: In
+      values: ["4.14", "4.15"]
--- a/deploy/osd-monitoring-plugin/00-cluster-monitoring-config.ConfigMap.yaml
+++ b/deploy/osd-monitoring-plugin/00-cluster-monitoring-config.ConfigMap.yaml
@@ -1,0 +1,6 @@
+apiVersion: ""
+kind: ConfigMap
+name: cluster-monitoring-config
+namespace: openshift-monitoring
+patch: |-
+  { "data": { "config.yaml": "monitoringPlugin:\n  nodeSelector:\n    node-role.kubernetes.io/infra: \"\"\n  tolerations:\n    - effect: NoSchedule\n      key: node-role.kubernetes.io/infra\n      operator: Exists" } }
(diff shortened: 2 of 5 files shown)


yithian:
Please pay special attention to the configmap name and namespace

Since this is for UWM, [the docs](https://docs.openshift.com/container-platform/4.14/monitoring/config-map-reference-for-the-cluster-monitoring-operator.html#userworkloadconfiguration) suggest a different configmap than is specified in OSD-20677

mjlshen (commented):
/hold

Happy to be overriden, but I don't think we want to schedule UWM pods on infra nodes and want them on worker nodes.

fahlmant:
+1 Michael. UWM is purposefully put on worker nodes as customers control it. They can provision persistent storage and configure UWM to use it so they don't lose data. They can also configure the scheduling onto their own specific nodes https://docs.openshift.com/rosa/monitoring/configuring-the-monitoring-stack.html#using-node-selectors-to-move-monitoring-components_configuring-the-monitoring-stack

yithian:
Sorry, I had my mental wires crossed and thought OSD-20677 was about UWM. I've updated the PR to patch the regular cluster-monitoring-config

fahlmant:
@yithian Would you mind updating the title to reflect the change in the PR?

yithian:
done; I updated the description too since it was backwards as well

iamkirkbater:
Wait, the ticket says that these need to be scheduled to infra nodes because these are new pods in 4.14, so my guess is that they're currently deploying to worker nodes by default, but the PR explicitly deploys them to worker nodes.

If I'm understanding this correctly, the original confusion was that this was to put UWM on infra nodes, then we corrected and patched the regular monitoring config, and then corrected again to put the monitoring config onto the worker nodes. I think that last correction was an overcorrection, and that these still need to go on the infra nodes.

Did I miss something?

deploy/osd-monitoring-plugin/config.yaml
```diff
@@ -0,0 +1,7 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
+    - key: hive.openshift.io/version-major-minor
+      operator: In
+      values: ["4.14"]
```
iamkirkbater, line 7: Since 4.15 is only a week or so away should we also add that to this list?

iamkirkbater:
/lgtm
/approve
/hold cancel

Thanks!!

PR #2024, opened 2024-02-19
[OSD-20486] Adding an automated SL when KubeNodeUnschedulableSRE alerting and changing the alert severity to Warning

aliceh:
Adding an automated SL when KubeNodeUnschedulableSRE alerting and changing the alert severity to Warning

https://issues.redhat.com/browse/OSD-20486

### What type of PR is this?

feature 
https://issues.redhat.com/browse/SDE-3667

Documentation is updated in https://issues.redhat.com/browse/OSDOCS-9404

SOP is updated in https://github.com/openshift/ops-sop/pull/3042

### DIFF
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -87,3 +87,9 @@ spec:
       severity: Error
       summary: "Filesystem has less than 3% inodes left"
       logType: Cluster Configuration
+    - activeBody: |-
+        Your cluster is currently alerting with 'KubeNodeUnschedulableSRE' because a worker node '${node}' can not drain. This is not an actionable alert for the SRE team. Please troubleshoot the node using this reference documentation: https://docs.openshift.com/rosa/rosa_cluster_admin/rosa-cordoning-nodes.
+      name: KubeNodeUnschedulableSRE
+      resendWait: 1
+      severity: Error
+      summary: "Action required: a worker node is Unschedulable for more than an hour"
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -88,8 +88,8 @@ spec:
       summary: "Filesystem has less than 3% inodes left"
       logType: Cluster Configuration
     - activeBody: |-
-        Your cluster is currently alerting with 'KubeNodeUnschedulableSRE' because a worker node '${node}' can not drain. This is not an actionable alert for the SRE team. Please troubleshoot the node using this reference documentation: https://docs.openshift.com/rosa/rosa_cluster_admin/rosa-cordoning-nodes.
+        Your cluster currently has at least one unschedulable worker node. Please troubleshoot this node using the following reference documentation: https://docs.openshift.com/rosa/rosa_cluster_admin/rosa-cordoning-nodes.
       name: KubeNodeUnschedulableSRE
-      resendWait: 1
+      resendWait: 6
       severity: Error
       summary: "Action required: a worker node is Unschedulable for more than an hour"
--- a/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
@@ -16,6 +16,8 @@ spec:
       labels:
         severity: Warning
         namespace: openshift-monitoring
+        send_managed_notification: "true"
+        managed_notification_template: "KubeNodeUnschedulableSRE"  
       annotations:
         message: "The node {{ $labels.node }} has been unschedulable for more than an hour."
     - alert: ControlPlaneNodeUnschedulableSRE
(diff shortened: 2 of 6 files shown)


deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
```diff
@@ -143,3 +144,11 @@ spec:
...
         managed_notification_template: WorkerNodeFilesystemAlmostOutOfFiles
         send_managed_notification: "true"
+    - alert: KubeNodeUnschedulableSRE
+      # KubeNodeUnschedulableSRE alert firing).
+      expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}) or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade$"}) > 0)
+      labels:
+        severity: Info
+        namespace: openshift-monitoring
+        send_managed_notification: "true"
+        managed_notification_template: "KubeNodeUnschedulableSRE"
```
iamkirkbater, line 154: Is there a reason that we're redefining this here? It's already defined at https://github.com/openshift/managed-cluster-config/pull/2024/files#diff-4ec82a4ca9332ac099cc824aaf84ad8c9b48c6736af18aebcfea99e5644bf658L13, so we should be able to just update the labels there.

deploy/sre-prometheus/100-node-unschedulable.PrometheusRule.yaml
```diff
@@ -14,7 +14,7 @@ spec:
       expr: (kube_node_spec_unschedulable > 0 and on(node) kube_node_role{role="master"}) or (kube_node_spec_unschedulable > 0 unless ignoring(node,container,endpoint,job,namespace,service) sum(mapi_machine_set_status_replicas{name=~".*upgrade$"}) > 0)
       for: 61m
       labels:
-        severity: critical
+        severity: Warning
         namespace: openshift-monitoring
```
iamkirkbater, line 18:
```suggestion
        namespace: openshift-monitoring
        send_managed_notification: "true"
        managed_notification_template: "KubeNodeUnschedulableSRE"  
```

Per the other comment, we should be able to just add these labels here.

I wonder if we want to reduce the time until the alert fires here as well. 60 minutes was to protect SREs from flapping alerts - but a customer may want to know earlier since it directly affects their cloud spend. Maybe 30 minutes here? I don't feel strongly enough to make that a required change if there's any argument against it though.

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
```diff
@@ -87,3 +87,9 @@ spec:
       severity: Error
       summary: "Filesystem has less than 3% inodes left"
       logType: Cluster Configuration
+    - activeBody: |-
+        Your cluster is currently alerting with 'KubeNodeUnschedulableSRE' because a worker node '${node}' can not drain. This is not an actionable alert for the SRE team. Please troubleshoot the node using this reference documentation: https://docs.openshift.com/rosa/rosa_cluster_admin/rosa-cordoning-nodes.
```
iamkirkbater, line 91: I don't believe that OCM Agent has the capability of templating variable names the same way our PD notifications do. In this case we'll want to either tell them what metric to look at, or give them a command to run.  See the `Workload preventing machine deletion` notification above.

Additionally - this alert can be triggered just from someone cordoning the node. In this case, I think it might be better to be more vague here as to what's going on.

Maybe something along the lines of "Your cluster currently has at least one unschedulable worker node. Please troubleshoot this node using the following reference documentation: [link]"?

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
```diff
@@ -87,3 +87,9 @@ spec:
...
+      name: KubeNodeUnschedulableSRE
+      resendWait: 1
```
iamkirkbater, line 93: Is 1 hour too soon to re-notify? I think as long as we send the first notification quickly that we could define this re-notification to be something like 6 hours, especially if they cordon'd the node on purpose for like a weekend - might not be great to send like 50 service logs about the same issue when they did it on purpose.  Open to other suggestions here, though :D

iamkirkbater (changes requested):
A few small changes requested here.  Great work so far!

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
iamkirkbater: (Disregard the template part, I've since been informed that templating is now a thing in OCM Agent). I'd still like to consider a more generic body text, though :D

deploy/sre-prometheus/ocm-agent/100-ocm-agent.PrometheusRule.yaml
aliceh: Good point. I guess I did not fully understand how the alert directories are structured.

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
aliceh: Good point!

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
aliceh: I agree.

PR #2026, opened 2024-02-22
SDCICD-1229: add cluster metadata cm

jbpratt:
### What type of PR is this?

/feature
/hold

Hive has implemented templating which allows us to propogate cluster
metadata into the cluster to be consumed by PKO and others. Additional
data can be added if there is a usecase.

Signed-off-by: Brady Pratt <bpratt@redhat.com>

Fixes [SDCICD-1229](https://issues.redhat.com//browse/SDCICD-1229)

The goal is to allow for PKO to perform conditonal templating to the resources
based on the data in the configmap

We are leveraging a custom function within hive to pull labels from the
`ClusterDeployment` object

https://github.com/openshift/hive/blob/8c54fc9cac459e0c5852a55775ac570695e3b465/pkg/controller/clustersync/templates.go#L33-L37

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-cluster-metadata/10-osd-cluster-metadata.openshift-config.ConfigMap.yaml
+++ b/deploy/osd-cluster-metadata/10-osd-cluster-metadata.openshift-config.ConfigMap.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: osd-cluster-metadata
+  namespace: openshift-config
+data:
+  api.openshift.com_ccs: "{{ fromCDLabel \"api.openshift.com/ccs\" }}"
+  api.openshift.com_channel-group: "{{ fromCDLabel \"api.openshift.com/channel-group\" }}"
+  api.openshift.com_environment: "{{ fromCDLabel \"api.openshift.com/environment\" }}"
+  hive.openshift.io_cluster-platform: "{{ fromCDLabel \"hive.openshift.io/cluster-platform\" }}"
+  hive.openshift.io_cluster-region: "{{ fromCDLabel \"hive.openshift.io/cluster-region\" }}"
--- a/deploy/osd-cluster-metadata/config.yaml
+++ b/deploy/osd-cluster-metadata/config.yaml
@@ -1,0 +1,4 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  enableResourceParameters: true
+  resourceApplyMode: "Sync"
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -26616,6 +26616,36 @@ objects:
       managed.openshift.io/gitHash: ${IMAGE_TAG}
       managed.openshift.io/gitRepoName: ${REPO_NAME}
       managed.openshift.io/osd: 'true'
+    name: osd-cluster-metadata
+  spec:
+    clusterDeploymentSelector:
+      matchLabels:
+        api.openshift.com/managed: 'true'
+    resourceApplyMode: Sync
+    enableResourceParameters: true
+    resources:
+    - apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: osd-cluster-metadata
+        namespace: openshift-config
+      data:
+        api.openshift.com_ccs: '{{ fromCDLabel "api.openshift.com/ccs" }}'
+        api.openshift.com_channel-group: '{{ fromCDLabel "api.openshift.com/channel-group"
+          }}'
+        api.openshift.com_environment: '{{ fromCDLabel "api.openshift.com/environment"
+          }}'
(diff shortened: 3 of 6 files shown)


deploy/osd-cluster-metadata/10-osd-cluster-metadata.openshift-config.ConfigMap.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: osd-cluster-metadata
+  namespace: openshift-config
+data:
+  api.openshift.com/ccs: {{ fromCDLabel "api.openshift.com/ccs" }}
+  api.openshift.com/channel-group: {{ fromCDLabel "api.openshift.com/channel-group" }}
+  api.openshift.com/environment: {{ fromCDLabel "api.openshift.com/environment" }}
+  hive.openshift.io/cluster-platform: {{ fromCDLabel "hive.openshift.io/cluster-platform" }}
+  hive.openshift.io/cluster-region: {{ fromCDLabel "hive.openshift.io/cluster-region" }}
+  hive.openshift.io/version: {{ fromCDLabel "hive.openshift.io/version" }}
```
mmazur, line 12: Is this where we should be taking versions from? I would assume PKO takes it from the system and we should depend on that.

thetechnick: PKO exposes the Kubernetes API Server version, as reported by the /version API endpoint and the OpenShift Version from the ClusterVersion CR to the templating engine.

Would be interesting to see how the different version properties are changing over time during an upgrade.
I also can't imagine a use-case where the version as seen from hive might be important.

jbpratt: quoted the values, trying to fix the pipeline and removed the version keys. I forgot that PKO exposes this info already

2uasimojo: I agree it would be better to take the version information from the spoke itself. Same goes for anything that's already available there. There will be a delay between when the upgrade completes (per CVO) and when hive updates the labels on the hub CD. See [thread](https://redhat-internal.slack.com/archives/CB53T9ZHQ/p1704199303787309) for discussion around this.

2uasimojo:
PR seems sane to me, syntax-wise.

deploy/osd-cluster-metadata/10-osd-cluster-metadata.openshift-config.ConfigMap.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: osd-cluster-metadata
+  namespace: openshift-config
+data:
+  api.openshift.com/ccs: "{{ fromCDLabel \"api.openshift.com/ccs\" }}"
+  api.openshift.com/channel-group: "{{ fromCDLabel \"api.openshift.com/channel-group\" }}"
+  api.openshift.com/environment: "{{ fromCDLabel \"api.openshift.com/environment\" }}"
+  hive.openshift.io/cluster-platform: "{{ fromCDLabel \"hive.openshift.io/cluster-platform\" }}"
+  hive.openshift.io/cluster-region: "{{ fromCDLabel \"hive.openshift.io/cluster-region\" }}"
```
jewzaam, line 11: Slash is not permitted in CM data keys.

```
$ oc explain configmap.data
KIND:     ConfigMap
VERSION:  v1

FIELD:    data <map[string]string>

DESCRIPTION:
     Data contains the configuration data. Each key must consist of alphanumeric
     characters, '-', '_' or '.'. Values with non-UTF-8 byte sequences must use
     the BinaryData field. The keys stored in Data must not overlap with the
     keys in the BinaryData field, this is enforced during validation process.
```

```
$ cat nmalik-cm.configmap.yaml 
apiVersion: v1
data:
  testing.com/nmalik: "good luck"
kind: ConfigMap
metadata:
  name: nmalik-cm
  namespace: openshift-config
```

And it fails.
```
$ oc create -f nmalik-cm.configmap.yaml 
The ConfigMap "nmalik-cm" is invalid: data[testing.com/nmalik]: Invalid value: "testing.com/nmalik": a valid config key must consist of alphanumeric characters
, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

jewzaam (changes requested):
/hold

jharrington22 (approved):
/lgtm

2uasimojo:
Gah! Should be `enableResourceTemplates`, not `enableResourceParameters`! That's my bad on the review -- I had changed it at some point in the dev process and I guess I forgot which one landed (and didn't double check).

I put up #2030 to fix it. Sorry about that!

PR #2029, opened 2024-02-28
[OSD-21260] hold fire for the node condition alerts to avoid flipping notification sending

bmeng:
### What type of PR is this?
_(bug)_

We are seeing multiple customers reported that the node condition related notification are sending and resolved before they take any action.
Update the alert to hold a bit longer before firing.

_Fixes #[OSD-21260](https://issues.redhat.com//browse/OSD-21260)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid (approved):
/lgtm
/approve

PR #2041, opened 2024-03-13
Revert "ensure CPMS is disabled on 4.15 for now"

iamkirkbater:
Since this didn't make it into GA and we're in the process of enabling CPMS across the fleet, let's not work against ourselves and allow it to be present entirely on fresh-installs of 4.15+ clusters.


typeid (approved):
/lgtm
/approve

PR #2043, opened 2024-03-13
Add route-monitor-operator to sre:operators:succeeded recording rule

zmird-r:
Adds route-monitor-operator to the sre operators succeeded metric.


NautiluX (commented):
/lgtm

PR #2049, opened 2024-03-20
[OSD-21571] Create secret to configure OBO Alertmanager

xiaoyu74:
### What type of PR is this?
_(feature)_

Create secret to configure OBO Alertmanager across fleet of HCP Management Clusters to route alerts payload to ocm-agent-fleet

https://issues.redhat.com/browse/OSD-21571


deploy/hypershift-obo-alertmanager-config/config.yaml
```diff
@@ -0,0 +1,12 @@
+deploymentMode: "SelectorSyncSet"
+SelectorSyncSet:
+  resourceApplyMode: Sync
+  matchExpressions:
```
bmeng, line 4: I think we need to deploy this on management cluster only, you can refer to the other one for ocm-agent  SSS.

deploy/hypershift-obo-alertmanager-config/hypershift-obo-alertmanager-config.secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+data:
+  alertmanager.yaml: Z2xvYmFsOgogIHJlc29sdmVfdGltZW91dDogNW0Kcm91dGU6CiAgcmVjZWl2ZXI6ICJudWxsIgogIHJvdXRlczoKICAtIHJlY2VpdmVyOiBvY21hZ2VudAogICAgbWF0Y2hlcnM6CiAgICAtIHNlbmRfbWFuYWdlZF9ub3RpZmljYXRpb24gPSB0cnVlCiAgICByZXBlYXRfaW50ZXJ2YWw6IDEwbQogIC0gcmVjZWl2ZXI6ICJudWxsIgogIGdyb3VwX3dhaXQ6IDMwcwogIGdyb3VwX2ludGVydmFsOiA1bQogIHJlcGVhdF9pbnRlcnZhbDogMTJoCnJlY2VpdmVyczoKLSBuYW1lOiBvY21hZ2VudAogIHdlYmhvb2tfY29uZmlnczoKICAtIHNlbmRfcmVzb2x2ZWQ6IHRydWUKICAgIHVybDogaHR0cDovL29jbS1hZ2VudC1mbGVldC5vcGVuc2hpZnQtb2NtLWFnZW50LW9wZXJhdG9yLnN2Yy5jbHVzdGVyLmxvY2FsOjgwODEvYWxlcnRtYW5hZ2VyLXJlY2VpdmVyCi0gbmFtZTogIm51bGwiCg==
```
bmeng, line 3: url: http://ocm-agent-fleet.openshift-ocm-agent-operator.svc.cluster.local:8081/alertmanager-receiver

not sure if we can use a variable here

deploy/hypershift-obo-alertmanager-config/config.yaml
xiaoyu74: @bmeng - You mean only keep the key section below? 
```
key: api.openshift.com/managed
```
I saw both `ocm-agent-operator-managedfleetnotifications` and `ocm-agent-operator-managednotifications` have SSS with the `fedramp` only which indicates U.S. government-wide clusters?
```
```

bmeng: to deploy on management cluster only, we need https://github.com/openshift/ocm-agent/blob/master/hack/00-ocm-agent.selectorsyncset.yaml.tmpl#L49

deploy/hypershift-obo-alertmanager-config/hypershift-obo-alertmanager-config.secret.yaml
xiaoyu74: > url: http://ocm-agent-fleet.openshift-ocm-agent-operator.svc.cluster.local:8081/alertmanager-receiver
> 
> not sure if we can use a variable here

You mean having the variable store in a separated file in the same folder of the secret file and just defined the variable name in the secret.yaml?  Let me check if there is any existing SSS following the pattern.

bmeng: hmm, never mind, I think we hard coded the ocm-agent name here as well. https://github.com/openshift/ocm-agent/blob/master/hack/00-ocm-agent.selectorsyncset.yaml.tmpl#L55
So it should be ok.

PR #2052, opened 2024-03-20
OCM-6862 | feat: add tag condition when listing policies attached to roles

davidleerh:
Follow up to thread:
https://redhat-internal.slack.com/archives/CB53T9ZHQ/p1710949371262269

**Manual Test:**
After assuming into a role with a specific policy that has the role tag conditions defined, as shown in this commit:

Assuming into role / session: "dle-stage-HCP-ROSA-Installer-Role"
image

Then,
Querying / listing policies attached to role: "croche-HCP-ROSA-Installer-Role"
image

In the top of the image, I was able to confirm that retrieving list of policies was allowed only when the role itself carried the tag: "red-hat-managed: true"
image

I then removed the "red-hat-managed: true" from role, "croche-HCP-ROSA-Installer-Role"
image

and re-ran the same commands, to confirm the denials:
image

### What type of PR is this?
feature

Adding a condition to permissions (for AWS Owned - HCP Managed Installer Role Policy), IAM: ListRolePolicies and IAM: ListAttachedRolePolicies to only work on roles that carry the tag "red-hat-managed: true".

[OCM-6862](https://issues.redhat.com/browse/OCM-6862)

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.15/hypershift/sts_hcp_installer_permission_policy.json
+++ b/resources/sts/4.15/hypershift/sts_hcp_installer_permission_policy.json
@@ -22,8 +22,6 @@
                 "elasticloadbalancing:DescribeLoadBalancers",
                 "iam:GetOpenIDConnectProvider",
                 "iam:GetRole",
-                "iam:ListAttachedRolePolicies",
-                "iam:ListRolePolicies",
                 "route53:GetHostedZone",
                 "route53:ListHostedZones",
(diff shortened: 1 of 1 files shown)


fahlmant (approved):
/lgtm

PR #2053, opened 2024-03-21
OCMAgentServiceLogsSentExceedingLimit Feature

TomNelsonTembo:
### What type of PR is this?
_Feature_

This PR contains a alerting rule for the ocm_service_log_sent metric.
It alerts SRE if a more than 10 Sevice logs have been sent to a customer in the span of 3 days

[OSD-12237]-(https://issues.redhat.com/browse/OSD-12237) - Alert if the rate of OAO service logs sent is too high

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
@@ -43,7 +43,6 @@ spec:
     # Alert SRE if the OCM agent service log metric has been sent more than 10 within a 3-day period.
     - alert: OCMAgentServiceLogsSentExceedingLimit
       expr: sum(increase(ocm_agent_service_log_sent_total[3d])) > 10
-      for: 1m
       labels:
         severity: warning
         namespace: openshift-monitoring
--- a/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
@@ -40,9 +40,9 @@ spec:
         message: OCM Agent pull secret auth token is not valid.
 
 
-    # Alert SRE if the OCM agent service log metric has been sent more than 20 within a 24-hour period.
+    # Alert SRE if the OCM agent service log metric has been sent more than 10 within a 3-day period.
     - alert: OCMAgentServiceLogsSentExceedingLimit
-      expr: sum(increase(ocm_agent_service_log_sent_total[1d])) > 20
+      expr: sum(increase(ocm_agent_service_log_sent_total[3d])) > 10
       for: 1m
       labels:
         severity: warning
--- a/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
@@ -38,3 +38,15 @@ spec:
         link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/OCMAgentResponseFailureServiceLogsSRE.md#verify-cluster-pull-secrets"
       annotations:
         message: OCM Agent pull secret auth token is not valid.
+
+
+    # Alert SRE if the OCM agent service log metric has been sent more than 20 within a 24-hour period.
+    - alert: OCMAgentServiceLogsSentExceedingLimit
+      expr: sum(increase(ocm_agent_service_log_sent_total[1d])) > 20
+      for: 1m
+      labels:
+        severity: warning
+        namespace: openshift-monitoring
+        link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/OCMAgentServiceLogsSentExceedingLimit.md"
+      annotations:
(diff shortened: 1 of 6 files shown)


deploy/sre-prometheus/ocm-agent/100-ocm-agent-proxy.PrometheusRule.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       labels:
         severity: Info
         namespace: openshift-monitoring
-        managed_notification_template: "AdditionalTrustBundleCAExpiring"
+        managed_notification_template: "AdditionalTrust BundleCAExpiring"
```
typeid, line 19: **issue**: this is an unwanted change, correct?

deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
```diff
@@ -38,3 +38,15 @@ spec:
         link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/OCMAgentResponseFailureServiceLogsSRE.md#verify-cluster-pull-secrets"
       annotations:
         message: OCM Agent pull secret auth token is not valid.
+
+
+    # Alert SRE if the OCM agent service log metric has been sent more than 10 within a 3-day period.
+    - alert: OCMAgentServiceLogsSentExceedingLimit
+      expr: sum(increase(ocm_agent_service_log_sent_total[3d])) > 10
+      for: 1m
```
typeid, line 46: We don't really need the `for`, might as well drop it instead of having it at 1 minute.

deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
```diff
@@ -38,3 +38,15 @@ spec:
         link: "https://github.com/openshift/ops-sop/blob/master/v4/alerts/OCMAgentResponseFailureServiceLogsSRE.md#verify-cluster-pull-secrets"
       annotations:
         message: OCM Agent pull secret auth token is not valid.
+
+
+    # Alert SRE if the OCM agent service log metric has been sent more than 10 within a 3-day period.
+    - alert: OCMAgentServiceLogsSentExceedingLimit
+      expr: sum(increase(ocm_agent_service_log_sent_total[3d])) > 10
```
typeid, line 45: One thing that might cause unwanted pages: we have ocm-agent-operator running on the centralized observability cluster in `fleet-mode` and sending the service logs for all hypershift clusters. 
While 10 service logs is a lot to send to a single cluster (when running in standard mode), it's a pretty low number when looking at the whole fleet.

**Suggestion:** we could scope the alert expression per `_id` which represents the cluster identifier, so that it works the same way for fleet mode and standard mode.

typeid (changes requested):

deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
typeid: We don't seem to be adding `_id` in the `ocm_agent_service_log_sent_total` metric.

@ravitri / @bmeng - probably something we want to figure out with this alert. We need to make sure it works for fleet-mode as well as non-fleet mode. Do you think we should be adding `_id` to the `ocm_agent_service_log_sent_total` metric, then we could do something like:
```
sum(increase(ocm_agent_service_log_sent_total[3d])) by (_id) > 10
```

bmeng: The function `SetTotalServiceLogCount` which is setting the metrics is not being called by the fleetmode webhook for now.
I feel it would be easier if we just define different metrics if we would like to cover fleet mode

bmeng:
You need to run `make` to generate the templates

deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
typeid: That makes sense @bmeng, thanks.

RaphaelBut:
I think the threshold of 10 SLs over 3 days is a reasonable start, however, after playing with the metrics numbers in prod, it seems like we have a couple that would fire right after promoting this https://promlens.stage.devshift.net/?l=A4OtghOlc9k  
I feel like we may want to start with a higher threshold in the beginning and start scraping of the top without too many alerts for primary at once.

In the linked query, I sum over the pod names, which is not very accurate, when two pods on the same cluster send SLs. 

This ties into the '_id' discussion above, which would allow us to be more precise on these queries and find suspicious clusters from observatorium, but it will increase metric cardinality. ;)

deploy/sre-prometheus/100-ocm-agent-operator.PrometheusRule.yaml
Nikokolas3270: Well the `for` keyword is used to avoid alert flapiness.
That said the `ocm_agent_service_log_sent_total` metric is not supposed to increase frequently.
So indeed we could have that removed.

typeid (approved):
/lgtm
/approve

PR #2060, opened 2024-03-26
Add alert for controlplanemachineset has additional machines

RaphaelBut:
### What type of PR is this?
feature

adds an alert for when the controlplanemachineset  is active and the cluster has more than 3 control plane nodes for more than one hour

I put an 'OR' for the node role as we have clusters with 'master' & 'control-plane' and older ones with only 'master'.

https://issues.redhat.com/browse/OSD-20595

- [X] Included documentation changes with PR https://github.com/openshift/ops-sop/pull/3101 

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-controlplanemachineset.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-controlplanemachineset.PrometheusRule.yaml
@@ -1,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: sre-control-plane-machine-set-alerts
+    role: alert-rules
+  name: sre-control-plane-machine-set-alerts
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: sre-control-plane-machine-set-alerts
+    rules:
+        - alert: SREControlPlaneMachineSetStuck
+          expr: cpms_enabled > 0 AND (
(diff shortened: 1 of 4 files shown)


iamkirkbater:
I don't want to block this PR but I do want to note that we _may_ need to revisit this alert in the future depending on how long it takes to replace all three masters.

I potentially see an issue where CPMS is rolling out a new set of machines and replacing all three control plane nodes and each node replacement takes >20m - this would trigger this alert to be fired within the 1h timeframe depending on whether or not the prometheus metric is scraped inbetween the master replacements.

To visualize - we have a control plane size of 3 and the CPMS spec is updated (say to scale) - this will then add a new machine.

``` 3 Machines -> CPMS Spec update -> CPMS replaces machine 1 -> 4 machines -> old machine deleted -> 3 machines -> CPMS replaces machine 2 -> 4 machines -> old machine deleted -> 3 machines -> CPMS replaces machine 3 -> old machine deleted -> 3 machines```

If metrics are not scraped in the few seconds during the `CPMS replaces machine 2` or `CPMS replaces machine 3` steps then CPMS when there are only 3 machines then this alert will fire if the machine replacements just run long and will be a flappy alert.

HOWEVER - I think it would be premature to optimize for this now.

RaphaelBut:
closing as this alert wont be needed currently https://issues.redhat.com/browse/OSD-20595

PR #2051, opened 2024-03-20
[OSD-21066] Remove CAMO olm dance cronjob

zmird-r:

PR #2072, opened 2024-04-05
ITN-56: Fix OSD-21066 cronjob

typeid:
### What type of PR is this?
bug

This PR adds a cronjob for a fleet wide CAMO olm dance, which was not fully successful with https://github.com/openshift/managed-cluster-config/pull/2051. 

The previous PR had a bug that prevented CSVs from being deleted:
```
# This never works, the CSV doesn't have this label
oc -n openshift-monitoring delete clusterserviceversions.operators.coreos.com -l operators.coreos.com/configure-alertmanager-operator.openshift-monitoring=
```

The faulty olm dance is causing a lot of logs to be written in the catalog operator pod and potentially causes weird states on the clusters it ran on (according to telemetry `count(sre:operators:succeeded{name="ocm-agent-operator"})` is 1755, whereas  `count(sre:operators:succeeded{name="configure-alertmanager-operator"})`  is around 200). 

We need this olm dance to put the fleet back on the right version and resolve the excessive logging and potential high networking loads (the latter is unconfirmed to be related). 

The changes with this OLM dance compared to the previous one:
- fixed csv deletion
- keep the subscription as is, it will allow olm to get the latest version

Fixes ITN56

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-21066-camo-olm-dance/01-cronjob.yaml
+++ b/deploy/osd-21066-camo-olm-dance/01-cronjob.yaml
@@ -1,0 +1,43 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: sre-operator-reinstall
+  namespace: openshift-monitoring
+spec:
+  ttlSecondsAfterFinished: 100
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/5 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: sre-operator-reinstall-sa
+          restartPolicy: Never
+          containers:
+          - name: operator-reinstaller
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - sh
+            - -c
+            - |
+              #!/bin/bash
+              set -euxo pipefail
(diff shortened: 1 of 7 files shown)


typeid:
/unhold

Removed subscription deletion, not necessary for the dance to work.

deploy/osd-21066-camo-olm-dance/01-cronjob.yaml
```diff
@@ -0,0 +1,43 @@
...
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/5 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: sre-operator-reinstall-sa
+          restartPolicy: Never
```
RaphaelBut, line 18: why do we go for no restart here? :thinking:

RaphaelBut: ah i guess the cronjob will run again and it does not matter

typeid: I copy pasted most of this from the previous attempt to olm dance, but I believe `restartPolicy`  would restart the pod for a job in case the pod fails, but we would have a new job after 5 minutes anyway, so I think it's no big deal.

NautiluX (commented):
/lgtm

PR #2073, opened 2024-04-05
added /health so console would curl lightweight health endpoint

lnguyen1401:
### What type of PR is this?
Bug
change the URL so RMO would query the lightweight /health endpoint.Not /healthz because that would redirect to the console.
https://issues.redhat.com/browse/OSD-22053
_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #2077, opened 2024-04-10
Add patch for default ingress router replicas to 3, based on the Org ID of the ClusterDeployment

fahlmant:
### What type of PR is this?
feature

Copies OSD-8028, but instead of cluster ID, based on org ID

https://issues.redhat.com/browse/OSD-20989

### DIFF
--- a/deploy/osd-ingress/routerreplicas-osd-20989/config.yaml
+++ b/deploy/osd-ingress/routerreplicas-osd-20989/config.yaml
@@ -1,0 +1,9 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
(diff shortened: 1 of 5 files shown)


deploy/osd-ingress/routerreplicas-osd-20989/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
```
dustman9000, line 3: We should also only do this on multi-az clusters, correct?
```suggestion
  matchExpressions:
      - key: api.openshift.com/multi-az
      operator: In
      values: "true"
```

PR #2079, opened 2024-04-11
OSD-22258: need to use full kind name to avoid confusion

Tof1973:
### What type of PR is this?
_bug_

subscriptions.operators.coreos.com kind should be used to not make confusion when other subscriptions kind are defined

_Fixes OCPBUGS-32102_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid:
@Tof1973 do the other mentions of `subscription` need the full api group as well? e.g. https://github.com/openshift/managed-cluster-config/blob/f48513b2867610e02a2b50979bb9f13658cae17e/deploy/osd-curated-operatorsources-revert/01-osd-patch-subscription-source.ClusterRole.yaml#L9

typeid (approved):
/lgtm
/approve

PR #2087, opened 2024-04-16
[OSD-17651] Send all UWM alerts to the CU

mrWinston:
### What type of PR is this?
_feature_

- Silence UWM Budget Burn alerts and send ManagedNotification to customers instead

_Fixes [OSD-17651](https://issues.redhat.com//browse/OSD-17651)_

### DIFF
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -96,11 +96,12 @@ spec:
       references:
         - "https://docs.openshift.com/container-platform/latest/nodes/nodes/nodes-nodes-working.html#nodes-nodes-working-marking_nodes-nodes-working"
     - activeBody: |-
-        Your cluster requires your attention. User Workload Montioring is currently degraded because unscheduleable worker nodes are preventing deployment of UWM pods. Please scale up the MachinePool for this cluster or uncordon any cordoned worker nodes.
-      name: "UnscheduleableNodesPreventingUWMDeployment"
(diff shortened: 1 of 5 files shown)


deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
```diff
@@ -96,11 +96,12 @@ spec:
       references:
         - "https://docs.openshift.com/container-platform/latest/nodes/nodes/nodes-nodes-working.html#nodes-nodes-working-marking_nodes-nodes-working"
     - activeBody: |-
-        Your cluster requires your attention. User Workload Montioring is currently degraded because unscheduleable worker nodes are preventing deployment of UWM pods. Please scale up the MachinePool for this cluster or uncordon any cordoned worker nodes.
-      name: "UnscheduleableNodesPreventingUWMDeployment"
+        Your cluster requires your attention. User Workload Monitoring is currently degraded. Please review your configuration alongside the product documentation: https://docs.openshift.com/rosa/observability/monitoring/configuring-the-monitoring-stack.html. If you need assistance to restore the User Workload Monitoring, please open a support case.
```
iamkirkbater, line 99: The Service Log template will include the links in the email in the references field below and will also say that if further assistance is needed they can open a support case, so i don't think we need to duplicate that here.

I think "Please review your configuration" is enough here - maybe even "Please review your User Workload Monitoring configuration and ensure your worker nodes have capacity for User Workload Monitoring pods to be scheduled." might work?

iamkirkbater (changes requested):

deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
mrWinston: Makes sense! I'll update the PR.

PR #2092, opened 2024-04-17
OSD-20859 - Use insecure probes for console to avoid failures from redirects

tnierman:
### What type of PR is this?
_bug_

Provides a "stop-gap" solution to the problem of blackbox probes failing when the console pods return a 301. This http code is returned when customers set custom console URLs, as it redirects the probe to use that URL instead of the one configured by the routemonitor object.

Related to https://issues.redhat.com/browse/OSD-20859 (this is a stop-gap solution, it does not fix the root cause)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


tnierman:
Closing in favor of https://github.com/openshift/route-monitor-operator/pull/291, which addresses the root cause of the problem

PR #2101, opened 2024-04-24
[OCPBUGS-15043] fix cronjob failing to detect seccomp issues

mrWinston:
### What type of PR is this?
_bug_

- Currently, the cronjob for detecting the seccomp issue can't read the events for a particular pod, causing it to not detect the condition properly
- This pr adds permissions to list, get and watch events for the cronjob

### DIFF
--- a/deploy/ocpbugs-15043/01-ocpbugs-15043.Role.yaml
+++ b/deploy/ocpbugs-15043/01-ocpbugs-15043.Role.yaml
@@ -68,9 +68,11 @@ rules:
   - 
   resources:
   - pods
+  - events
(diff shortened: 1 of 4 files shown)


iamkirkbater:
I have no issue with this, and give it a +1, but I want to get more than just my own eyes on this.

mjlshen (approved):
/lgtm

It's needed when the script tries to describe pods https://github.com/openshift/managed-cluster-config/blob/6ebf377103474a4035b814d497c9b211f1254739/deploy/ocpbugs-15043/script.sh#L22-L29

PR #2117, opened 2024-05-07
Refactors alert definition.

iamkirkbater:
### What type of PR is this?
Cleanup

Additionally excludes the `openshift-kyverno` namespace as well as makes it easier to add additional exclusions in the future by simplifying the alert.

Additional work for [OSD-6598](https://issues.redhat.com//browse/OSD-6598)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/100-sre-podDisruptionBudget.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-sre-podDisruptionBudget.PrometheusRule.yaml
@@ -13,20 +13,16 @@ spec:
     - alert: PodDisruptionBudgetLimitSRE
       # This takes the original PodDisruptionBudgetLimit alert and adjusts to filter by namespace
       # In the first max by() statement we ONLY care about the `openshift-*` namespaces. We then
-      # do an AND on any namespaces that are NOT `openshift-logging`, `openshift-operators`,
-      # or `openshift-user-workload-monitoring`. When we join the two statements together we are
-      # left with only the openshift-* namespaces that are NOT excluded by the second statement
+      # do an AND on any namespaces that are NOT `openshift-logging`, `openshift-operators`, etc.
+      # When we join the two statements together we are left with only the openshift-* namespaces
+      # that are NOT excluded by the second statement
       #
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/100-sre-podDisruptionBudget.PrometheusRule.yaml
```diff
@@ -13,20 +13,16 @@ spec:
...
       expr: |-
-        max by(namespace, poddisruptionbudget) (
-            kube_poddisruptionbudget_status_current_healthy{namespace=~"openshift-.*"} < kube_poddisruptionbudget_status_desired_healthy{namespace=~"openshift-.*"}
-        )
-        and
-        max by(namespace, poddisruptionbudget) (
-            kube_poddisruptionbudget_status_current_healthy{namespace!~"openshift-(logging|user-workload-monitoring|operators)"} < kube_poddisruptionbudget_status_desired_healthy{namespace!~"openshift-(logging|user-workload-monitoring|operators)"}
-        )
+        max by(namespace, poddisruptionbudget) ( kube_poddisruptionbudget_status_current_healthy{namespace=~"openshift-.*"} < kube_poddisruptionbudget_status_desired_healthy{namespace=~"openshift-.*"} )
+        unless on(namespace)
```
xiaoyu74, line 24: I see that the only thing you've updated for the prometheus rule is to use the `unless` for excluding specified `openshift-*` namespaces, with the newly added `openshift-kyverno`. So, the main logic for excluding namespaces **remains same** with the previous, but just introduced the `unless` expression?

iamkirkbater: Yes, the only effective change here is that we add the kyverno namespace to the list of exclusions. The overall query remains the same, but this is overall much easier to parse. 

As for the old alert not being sent to the null receiver, you're right, it's not in this PR, it was handled in a separate PR in the CAMO repo. 

I should have included those notes in the PR description. Thanks for being thorough!!

PR #2129, opened 2024-05-22
SDCICD-1292: set MGO NS resourceApplyMode=Upsert

jbpratt:
### What type of PR is this?

/cleanup

This namespace should not be defined in MCC as it conflicts with the namespace defined in the operator repository. Set the resourceApplyMode to Upsert to allow deleting the SSS in a future PR, and avoiding deletion of the namespace itself (the other SSS will take care of that).

The namespace is also defined here:

https://github.com/openshift/must-gather-operator/blob/36981769f414e8c138484e4c8f370cba63aafa1c/hack/olm-registry/olm-artifacts-template.yaml#L33

https://redhat-internal.slack.com/archives/CFJD1NZFT/p1716216917180149?thread_ts=1715864690.782999&cid=CFJD1NZFT

https://issues.redhat.com/browse/SDCICD-1292

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-must-gather-operator/config.yaml
+++ b/deploy/osd-must-gather-operator/config.yaml
@@ -1,5 +1,5 @@
 deploymentMode: "SelectorSyncSet" 
 selectorSyncSet: 
-  resourceApplyMode: "Sync"
+  resourceApplyMode: "Upsert"
 policy:
   destination: "acm-policies"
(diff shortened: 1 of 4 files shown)


mjlshen (approved):
/lgtm

PR #2130, opened 2024-05-22
OSD-22575: add permissions to dedicated-reader and dedicated-admin to view console/monitoring tab data

typeid:
### What type of PR is this?
bug

With 4.15, `cluster-monitoring-view` [was updated](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/cluster-role-view.yaml#L15-L24) to work with `kube-rbac-proxy`, but now requires access to `prometheuses/api` in the `openshift-monitoring` namespace. Previously, `oauth-proxy` was used and permissions were gated on `get namespaces`. 

Resources  needed for the console monitoring tab for 4.15+ are now queried through the prometheuses/api subresource in the `openshift-monitoring` namespace

https://issues.redhat.com/browse/OSD-22575

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/acm-policies/50-GENERATED-rbac-permissions-operator-config.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-rbac-permissions-operator-config.Policy.yaml
@@ -898,6 +898,16 @@ spec:
                                 - get
                                 - list
                                 - watch
+                            - apiGroups:
+                                - monitoring.coreos.com
+                              resourceNames:
+                                - k8s
+                              resources:
+                                - prometheuses/api
+                              verbs:
+                                - get
+                                - create
+                                - update
                     - complianceType: mustonlyhave
                       metadataComplianceType: musthave
                       objectDefinition:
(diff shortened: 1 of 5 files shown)


deploy/rbac-permissions-operator-config/05-dedicated-admins-aggregate-cluster.ClusterRole.yaml
```diff
@@ -363,4 +363,16 @@ rules:
...
+- apiGroups:
+  - monitoring.coreos.com
+  resourceNames:
+  - k8s
+  resources:
+  - prometheuses/api
+  verbs:
+  - get
+### END - customer can view observe tab data in console
+
```
jewzaam, line 377: Changes to this role are not needed.  The osd-readers ClusterRole aggregates to dedicated-readers which in turn aggregates to dedicated-admins.

jewzaam (approved):
Noted in a comment, the change to dedicated-admins-aggregate-cluster is not needed because of aggregation.

osd-readers --> dedicated-readers --> dedicated-admins

But it doesn't cause any negative side effects and this specific change set is tested so won't block on it.

/approve
/lgtm

jewzaam (commented):
/lgtm

PR #2134, opened 2024-06-04
removes fedramp clusters from commercial specific sss

tonytheleg:
### What type of PR is this?

bug

FedRAMP does not have the cloud ingress operator, nor does it support Adobe in FedRAMP as a customer. Since we do not have FedRAMP specific values to deploy with for these SSS, removing them from our clusters as it causes sync errors and noise.

_Fixes #_ FedRAMP in-boundary ticket

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/approve
/lgtm

PR #2136, opened 2024-06-05
[OSD-22081] Add new patch to allow dns-default pods to schedule on infra nodes.

bergmannf:
This is to fix OSD-22081 - probes from blackbox-exporter might fail if an infra nodes uses an overloaded workers node's DNS pod.

Simply moving the blackbox-exporter lead to problems for 4.12 clusters.

### What type of PR is this?
Feature

It will patch the DNS operator to allow running the pods on infra nodes to prevent spurios probe failures for the blackbox exporter, when the infra node uses a worker node's DNS pod but the worker is too overloaded to respond.

OSD-22081

- [X] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-dns-default/README.md
+++ b/deploy/osd-dns-default/README.md
@@ -1,0 +1,13 @@
+# OSD-22081
+
+This will allow dns-default pods to run on every node of a cluster.
+
+This did lead to problems, when infra nodes were using worker nodes that were
+overloaded to handle the DNS requests.
+
+Instead with this patch, the infra nodes will get their own dns-default pods,
+removing this source of issues.
+
+This patch has to patch-in the default master-toleration as well, as just adding
+the infra toleration will remove the implicit master toleration that is present
+in a default installation.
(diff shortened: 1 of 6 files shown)


bergmannf:
@bmeng This would be an alternative way to solve OSD-22081, given that we had problems when rescheduling the BBE pods to masters - would be interested what you think about doing it this way.

bmeng:
Changing this in an alternative way sounds good to me. The only concern to me is I am not sure if this would have any side effect, like the race condition (I assume it should not have).
Another thing to note, the toleration for master exists for the dns-default daemonset even without configured in the dns.operator.openshift.io CR.

bergmannf:
It seems as soon as you put an explicit toleration into the CR it removes the automatic-master toleration unfortunately.
At least when testing this on a staging cluster I had to explicitly add it as well, or our master nodes would loose their dns-default pod.

PR #2138, opened 2024-06-05
MUO: ignore kubelet debug

typeid:
### What type of PR is this?
bug

Debug handlers are expected to be disabled on int/staging. The alert blocks MUO from ugprading.

Reverts https://github.com/openshift/managed-cluster-config/pull/2137 and fixes the issue via MUO

Fixes # https://github.com/openshift/managed-cluster-config/pull/2137

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bmeng:
Wondering what is the purpose for this quick change and revert, just the workaround for some of the clusters being blocked during the upgrade?
If it is something we can ignore during upgrade, maybe just put it there permanent?

PR #2142, opened 2024-06-10
Revert osd 22081

bergmannf:
### What type of PR is this?

Reverting the patch - right now this does put dns-default onto infra nodes, but they get rescheduled periodically leading to unneeded churn on infra nodes.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


iamkirkbater:
/lgtm

for when the CI jobs finish

PR #2143, opened 2024-06-10
Remove 4.16 STS acks as they aren't required by OCP, and there are no STS changes

fahlmant:
Removing STS 4.16 ack added in https://github.com/openshift/managed-cluster-config/pull/2135
OCP 4.15 -> 4.16 does not require this as there are no STS changes between the two versions.

https://issues.redhat.com/browse/OSD-21742


xiaoyu74:
@fahlmant - Let's hold the PR for now based on the discussion: https://redhat-internal.slack.com/archives/C070BJ1NS1E/p1718024423392559

PR #2144, opened 2024-06-11
[OSD-22081] Put dns-default pods on infras

bergmannf:
### What type of PR is this?
Feature: reintroduce the patch for [OSD-22081](https://issues.redhat.com//browse/OSD-22081) - this allows dns-default pods to run on infras and updates the cronjob to rebalance infra nodes to not remove them.

Fixes [OSD-22081](https://issues.redhat.com//browse/OSD-22081)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-dns-default/README.md
+++ b/deploy/osd-dns-default/README.md
@@ -1,0 +1,13 @@
+# OSD-22081
+
+This will allow dns-default pods to run on every node of a cluster.
+
+This did lead to problems, when infra nodes were using worker nodes that were
+overloaded to handle the DNS requests.
+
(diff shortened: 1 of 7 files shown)


typeid (approved):
/lgtm
/approve

Context from slack: there was no business reason to not have dns-default pods on infras, we just never had them run there by default. This caused some other alert to trigger when these pods scheduled onto infras due to a race condition. This alert no longer exists and also having dns-default pods on infras, and there doesn't seem any good reason to not have dns-default pods on infra nodes.

PR #2147, opened 2024-06-13
ROX-24728: additional ACS permissions

ludydoo:
### What type of PR is this?
_feature_

Allow ACS SREs to exec and portforward onto ACS-workloads

Fixes [ROX-24728](https://issues.redhat.com//browse/ROX-24728)

Context: https://redhat-internal.slack.com/archives/C0313JYKH8W/p1718237825072439?thread_ts=1717613729.366549&cid=C0313JYKH8W

Please review the "Considerations" section in the ACS permissions docs: 

#### Considerations
- The `AcsFleetShard` addon installs the resources needed to run an ACS-CS DataPlane cluster:
  - It installs `fleetshard` which is responsible for creating the tenant namespaces.
  - It installs supporting resources, such as monitoring, logging.
- The `AcsFleetShard` addon is *only* installed on those clusters. It will *never* be installed on customer clusters.
- These clusters are *only* accessible by internal staff, such as SRE and ACS-CS team members. Customers will *not* have access to these clusters.
- Each tenant gets its own namespace, `Central` instance, `Route` and DNS record.
- Tenants are only given the URL to their `Route`. They do not interact with the cluster in any other way.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


fahlmant (approved):
/lgtm

PR #2148, opened 2024-06-17
[OSD-22254] initial permissions for gcp wif

mrWinston:
### What type of PR is this?
_feature_

This PR adds the first iteration of WIF templates. These can be seen as the gcp counterpart to the sts policies stored in this repo as well.

This PR must be merged before it's app-interface counterpart: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/105688

_Fixes [OSD-22254](https://issues.redhat.com//browse/OSD-22254)_


resources/wif/README.md
```diff
@@ -0,0 +1,57 @@
...
+service accounts and permissions needed for the cluster's deployment and
+operation. This criteria is defined here, and then passed to OCM where it hosts
+an API endpoint for clients to query. The API provided by Clusters Service
+watches for changes in [this](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/ocm/shared-resources/common.yml)
+file. When this file changes a new [ConfigMap](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/resources/services/ocm/gcp-wif-template.configmap.yaml)
+is templated which Clusters Service pods (OCM) will read.
+
+## How to update WIF templates
+
+When Service Accounts, Roles and/or Permissions need to be updated, execute the following.
```
typeid, line 14: Can permissions be updated? AFAIU these permissions are consumed by the ocm-cli and changing them would also require customers to re-create the roles. I think for STS we don't update the permissions during a y stream.

fahlmant: That's the only time that STS permissions get updated on AWS clusters; as part of a y-stream upgrade.

typeid: Might have expressed it wrong but that's what I meant. The references piece of text sounds like upgrading within a y stream is a possibility.

mrWinston: I reworded the sentence slightly. Once I've updated the GAP analysis document for WIF, i will also link that here.

resources/wif/4.15/vanilla.yaml
```diff
@@ -0,0 +1,354 @@
...
+        predefined: true
+      - id: iam.ServiceAccountKeyAdmin
+        kind: Role
+        predefined: true
+      - id: iam.ServiceAccountUser
+        kind: Role
+        predefined: true
+      - id: storage.admin
+        kind: Role
+        predefined: true
```
typeid, line 32: Shouldn't this one just need the installer permissions?

typeid (approved):
/lgtm
/approve

Let's merge this to unblock OCM and see if we missed anything in the permissions.

PR #2157, opened 2024-07-02
[SDCICD-1292] Remove osd-must-gather-operator SSS containing MGO namespace

mmazur:
### What type of PR is this?
cleanup

Primary reason for this change: having MCC's SSS and MGO's SSS co-own the namespace blocks adoption of MGO into the Release Bundle ([SDE-3672](https://issues.redhat.com/browse/SDE-3672) aka Release Versioning for Managed OpenShift), because it would mean Hive and Package Operator would keep fighting about the namespace, which is something we don't want to risk.

The SSS has already been set to Upsert mode over a month ago, so there's no additional prep work required here. This change should not have any adverse effects.

#### Details
The config being removed creates he `openshift-must-gather-operator` namespace on osd/rosa classic and rosa hcp clusters. However:
* Last I checked, we don't run mgo on rosa hcp clusters, so it's just an empty namespace.
* On osd clusters each operator has its own SSS and the first thing that SSS does is [set up the operator's namespace](https://github.com/openshift/must-gather-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L32-L40).

On the second point – there is no benefit (that I'm aware of) in MCC duplicating the ns setup and indeed MGO is the only operator I found that has this duplication present in MCC (and it was set up slightly wrong, since the ns definition in MCC was a bit different than the definition in MGO's SSS, which we already corrected in https://github.com/openshift/must-gather-operator/pull/136).

As as I see it: this namespace setup duplication is either actually required, in which case it would've been present for all osd operators (and it's definitively not) or it's superfluous, in which case unblocking RVMO by merging this is the correct action.

#### Ref
* [This slack thread](https://redhat-internal.slack.com/archives/CFJD1NZFT/p1716216917180149?thread_ts=1715864690.782999&cid=CFJD1NZFT)

lgtmrvmo

_Fixes [SDCICD-1292](https://issues.redhat.com//browse/SDCICD-1292)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid (approved):
Thanks for the great description, makes perfect sense to me. 

/lgtm
/approve

PR #2159, opened 2024-07-02
[OHSS-32768] Add a job to remove the added dns tolerations.

bergmannf:
This should remove the pods from infra nodes, instead of splitting the fleet into clusters with tolerations and those without.
This would be the case, because [OSD-22081](https://issues.redhat.com//browse/OSD-22081) added tolerations for dns-default pods for infra nodes, but the removal of the patch did not remove those tolerations, so this will remediate this.

### What type of PR is this?
Bug

Remove tolerations that were added to run dns-default pods on infra nodes.

[OHSS-35778](https://issues.redhat.com//browse/OHSS-35778)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-22018-remove-dns-tolerations/04-job.yaml
+++ b/deploy/osd-22018-remove-dns-tolerations/04-job.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "sre-remove-dns-tolerations"
+  namespace: "openshift-sre-pruning"
+spec:
+  ttlSecondsAfterFinish: 100
+  template:
+    spec:
+      serviceAccountName: "sre-remove-dns-tolerations"
+      restartPolicy: Never
+      containers:
+      - name: delete-tolerations
+        image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+        imagePullPolicy: Always
+        command:
+        - sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euxo pipefail
+          oc patch dnses.operator.openshift.io default --type merge -p '{"spec": {"nodePlacement": {"tolerations": []}}}'
--- a/deploy/osd-22018-remove-dns-tolerations/01-serviceaccount.yaml
+++ b/deploy/osd-22018-remove-dns-tolerations/01-serviceaccount.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "sre-remove-dns-tolerations"
+  namespace: "openshift-sre-pruning"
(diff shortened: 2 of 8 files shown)


bergmannf:
/hold

Tested these changes against one staging cluster only, so I will hold this until I can test it against another one.

deploy/osd-22018-remove-dns-tolerations/04-job.yaml
```diff
@@ -0,0 +1,23 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "sre-remove-dns-tolerations"
+  namespace: "openshift-dns"
```
bmeng, line 6: Do we need to run in this namespace?

Maybe out of topic, but I think we do not have a good way to clean up those sa/job/roles, etc. If we have convention to put those temporary one-off resources in a single place, it would be easy for us to recycle?

bergmannf: I also am not the most happy putting it to openshift-dns, but it seemed to be the most appropriate one.

I check with `oc get ns` but didn't really see a decent namespace - it could probably go to `openshift-sre-pruning`?

bmeng: I have no better idea, just as I said above it might be out of topic. maybe something we can figure out later on.

bergmannf: I've changed to `openshift-sre-pruning` for now so it doesn't clutter a namespace that is more common to be checked by other engineers / customers.

bergmannf:
For completeness the tests I ran:
1. Add some tolerations to the DNS:
```
oc patch dnses.operator.openshift.io default --type merge -p '{"spec": {"nodePlacement": {"tolerations": [{"effect": "NoSchedule", "key": "node-role.kubernetes.io/infra", "operator": "Exists"}, {"effect": "NoSchedule", "key": "node-role.kubernetes.io/master", "operator": "Exists"}]}}}' --as backplane-cluster-admin
```
2. Create all the objects in the cluster: 
```
cat 0*.yaml | oc apply -f - --as backplane-cluster-admin
serviceaccount/sre-remove-dns-tolerations created
clusterrole.rbac.authorization.k8s.io/sre-remove-dns-tolerations created
clusterrolebinding.rbac.authorization.k8s.io/sre-remove-dns-tolerations created
job.batch/sre-remove-dns-tolerations created
```
3. Verify the patch command is run:
```
+ oc patch dnses.operator.openshift.io default --type merge -p '{"spec": {"nodePlacement": {"tolerations": []}}}'
dns.operator.openshift.io/default patched
```
4. Reapply the job to make sure patching an already good state is error-free:
```
oc delete job -n openshift-sre-pruning sre-remove-dns-tolerations
cat 04-job.yaml | oc apply -f -
oc logs -n openshift-sre-pruning sre-remove-dns-tolerations-kl4fq
+ oc patch dnses.operator.openshift.io default --type merge -p '{"spec": {"nodePlacement": {"tolerations": []}}}'
dns.operator.openshift.io/default patched (no change)
```

```

PR #2161, opened 2024-07-05
adding machineautoscalers RBAC and requirements

diakovnec:
### What type of PR is this?

Based on the slack thread mentioned in `Special notes` below was decided to create a  PR in order to update RBAC rules in a proper way as CEE should have read access to the non-sensetive objects so that MCS/CEE can get the object from `oc` directly.

By adding read-only RBAC to `machineautoscalers` will allow CEE to troubleshoot issue related to autos-calling easier

`oc get machineautoscaler -o yaml -n openshift-machine-api

apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
Error from server (Forbidden): machineautoscalers.autoscaling.openshift.io is forbidden: User "system:serviceaccount:openshift-backplane-cee:8d5ec3a2bb1e28f30a0c48a9eeb707f3" cannot list resource "machineautoscalers" in API group "autoscaling.openshift.io" in the namespace "openshift-machine-api"`

Discussed in the backplane slack thread 

https://redhat-internal.slack.com/archives/C016S65RNG5/p1720078609433969

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/cee/01-cee-cluster-readers-cluster.ClusterRole.yml
+++ b/deploy/backplane/cee/01-cee-cluster-readers-cluster.ClusterRole.yml
@@ -39,3 +39,13 @@ rules:
   - get
   - list
   - watch
+# CEE can view machineautoscalerconfigs
+- apiGroups:
+  - autoscaling.openshift.io
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
+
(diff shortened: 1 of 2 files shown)


diakovnec:
trying to figure out why actions are failing as its allowed for this fork

PR #2166, opened 2024-07-09
Correct bundle version

fahlmant:
Corrects the bundle version to match:
saas-custom-domains-operator-bundle/-/tree/production/custom-domains-operator/0.1.247-394c11f


jharrington22 (approved):
Validated the version is correct. 

/lgtm

PR #2170, opened 2024-07-11
Add permission to tag ENI's with red-hat-managed: true

fahlmant:
https://issues.redhat.com/browse/OSD-19650


rafael-azevedo (approved):
tested the new permissions set and it is working as expected

PR #2171, opened 2024-07-11
ROX-25275: Add permissions to read ACS addon objects for ACS team

kurlov:
### What type of PR is this?
_feature_

ACS team should be able to list and view ACS Addon related resources

[ROX-25275](https://issues.redhat.com/browse/ROX-25275)

~- [ ] Tested latest changes against a cluster~
~- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:~

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/backplane/acs/01-acs-admins-project.ClusterRole.yaml
+++ b/deploy/backplane/acs/01-acs-admins-project.ClusterRole.yaml
@@ -45,10 +45,9 @@ rules:
   - get
   - list
   - watch
-# ACS Team can view ACS Addon CRs
+# ACS Team can view ACS Addon CRs in package-operator.run API group
 - apiGroups:
   - package-operator.run
-  - addons.managed.openshift.io
   resources:
   - clusterpackages
   - clusterobjectdeployments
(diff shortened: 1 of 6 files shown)


deploy/backplane/acs/01-acs-admins-project.ClusterRole.yaml
```diff
@@ -45,6 +45,26 @@ rules:
   - get
   - list
   - watch
+# ACS Team can view ACS Addon CRs
+- apiGroups:
+  - package-operator.run
+  - addons.managed.openshift.io
```
erdii, line 51: i think the rules should be separated by apigroup so it's clear which resources belong to the `package-operator.run` group and which ones belong to `addons.managed.openshift.io`

erdii (commented):
I don't know which clusters in our managed fleet receive these permissions via ACM so I don't know how relevant the following is:

The cluster selector for the ACM PlacementRule [1] only contains the prod version of the addon.

[1] https://github.com/openshift/managed-cluster-config/blob/a5db7d40837d82bd4bab81be10edb7d274de2dec/deploy/backplane/acs/config.yaml#L11

deploy/backplane/acs/01-acs-admins-project.ClusterRole.yaml
kurlov: Splitted it into two groups

PR #2174, opened 2024-07-22
Restore MGO's NS, but for fedramp only

mmazur:
### What type of PR is this?
bug

There are parts of MCC that assume the NS is present. In commercial this works perfectly fine, because the Release Bundle sets up MGO. In FedRamp MGO does not exist, so the (superflous in FR) configs that depend on the NS cause SSSes to error out.

The more complex solution would be to make sure those unneeded MGO NS-dependant configs do not get applied in FR. The easier solution is to just make sure the namespace exists in FR, so nothing errors out. Hence this change.

This is a followup fix for #2157

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/config.yaml
+++ b/deploy/config.yaml
@@ -1,0 +1,8 @@
+deploymentMode: "SelectorSyncSet" 
+selectorSyncSet: 
+  resourceApplyMode: "Sync"
+  matchExpressions:
+  - key: api.openshift.com/fedramp
+    operator: In
+    values:
+      - "true"
--- a/deploy/01-openshift-must-gather-operator.Namespace.yaml
+++ b/deploy/01-openshift-must-gather-operator.Namespace.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-must-gather-operator
+  annotations:
+    openshift.io/node-selector: ""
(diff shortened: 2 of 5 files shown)


deploy/config.yaml
```diff
@@ -0,0 +1,8 @@
+deploymentMode: "SelectorSyncSet" 
+selectorSyncSet: 
+  resourceApplyMode: "Sync"
```
mjlshen, line 3:
```suggestion
  resourceApplyMode: "Upsert"
```
Should this be `Upsert` instead? That's what was there previously: https://github.com/openshift/managed-cluster-config/pull/2157/files#diff-d80f8bf0f9ac8d5f85a7a45a38c33f50f9925fd44e75043151bd6820d7656d80L3, though I'm not sure if anything else modifies this namespace's labels

mjlshen: Ah ok, it's so we don't accidentally delete the namespace when we delete this syncset https://github.com/openshift/hive/blob/master/docs/syncset.md - imo yes we should change this to `Upsert`

/hold

mjlshen: This was changed to Upsert during the migration and was originally `Sync`

/unhold
/approve

PR #2176, opened 2024-07-25
Remove dedicated-admin access to must-gather-operator resources

sam-nguyen7:
### What type of PR is this?
security enhancement

Removes the ability for dedicated-admin users and service accounts to access and create must-gather-operator resources.

https://issues.redhat.com/browse/OSD-21667

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


sam-nguyen7:
/hold for review. 

This doesn't fully cover the use case because mustgathers can still be created in any other namespace and be reconciled by MGO

PR #2186, opened 2024-08-06
removing openshift-cluster-api-gcp from WIF templates

renan-campos:
This service account is used for the tech preview feature 'cluster-api', and disrupts the provisioning of WIF clusters. It has therefore been removed from the list of service accounts that need to be created for each version of WIF.

### What type of PR is this?
cleanup

https://issues.redhat.com/browse/OSD-24925

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid (approved):
/lgtm
/approve

PR #2188, opened 2024-08-09
OSD-20975: Not deploying MVO on GCP WIF clusters

Nikokolas3270:
### What type of PR is this?
_feature_

### What this PR does / why we need it?
Lets make sure MVO (Managed Velero Operator) is not deployed on OSD clusters running on GCP with WIF (Workload Identity Federation) enabled.
WIF is the GCP equivalent of AWS STS; both are implementing OAuth 2.0. Our policy in that managed clusters should only deploy base operators (i.e. operators that come with OCP) when OAuth 2.0 is in use; indeed we want to keep to a minimal the actions eventually performed by the cluster on the customer cloud provider account when this authentication scheme is in use.
This excludes MVO which is:
- Not a "base" operator, that is to say not an operator bundled with OCP
- Accessing the customer cloud provider

### Which Jira/Github issue(s) this PR fixes?
Implements [OSD-20975](https://issues.redhat.com//browse/OSD-20975)

### Special notes for your reviewer:
The new label has been added in OCM with this MR:
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8410

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #2189, opened 2024-08-09
OSD-24978: Not deploying CIO on GCP WIF clusters + uniformizing WIF handling with STS

Nikokolas3270:
### What type of PR is this?
_feature_

Lets make sure CIO (Cloud Ingress Operator) is not deployed on OSD clusters running on GCP with WIF (Workload Identity Federation) enabled.
WIF is the GCP equivalent of AWS STS; both are implementing OAuth 2.0. Our policy in that managed clusters should only deploy base operators (i.e. operators that come with OCP) when OAuth 2.0 is in use; indeed we want to keep to a minimal the actions eventually performed by the cluster on the customer cloud provider account when this authentication scheme is in use.
This excludes CIO which is:
- Not a "base" operator, that is to say not an operator bundled with OCP
- Accessing the customer cloud provider

Implements [OSD-24978](https://issues.redhat.com//browse/OSD-24978)

The new `api.openshift.com/wif` label has been added in OCM with this MR:
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8410

We are also renaming a few files to uniformize the handling of AWS STS & GCP WIF

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/cloud-ingress-operator-configuration/apischeme/config.yaml
```diff
@@ -7,6 +7,9 @@ selectorSyncSet:
   - key: api.openshift.com/sts
     operator: NotIn
     values: ["true"]
+  - key: api.openshift.com/wif
+    operator: NotIn
+    values: ["true"]
```
typeid, line 12: Does this actually stop the operator from deploying? I would think this only removes configuration. Is this good enough? 

IIRC it's also deployed to AWS STS, it just doesn't do anything. If that's the case, I think it's fine.

deploy/osd-rebalance-infra-nodes/non-fr-sts-wif-privatelink/config.yaml
```diff
@@ -4,6 +4,9 @@ selectorSyncSet:
   - key: api.openshift.com/sts
     operator: NotIn
     values: ["true"]
+  - key: api.openshift.com/wif
+    operator: NotIn
+    values: ["true"]
```
typeid, line 9: Could we also update the README in deploy/osd-rebalance-infra-nodes to match the new behavior?

deploy/osd-serviceaccounts/non-sts-wif-private-link/config.yaml
```diff
@@ -5,6 +5,9 @@ selectorSyncSet:
   - key: api.openshift.com/sts
     operator: NotIn
     values: ["true"]
+  - key: api.openshift.com/wif
```
typeid, line 8: Can we also update the README(s) in osd-serviceaccounts/*?

deploy/osd-serviceaccounts/wif/00-serviceaccounts.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
typeid, line 1: Can we merge this whole file into the sts folder and call the folder `sts-wif`? If I'm not mistaken, it's the same config.

deploy/cloud-ingress-operator-configuration/apischeme/config.yaml
mjlshen: We will need to update [https://github.com/openshift/cloud-ingress-operator/blob/03bb9046ea211f56b782012100b0c72d58a1418f/hack/olm-registry/olm-artifacts-template.yaml#[…]5](https://github.com/openshift/cloud-ingress-operator/blob/03bb9046ea211f56b782012100b0c72d58a1418f/hack/olm-registry/olm-artifacts-template.yaml#L37-L45) to not deploy the CIO operator itself on WIF clusters (and similar for MVO)

Nikokolas3270: Hello Michael is right, I need to update the CIO repo as well
This won't be enough

deploy/osd-serviceaccounts/wif/00-serviceaccounts.yaml
Nikokolas3270: True, it is the same config indeed. Will try to do that if there is a way to do an OR on "sts" and "wif" conditions

Nikokolas3270:
FYI, this PR is related to this one: https://github.com/openshift/managed-cluster-config/pull/2188

deploy/cloud-ingress-operator-configuration/apischeme/config.yaml
Nikokolas3270: FYI, CIO PR:
https://github.com/openshift/cloud-ingress-operator/pull/340

deploy/osd-serviceaccounts/wif/00-serviceaccounts.yaml
Nikokolas3270: OR is not a possibility
https://github.com/openshift/hive/blob/master/config/crds/hive.openshift.io_selectorsyncsets.yaml#L64
`matchExpressions` can only be ANDed
Checking if `00-serviceaccounts.yaml` can be defined at a parent level.

Nikokolas3270: Ok it does not look it is possible to do this parent level sharing:
```
> for f in `find . | grep '\.yaml$'`; do if ! [ -f "$(dirname "$f")/config.yaml" ]; then echo "$f"; fi; done
./resources/managed/all-osd-resources.yaml
./resources/prometheusrules/fluentd_openshift-logging_collector.PrometheusRule.yaml
./resources/prometheusrules/elasticsearch_openshift-logging_elasticsearch-prometheus-rules.PrometheusRule.yaml
./resources/wif/4.15/vanilla.yaml
./resources/wif/4.14/vanilla.yaml
./resources/wif/4.16/vanilla.yaml
./resources/addons-namespaces/main.yaml
./resources/tailored-profiles/hypershift-cis-tp.yaml
./resources/tailored-profiles/hypershift-pcidss-tp.yaml
./scripts/policy-generator-config.yaml
./scripts/templates/template.yaml
./scripts/templates/selectorsyncset.yaml
```
YAML files in `./deploy` are always with a `config.yaml` file in the same folder.

PR #2190, opened 2024-08-09
[OSD-24939] This commit adds back the STS roles needed by AVO in the …

pheckenlWork:
…FedRamp environment.

### What type of PR is this?

This PR adds roles needed for AVO in the FedRamp environment.

We will be unable to monitor 4.13 and 4.14 clusters fully in the FedRamp environment without this change.

[OSD-24939](https://issues.redhat.com//browse/OSD-24939)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/resources/sts/4.13/openshift_aws_vpce_operator_avo_aws_creds_policy.json
+++ b/resources/sts/4.13/openshift_aws_vpce_operator_avo_aws_creds_policy.json
@@ -17,10 +17,21 @@
         "ec2:CreateVpcEndpoint",
         "ec2:DeleteVpcEndpoints",
         "ec2:DescribeVpcEndpoints",
+        "ec2:DescribeVpcs",
         "ec2:ModifyVpcEndpoint",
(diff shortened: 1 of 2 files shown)


resources/sts/4.13/openshift_aws_vpce_operator_avo_aws_creds_policy.json
```diff
@@ -17,12 +17,40 @@
         "ec2:CreateVpcEndpoint",
         "ec2:DeleteVpcEndpoints",
         "ec2:DescribeVpcEndpoints",
+        "ec2:DescribeVpcs",
         "ec2:ModifyVpcEndpoint",
+        "ec2:DescribeVpcEndpointServices",
+        "route53:ChangeResourceRecordSets",
```
dustman9000, line 23: Seems like there is duplication here with both the line below and the other section in Route53ManageRecords

PR #2201, opened 2024-08-27
Added package-operator metrics

robshelly:
### What type of PR is this?

Add package operator metrics to list of scraped metrics so alerting can be set up.

https://issues.redhat.com/browse/PKO-10

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


erdii:
@robshelly I think you need to re-run `make` and check in the resulting changes.

From the pipeline logs:

```
10:52:01 Running 'make' caused changes.  Run 'make' and commit changes to the PR to try again. If you're removing ACM policies, you need to remove the generated file from deploy/acm-policies/50-GENERATED-* before running 'make'.
```

typeid:
@robshelly can you give us a rough overview of the cardinality  / count of time series per cluster that this would ingest additionally into the osd tenant?

typeid:
So roughly 50 time series per classic cluster and ~100 time series per HCP in the current state?
That would then be roughly 100k time series for the whole fleet of classic clusters? Do we even already have alerting for the osd-tenant? 

@saswatamcode correct me if I'm wrong, but that sounds like too much.

@robshelly @erdii feel free to put a sync in my calendar and we can work out the alerting for PKO, there's a good chance we don't need to go through the osd rhobs tenant for this.

typeid:
Had a chat with @saswatamcode. RHOBS could scale to handle the extra series - we would have to tell them in advance when the metrics land, not a problem in general.

However, the current utilization for osd-observatorium-prod is at 600k series (with 3x replica), so adding 100k series would be an extra 300k series with x3 replica. A 50% increase for a single operator. IMHO we should rethink and update cardinality for what we want to ship.

typeid (approved):
/lgtm
/approve

PR #2204, opened 2024-08-30
wif template: defining support group

renan-campos:
The support group specifies the permissions the user needs to grant the SRE team to perform the SOPs needed to service OSD-GCP clusters. The granting of these permissions occurs on the client side after they have created a WifConfig resource.

### What type of PR is this?
_feature_

OCM-10387


typeid (approved):
/lgtm
/approve

PR #2208, opened 2024-09-05
to double backofflimit for delete-backplane-serviceaccounts

sdminonne:
### What type of PR is this?
This PR increase the backfofflimit for delete-backplane-serviceaccounts. It double the default value.
In various QE runs the job fails because of imagePullBackOff but in other runs the the POD after some image pull failures it runs OK.

Goal of this PR is to retry a little bit more.


bmeng:
not sure why the make was not run automatically, I did it local and create new PR for it https://github.com/openshift/managed-cluster-config/pull/2209

sdminonne:
Closing this

PR #2211, opened 2024-09-06
osd-logging[-unsupported]: applyBehavior: CreateOnly

2uasimojo:

The `osd-logging` and `osd-logging-unsupported` SelectorSyncSets include the `openshift-logging` namespace. Customers wishing to add labels to it -- e.g. to enable logging -- do not want such changes reverted whenever hive reapplies the syncset. The default `spec.applyBehavior`, `Apply`, causes hive to use `kubectl apply` to compute the patch to apply to the target object; and `kubectl apply` uses the
`kubectl.kubernetes.io/last-applied-configuration` annotation to decide whether to remove or ignore map entries such as labels. If the customer used `kubectl apply` to add their label, the result would be hive removing it.

Switch these two SelectorSyncSets to use `spec.applyBehavior: CreateOnly` instead. This causes hive to *not* use `kubectl apply` to compute the patch, and thus the
`kubectl.kubernetes.io/last-applied-configuration` annotation is ignored, and thus hive will only assert map values (such as labels) that are present in the syncset resources, and not remove others.

### What type of PR is this?
_bug_

_Fixes [HIVE-2577](https://issues.redhat.com//browse/HIVE-2577)_

NOTE: This setting applies to the whole of a SelectorSyncSet. This means that the other resources in these two are also affected: additions made by customers will not be reverted. If this turns out to be a problem, the resources may need to be split into separate [Selector]SyncSets with different `applyBehavior`s as appropriate.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-logging/config.yaml
+++ b/deploy/osd-logging/config.yaml
@@ -3,6 +3,12 @@ selectorSyncSet:
   # if this config no longer applies, don't delete the resources
   resourceApplyMode: "Upsert"
 
+  # Ignore the kubectl.kubernetes.io/last-applied-configuration
+  # annotation when patching resources. The result is that map additions
+  # made externally (e.g. by the customer) will be ignored rather than
+  # reverted.
+  applyBehavior: CreateOnly
+
   matchExpressions:
   # Enable in-cluster logging alerts for those clusters that already have logging installed
   # https://issues.redhat.com/browse/OSD-7564
--- a/deploy/osd-logging/unsupported/config.yaml
+++ b/deploy/osd-logging/unsupported/config.yaml
@@ -2,6 +2,13 @@ deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   # if we ever remove this, do not remove the resources
   resourceApplyMode: "Upsert"
+
+  # Ignore the kubectl.kubernetes.io/last-applied-configuration
+  # annotation when patching resources. The result is that map additions
+  # made externally (e.g. by the customer) will be ignored rather than
+  # reverted.
+  applyBehavior: CreateOnly
+
   matchExpressions:
   # Disable in-cluster logging alerts for those clusters that do not already have logging installed
   # We removed the version check because that would conflict with this check here for clusters that
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -30682,6 +30682,7 @@ objects:
         values:
         - 'true'
     resourceApplyMode: Upsert
+    applyBehavior: CreateOnly
     resources:
     - apiVersion: v1
       kind: Namespace
@@ -30927,6 +30928,7 @@ objects:
         values:
         - 'true'
     resourceApplyMode: Upsert
+    applyBehavior: CreateOnly
     resources:
     - apiVersion: v1
       kind: Namespace
--- a/hack/00-osd-managed-cluster-config-production.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-production.yaml.tmpl
@@ -30682,6 +30682,7 @@ objects:
         values:
         - 'true'
     resourceApplyMode: Upsert
+    applyBehavior: CreateOnly
     resources:
     - apiVersion: v1
       kind: Namespace
@@ -30927,6 +30928,7 @@ objects:
         values:
         - 'true'
     resourceApplyMode: Upsert
+    applyBehavior: CreateOnly
     resources:
     - apiVersion: v1
       kind: Namespace
--- a/hack/00-osd-managed-cluster-config-stage.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-stage.yaml.tmpl
@@ -30682,6 +30682,7 @@ objects:
         values:
         - 'true'
     resourceApplyMode: Upsert
+    applyBehavior: CreateOnly
     resources:
     - apiVersion: v1
       kind: Namespace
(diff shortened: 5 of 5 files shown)


typeid (commented):
Thanks @2uasimojo!

From issues.redhat.com/browse/OHSS-35884, it looks like we'll want this behavior for `openshift-operators-redhat` as well: 
```
"openshift.io/cluster-monitoring": "true" label to be preserved on the openshift-logging and openshift-operators-redhat namespaces
```
--
Regarding:
```
NOTE: This setting applies to the whole of a SelectorSyncSet. This means that the other resources in these two are also affected: additions made by customers will not be reverted. If this turns out to be a problem, the resources may need to be split into separate [Selector]SyncSets with different applyBehaviors as appropriate.
```
 I think we should indeed split out namespace creation, as this could otherwise lead to some odd behavior. 

Asked for an extra set of eyes on this review. :)

deploy/osd-logging/config.yaml
```diff
@@ -3,6 +3,12 @@ selectorSyncSet:
   # if this config no longer applies, don't delete the resources
   resourceApplyMode: "Upsert"
 
+  # Ignore the kubectl.kubernetes.io/last-applied-configuration
+  # annotation when patching resources. The result is that map additions
+  # made externally (e.g. by the customer) will be ignored rather than
+  # reverted.
+  applyBehavior: CreateOnly
```
jewzaam, line 10:
```suggestion
  applyBehavior: CreateOrUpdate
```

deploy/osd-logging/unsupported/config.yaml
```diff
@@ -2,6 +2,13 @@ deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   # if we ever remove this, do not remove the resources
   resourceApplyMode: "Upsert"
+
+  # Ignore the kubectl.kubernetes.io/last-applied-configuration
+  # annotation when patching resources. The result is that map additions
+  # made externally (e.g. by the customer) will be ignored rather than
+  # reverted.
+  applyBehavior: CreateOnly
```
jewzaam, line 10:
```suggestion
  applyBehavior: CreateOrUpdate
```

jewzaam (changes requested):
CreateOrUpdate will not use `oc apply` and will ensure what we care about continues to be updated / set.  CreateOnly will result in orphaned resources, we will never try to reconcile roles and rolebindings we care about.  I believe this will meet the goals from this issue, but I have not tested it.

```
    ApplyBehavior indicates how resources in this syncset will be applied to the
    target cluster. The default value of "Apply" indicates that resources should
    be applied using the 'oc apply' command. If no value is set, "Apply" is
    assumed. A value of "CreateOnly" indicates that the resource will only be
    created if it does not already exist in the target cluster. Otherwise, it
    will be left alone. A value of "CreateOrUpdate" indicates that the resource
    will be created/updated without the use of the 'oc apply' command, allowing
    larger resources to be synced, but losing some functionality of the 'oc
    apply' command such as the ability to remove annotations, labels, and other
    map entries in general.
```

2uasimojo:
/hold

Yeah, it turns out not to be this simple.

TL;DR: `CreateOrUpdate` would be the right thing if we could count on the `last-applied-configuration` annotation always being absent... but we can't, so it isn't.

More detailed writeup in the card, but I'm suggesting we split affected syncsets into three parts:
1. `CreateOnly` the Namespace with no frills.
2. Patch the Namespace with any labels/annotations we want to assert/ensure.
3. Put everything else in a separate syncset. IIUC these things -- Role[Binding], OperatorGroup, ConfigMap -- we don't want the user to muck with, so we should use `Apply`, which ought to revert any external changes.

Reviewers, does this track for you?

(I may punt the implementation to SRE-P as it involves deeper intricacies of MCC than I'm familiar with.)

jewzaam:
With the NS create + patch could "Sync" be used or would it need to be CreateOnly?  What is the behavior of patch in CreateOnly?  Guess the patch could be moved out to the RBAC SS if those had to be separate.

2uasimojo:
> With the NS create + patch could "Sync" be used or would it need to be CreateOnly?

I think you're conflating `resourceApplyMode` and `applyBehavior` (which is confusingly easy to do*):
- `resourceApplyMode` just controls whether we will delete the object if it is removed from the syncset (including if the whole syncset is removed). If `Sync`, we will; if `Upsert`, we won't. Since you're controlling the syncsets and (I'm guessing) don't intend to delete them, I would think it wouldn't matter which you choose here :shrug:
- `applyBehavior` is what determines whether we use `kubectl apply` vs `kubectl create` -- noting that `CreateOrUpdate` kind of does both, depending on factors outside of your control, and should thus probably be avoided.

> What is the behavior of patch in CreateOnly?

`applyBehavior` doesn't affect `patches` at all; and `patches` are done after `resources`/`secretMappings` within a given syncset; so it would be fine to include the annotation/label `patches` in the same syncset with the namespace `resources` with `CreateOnly`.

> Guess the patch could be moved out to the RBAC SS if those had to be separate.

That would also work, as long as you name the RBAC one to sort after the namespace one. But given the above, it isn't necessary, and it probably makes more sense visually to combine the namespace pieces into the same syncset.

*I've been trying to author a [doc chunk](https://github.com/openshift/hive/blob/20cfbb8ada310292af9fcbc8efca78da5f99deea/docs/syncset.md#how-to-use-applybehavior) to clarify all of this -- comments welcome!

2uasimojo:
@jewzaam Should we just move HIVE-2577 over to OSD/OHSS to be handled in your regular queue?

2uasimojo:
> @jewzaam Should we just move [HIVE-2577](https://issues.redhat.com//browse/HIVE-2577) over to OSD/OHSS to be handled in your regular queue?

This will be handled via [OSD-25576](https://issues.redhat.com/browse/OSD-25576).

PR #2214, opened 2024-09-11
adding permission to 4.17 wif template

renan-campos:
The osd-deployer service account needs the ability to get organization policies in order to run preflights.

[OCM-10387](https://issues.redhat.com//browse/OCM-10387)


renan-campos:
The new permission set was test by manually adding the permission to the service account that comprised the osd-deployer role.

PR #2216, opened 2024-09-12
Update ids-tester SA name to fix failed job runs

jonahbrawley:
### What type of PR is this?
_(bug)_

Updates SA name in `/deploy/osd-ids-tester/110-ids-tester-CronJob.yaml`, the original name was as not a service account that existed and jobs would fail.

[OSD-19877](https://issues.redhat.com//browse/OSD-19877)

- [X] Tested new cronjob definition in integration cluster
- [X] On manual run, job/pod completed successfully


rhdedgar (approved):
/lgtm

PR #2217, opened 2024-09-12
Updated permissions for WIF permissions

anispate:
Part of [OSD-24681](https://issues.redhat.com/browse/OSD-24681) ticket:

Updated permissions for WIF permissions as well by comparing using below cmd:
1. osdctl iampermissions diff -c gcp -b 4.16.0 -t 4.17.0-rc.1

### DIFF
--- a/resources/wif/4.17/vanilla.yaml
+++ b/resources/wif/4.17/vanilla.yaml
@@ -302,6 +302,9 @@ service_accounts:
           - storage.objects.delete
           - storage.objects.get
           - storage.objects.list
+          - resourcemanager.tagValueBindings.create
+          - resourcemanager.tagValues.get
+          - resourcemanager.tagValues.list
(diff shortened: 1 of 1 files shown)


resources/wif/4.17/vanilla.yaml
```diff
@@ -302,6 +302,9 @@ service_accounts:
           - storage.objects.delete
           - storage.objects.get
           - storage.objects.list
+          - resourcemanager.tagValueBindings.create
+          - resourcemanager.tagValues.get
+          - resourcemanager.tagValues.list
```
typeid, line 307: These are already included a few lines above. :) 

FYI: We created this WIF config as part of the WIF epic, the usual expectation when following the Gap Analysis SOP would be that this file doesn't yet exist - so this is a special case.

PR #2218, opened 2024-09-13
Permissions required for GCP Inquiries API

ckandag:
The osd-deployer service account needs these additional permissions to support various GCP Inquiries API that retrieve VPC , KMS Keyrings , Keys   invoked by UI component to display them in the console during deployment workflow.

Ref: https://issues.redhat.com/browse/OCM-10780


ckandag:
fyi @renan-campos

PR #2221, opened 2024-09-16
OSD-25331 - updated prom queries for KubeNodeUnschedulableSRE

vaidehi411:
### What type of PR is this?
Enhancement

updated prom query for  - KubeNodeUnschedulableSRE

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -92,7 +92,7 @@ spec:
       name: KubeNodeUnschedulableSRE
       resendWait: 6
       severity: Warning
-      summary: "Action required: a worker node is Unschedulable for more than an hour"
+      summary: "Action required: Atleast one worker nodes is unschedulable for more than an hour"
       references:
         - "https://docs.openshift.com/container-platform/latest/nodes/nodes/nodes-nodes-working.html#nodes-nodes-working-marking_nodes-nodes-working"
     - activeBody: |-
--- a/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
+++ b/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml
@@ -92,7 +92,7 @@ spec:
       name: KubeNodeUnschedulableSRE
       resendWait: 6
       severity: Warning
-      summary: "Action required: Atleast one worker nodes is unschedulable for more than an hour"
+      summary: "Action required: At least one worker node is unschedulable for more than an hour."
       references:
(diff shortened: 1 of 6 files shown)


iamkirkbater:
Can you explain more about why this change is necessary or update the PR description to link an OSD ticket?

I believe that this change will result in a 2 hour delay before the alert fires now, hence I'm wondering what problem this is attempting to solve. We will need to have the average over time condition satisfied for 1 hour and then the alert would have to be pending for an hour to actually fire.

vaidehi411:
@iamkirkbater 
Purpose of this change is to reduce the number of SL to one that ocm-agent sends in case multiple worker nodes are cordoned. 
"(avg_over_time(kube_node_spec_unschedulable[1h])) == 1"  will pick nodes that are unschedulable for entire last hour. 
We are not changing the alert fire timing here.

iamkirkbater:
> Purpose of this change is to reduce the number of SL to one that ocm-agent sends in case multiple worker nodes are cordoned.

Makes sense. Thank you for clarifying, I'll admit I should have read the ticket before adding that comment.

> "(avg_over_time(kube_node_spec_unschedulable[1h])) == 1" will pick nodes that are unschedulable for entire last hour. We are not changing the alert fire timing here.

Forgive me if I'm wrong, but I'm under the impression that we would be inadvertently changing the alert fire timing here.

My interpretation of the alert is that by changing to an average over time of 1h, that means that the node would need to be unscheduleable for a full hour before this condition is met. I think you're saying the same thing.

However, where I think we find the issue is that the "for" condition on the promrule is unchanged from 61 minutes, adding an additional until when the alert will actually fire resulting in a service log to the customer.

For example, with our previous alert expression we _immediately_ triggered the alert into a Pending state as soon as the node was unscheduleable. The alert would then have to be in a Pending state for 61 minutes before it moved into a "Firing" state, at which time OCM Agent would pick up on it and send the SL to the customer.

The way that I understand this change is that now we require the node to be unscheduleable for a full hour because of the avg_over_time function before it even triggers as a "Pending" alert, which would then take an additional 61 minutes to move into a Firing state where it would then send an SL to the customer.

With all of that said, this is all from memory. If you've tested this alert on an actual cluster and have seen that it will fire the SL to the customer in ~1h after being marked unscheduleable then please let me know and I'll cede my argument. If we haven't tested that as thoroughly, then I recommend we adjust the "avg_over_time" function's time to be 30 minutes and we adjust the "for" part of the promrule to 31 minutes, which would then mean that it has to be unscheduleable for 30 minutes to put the alert into a Pending state and then sit for another 31 minutes in Pending before moving to a Firing state, which would keep the timing inline with what we have today.

vaidehi411:
Thanks Krik for the detail explanation. I tested the change on the staging cluster and alert is firing after ~2hrs, which is updating the alert fire timing here.
It makes sense to me to change the avg_over_time function's time to 30 mins  and adjust the "for" part of the prom-rule to 31 minutes as per your suggestion.

PR #2224, opened 2024-09-18
OSD-22337 remove the namespaces which are no longer supported by SRE

tkong-redhat:
### What type of PR is this?
bug

Removing openshift-logging namespace from managed-namespace as this namespace is no longer supported by SRE.
Removing openshift-storage namespace in PDBErrorBudgetBurn alert as this namespace is no longer supported by SRE

_Fixes #_ [OSD-22337](https://issues.redhat.com/browse/OSD-22337)
Some customer defined alert will not be sent to RH pagerduty


tkong-redhat:
Close PR as openshift-logging namespace is still used by some legacy clusters

PR #2227, opened 2024-09-23
Update permissions for sre-managed-support role

zmird-r:
Adding PSC permissions to wif template.

Ticket: https://issues.redhat.com/browse/OSD-20986


renan-campos (approved):
/lgtm

PR #2231, opened 2024-09-26
Reduce ingress labeller schedule to every 5 minutes

hectorakemp:
### What type of PR is this?
_(bug)_

- Right now the 'Managed Ingress' feature labeller runs every hour. For the feature to function more effectively for customers, we have decided to bring this interval down to 5 minutes. This is so that there is no 'lag time' between a cluster being upgraded and a customer being able to use the new features in `rosa edit ingress`. 

_Fixes #_
https://issues.redhat.com/browse/OSD-25776
https://redhat-internal.slack.com/archives/CCX9DB894/p1727358496734429?thread_ts=1727202789.794229&cid=CCX9DB894

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


mjlshen (approved):
/lgtm

PR #2232, opened 2024-09-27
Move CreateTags permission for ENIs to match Instances

fahlmant:
### What type of PR is this?
Refactor

This moves the permissions for creating tags on ENIs to match the same restrictions for tagging Instances. Specifically, this change meets the following criteria:
- ENIs must be tagged with red-hat-managed: true 
- ENIs cannot be tagged after creation
- ENIs can only be tagged when created as part of a call to RunInstances (creating an EC2 instance).


jharrington22 (commented):
This LGTM for when we are ready to merge.

PR #2235, opened 2024-10-02
Increase capa annotator cronjob frequency 

rafael-azevedo:
Increases the capa annotator cronjob frequency to decrease cluster worker node start up times.


fahlmant (approved):
/lgtm
/approve

PR #2223, opened 2024-09-16
Revert "Add cronjob to olm-dance openshift-ocm-agent-operator."

bergmannf:
This reverts commit f38430285cee9b805df5b0b40b750c16dd0a0b67.

### What type of PR is this?
cleanup

### Which Jira/Github issue(s) this PR fixes?

--

### Pre-checks (if applicable):
- [ ] Tested latest changes against a cluster
- [ ] Included documentation changes with PR
- [ ] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
    - key: api.openshift.com/fedramp
      operator: NotIn
      values: ["true"]
    ```

PR #2237, opened 2024-10-09
Add cronjob to olm-dance openshift-ocm-agent-operator

xiaoyu74:
### What type of PR is this?
_(bug)_

OLM dances the openshift-ocm-agent-operator installation.

https://issues.redhat.com//browse/OSD-25526

### DIFF
--- a/deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
+++ b/deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: sre-operator-reinstall
+  namespace: openshift-ocm-agent-operator
+spec:
+  ttlSecondsAfterFinished: 100
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "* * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: sre-operator-reinstall-sa
+          restartPolicy: Never
+          containers:
+          - name: operator-reinstaller
+            image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+            imagePullPolicy: Always
+            command:
+            - sh
+            - -c
+            - |
+              #!/bin/bash
+              set -euxo pipefail
+              NAMESPACE=openshift-ocm-agent-operator
+
+              # Check for the status OAO v0.1.278 version (previous version:
+              # ocm-agent-operator.v0.1.276-g09f3778). If we it's not succeeded,
+              # do OLM dance The resources will be re-created with the latest
+              # version
+              CSV_PHASE=$(oc -n openshift-ocm-agent-operator get csv ocm-agent-operator.v0.1.289-g742e13a -ojsonpath='{.status.phase}{"\n"}') || true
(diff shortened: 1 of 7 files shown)


deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+                # Recreate the resources to re-install the operator
+                oc -n "$NAMESPACE" create -f /tmp/01-ocm-agent-operator-subs.yaml
+                oc -n "$NAMESPACE" create -f /tmp/02-ocm-agent-operator-registry.yaml
+                oc -n "$NAMESPACE" create -f /tmp/03-ocm-agent-operator-og.yaml
+              else
+                echo "Skipping OLM dance as OAO already successully installed with latest version or expected version was not found."
+              fi
+
+              # Prevent the job from -rerunning
```
bmeng, line 60: we do not need to prevent here? The operator should be running with the expected version after the OLM dance.
Deleting the service account which is used by itself sounds a bit weird

deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
```diff
@@ -0,0 +1,64 @@
...
+              #CSV_STATUS=$(oc -n openshift-ocm-agent-operator get csv -l operators.coreos.com/ocm-agent-operator.openshift-ocm-agent-operator= -oname;echo $?)
+              # This does nothing if the expected CSV is not found.
+              if [[ -n ${CSV_PHASE} && ${CSV_PHASE} != "Succeeded" ]]; then
+                # Take backup of the files to be recreated (subscription, catalogsource and operatorgroup)
+                oc -n "$NAMESPACE" get subs ocm-agent-operator -oyaml > /tmp/01-ocm-agent-operator-subs.yaml
+                oc -n "$NAMESPACE" get catsrc ocm-agent-operator-registry -oyaml > /tmp/02-ocm-agent-operator-registry.yaml
+                oc -n "$NAMESPACE" get og ocm-agent-operator-og -oyaml > /tmp/03-ocm-agent-operator-og.yaml
+
+                oc -n "$NAMESPACE" delete managednotification --all
+                oc -n "$NAMESPACE" delete managedfleetnotifications --all
```
bmeng, line 45: I am not sure why we need to delete the CRs, but do we need to delete the CRD as well?

deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
bmeng: And we need to clean the whole job folder later anyway

deploy/osd-25526-oao-olm-dance/01-cronjob.yaml
xiaoyu74: > I am not sure why we need to delete the CRs, but do we need to delete the CRD as well?

Yeah, if the OAO can be reinstalled without deleting the CRs, I think it's safer to avoid deleting them

deploy/osd-25526-oao-olm-dance/readme.md
```diff
@@ -0,0 +1,3 @@
+# README
+
+This cronjob is a fix for https://issues.redhat.com/browse/OSD-25526 and ITN-193
```
ravitri, line 3: FYI, we can track the OLM dance requirement as per https://issues.redhat.com/browse/OSD-25349

bergmannf:
We added that CronJob to one-time OLM dance and then removed it again in https://github.com/openshift/managed-cluster-config/pull/2223 - are we sure we want to permanently reintroduce this one?

PR #2239, opened 2024-10-14
Remove unnecessary wif configs and rename SA ids and Role ids in 4.17…

zmird-r:
Renaming WIF config IDs so that they fit character limits:
- 50 chars for SA IDs
- 25 chars for Role IDs

The following changes were made:
- remove the "openshift" string from SA and role IDs
- rename "operator" string to "op"
- make role names more descriptions since they have a higher char limit

Ticket: https://issues.redhat.com/browse/OSD-25171


renan-campos (approved):
/approve

PR #2242, opened 2024-10-16
WIF : Add additional permission for PrivateServiceConnect

ckandag:
Additional permissions needed for PrivateServiceConnect usecases


typeid:
@ckandag Thanks! 

Could you please link a JIRA to this and add an explanation on where this is used? I'm assuming this is needed by OCM when using the installer role?

PR #2260, opened 2024-11-01
Add forwardingrules.use to wif deployer role

ckandag:
message\":\"error creating the Service Attachment: googleapi: Error 403: Required 'compute.forwardingRules.use' permission for 'projects/openshift-qe-group-i-osd/regions/us-east1/forwardingRules/u9a3q5n1h9y2k5d-9grwd-api-internal'


ckandag:
@renan-campos @zmird-r

PR #2262, opened 2024-11-06
Change namespace creation for osd-logging

petrkotas:
### What type of PR is this?
BUG

Moved namespace creation from osd-logging to be create only. Additional resources are created in patches.

This is change related to bug:
- https://issues.redhat.com/browse/OSD-25576

_Fixes #_ https://issues.redhat.com/browse/OSD-25576

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


petrkotas:
@jewzaam and @2uasimojo would you please be able to review this? Thank you.

jewzaam (changes requested):
/hold

This changes from creation of resources to patching only.  Implication is any place where this is applied for the first time for a cluster it will do nothing because there is nothing to patch.  JIRA is the right place to sort out the plan for how this is applied and implications to existing or new clusters.  Not clear right now what the requirements are.

petrkotas:
After the thorough analysis I came to conclusion that it is best to leave this one as won't do. The outcome of leaving it in current state is only a warning on user workload monitoring for missing label on managed namespace. Potential fallout for implementing the change is unknown. I tried to gather information on how the namespace is used, without 100% success. It should be OK, but we lack the full understanding if there is something observing. Also implementing this change, splits the current syncsets into two and breaking the pattern already used. Which is a potential for errors.

Finally, with the move to hypershift, this is no longer an issue.

Thus the pros of leaving it alone out weights the cons.

PR #2286, opened 2024-12-03
Make sure dns-default pods run on all nodes for 4.18+ clusters

mmazur:
### What type of PR is this?
fix

DaemonSets iptables-alerter (-n openshift-network-operator) and dns-default (-n openshift-dns) should run across all nodes. But for OSD/ROSA Classic they don't due to us having infra nodes and the DSs' tolerations not accounting for that properly (see [this thread](https://redhat-internal.slack.com/archives/C07QEA1PDFX/p1733226251687799?thread_ts=1733162994.979289&cid=C07QEA1PDFX)).

Apparently this has not been an issue in practice thus far, but we probably want this to work as intended going forward, so I've enabled this for 4.18+ clusters only.

Note: this PR only addresses the dns-default pods, since it's not possible to effectively patch the iptables-alerter DS, as that's tightly controlled by its operator and getsquickly reverted.

_Fixes [OSD-26887](https://issues.redhat.com/browse/OSD-26887)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-26887-deamonset-tolerations/config.yaml
+++ b/deploy/osd-26887-deamonset-tolerations/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: NotIn
+    values: ["4.6", "4.7", "4.8", "4.9", "4.10", "4.11", "4.12", "4.13", "4.14", "4.15", "4.16", "4.17"]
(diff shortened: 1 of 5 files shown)


bmeng:
I remember @bergmannf was working on the similar thing and turned out it is not working well.
https://github.com/openshift/managed-cluster-config/pull/2144
https://github.com/openshift/managed-cluster-config/pull/2158

bergmannf:
Thanks for the ping @bmeng - Yes there is an [ADR](https://docs.google.com/document/d/1xuJbNYRsd7uprtPhtTuZnf1pSmMNGDV0wfvO8CGrVXQ/edit?tab=t.0#heading=h.fbts9k7ez097) about this, because there are strange edge-cases where we had issues with customer workloads that led to a big escalation.
In case we run this on *all* nodes it can lead to issues with high-churn clusters - and without the ADR we break the customer being able to decide where to run the pods.

PR #2290, opened 2024-12-04
Sync OWNERS_ALIASES with boilerplate

abyrne55:
This cleanup PR syncs OWNERS_ALIASES with that of openshift/boilerplate, specifically [this file](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/OWNERS_ALIASES). The boilerplate version was ordered differently than this repo's, so the diff looks a lot more dramatic than it actually is (feel free to use [yamldiff](https://www.yamldiff.com/) to verify). The new version also renames the `sd-architects` group to `srep-architects`, so this PR also patches the only file in this repo that used that group alias.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,5 +1,7 @@
+# ================================ DO NOT EDIT ================================
+# This file is managed in https://github.com/openshift/boilerplate
 # See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#OWNERS_ALIASES
-
+# =============================================================================
 aliases:
   srep-functional-team-aurora:
     - abyrne55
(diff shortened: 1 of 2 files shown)


xiaoyu74:
Thanks Anthony for the PR and the detailed description. Ran yamldiff with `boilerplate` and only shows the 3 sections difference:
- sre-alert-sme:
- sre-mst-observatorium:
- srep-region-leads
Which should be expected as MCC-specific group aliases.

Just one nit, looks `srep-region-leads` list is not updated, @devppratik has taken over @MitaliBhalla for APAC West OSD/ROSA.

joshbranham:
/check-required-labels |  
-- | --


<br class="Apple-interchange-newline">

joshbranham:
@abyrne55 wanna rebase this so it can merge? Also idk why @iamkirkbater approval is not showing up, but maybe that gets resolved post-rebase.

PR #2293, opened 2024-12-06
OSD-26958 must gather operator re-install

luis-falcon:
Cronjob task build by following reinstall Operator SOP https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/reinstall-operator.md

Solves https://issues.redhat.com/browse/OSD-26958

### What type of PR is this?
Bug


joshbranham:
Couple notes here:

- This needs to be a new directory under `deploy`, not a CronJob attached to the operator itself
- Checking this manually against a cluster shows the script will fail, as in my case, there are not two CSVs for the operator. Opened https://github.com/openshift/ops-sop/pull/3378 to track updating the SOP

luis-falcon:
/

PR #2294, opened 2024-12-06
[OCM-12870] define management cluster cronjob to perform olm dance

cdoan1:
### What type of PR is this?
_cleanup_

When fleet manager enables the upgrade of OADP 1.2 -> OADP 1.3, OLM may get stuck.
introduce a new cronjob to do OLM dance.

_[OCM-12870](https://issues.redhat.com//browse/OCM-12870)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocm-12870/02-CronJob.yaml
+++ b/deploy/ocm-12870/02-CronJob.yaml
@@ -1,0 +1,61 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: sre-operator-reinstall
+  namespace: openshift-adp-operator
+spec:
+  ttlSecondsAfterFinished: 100
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "* * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: sre-operator-reinstall-sa
+          restartPolicy: Never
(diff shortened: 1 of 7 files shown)


cdoan1:
This MR may need to wait for FM promotion

deploy/ocm-12870/02-CronJob.yaml
```diff
@@ -0,0 +1,61 @@
...
+              #!/bin/bash
+              set -euxo pipefail
+              NAMESPACE=openshift-adp-operator
+
+              # Check for the status OADP v1.3.4 version 
+              # (previous version: ocm-agent-operator.v1.2.x). 
+              # If we it's not succeeded, do OLM dance The resources will be re-created 
+              # with the latest version
+
+              CSV_PHASE=$(oc -n openshift-adp-operator get csv oadp-operator.v1.3.4 -ojsonpath='{.status.phase}{"\n"}') || true
```
bergmannf, line 36: I think this won't work, as there will be 2 CSVs (one for 1.2 and one for 1.3 and I think both might have to be removed).

bmeng:
@cdoan1 do we still need this?

joshbranham:
This is using the old process for a reinstall, closing to limit risk. Please check the SOP if you need this still https://github.com/openshift/ops-sop/blob/master/v4/knowledge_base/reinstall-operator.md

/close

PR #2305, opened 2024-12-16
Fix oao olm dance csv/installplan listing

typeid:
### What type of PR is this?
bug

One last awkward PR to fix the install plan listing. Had tested on the situation we mostly have on production where the csv & IP don't exist, but found that this doesn't work when there is a csv and IP. We'd still want to olm dance in that case. 

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-26960/01-cronjob.yaml
+++ b/deploy/osd-26960/01-cronjob.yaml
@@ -31,11 +31,10 @@ spec:
 
                   # Path for temporary subscription JSON backup
                   SUB_JSON_PATH="/tmp/sub.json"
-    
-                  # The OLM dance
 
+                  # The OLM dance
                   # Delete CSVs if they exist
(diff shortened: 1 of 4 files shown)


deploy/osd-26960/01-cronjob.yaml
```diff
@@ -31,11 +31,10 @@ spec:
 
                   # Path for temporary subscription JSON backup
                   SUB_JSON_PATH="/tmp/sub.json"
-    
-                  # The OLM dance
 
+                  # The OLM dance
                   # Delete CSVs if they exist
-                  CSV_LIST=$(oc get clusterserviceversions.operators.coreos.com -n "$NAMESPACE" | grep "$OPERATOR" || true | awk '{print $1}')
+                  CSV_LIST=$(oc get clusterserviceversions.operators.coreos.com -n "$NAMESPACE" --no-headers | grep "$OPERATOR" | awk '{print $1}' || echo "")
```
typeid, line 37: We use `set -e` so we need to dodge the errors when nothing is found, thus `|| echo ""`

PR #2309, opened 2024-12-17
OSD-27279: Ensure ClusterMonitoringErrorBudgetBurnSRE alert excludes user-workload-monitoring failure reasons

joshbranham:
### What type of PR is this?
bug

We are currently being paged for `ClusterMonitoringErrorBudgetBurnSRE` even when the reason is an issue with the customer's UWM stack. Based on this [commit](https://github.com/openshift/managed-cluster-config/commit/6e8aae3b4711f13276adc8e4b642d712d13c879b#diff-51cf7e84af79c0ddc6db3ca4171c8b06df42b8a76e6f879b6d137c46181d2fb5R15), it would seem we meant to _exclude_ reasons that are like `reason!~".*UserWorkload.*"` for SRE alerts, and _include_ those only for the customer specific ones.

This PR updates our rules to take that into account and should prevent SRE from being paged.

Sample [alert](https://redhat.pagerduty.com/incidents/Q2LZFIGV1KZVZ2).

Fixes https://issues.redhat.com/browse/OSD-27279

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/sre-prometheus/monitoring/100-cluster-monitoring-error-budget-burn.yaml
+++ b/deploy/sre-prometheus/monitoring/100-cluster-monitoring-error-budget-burn.yaml
@@ -10,13 +10,13 @@ spec:
   groups:
   - name: sre-cluster-monitoring-error-budget-burn
     rules:
-    # Substitues ClusterOperatorDown{name="monitoring"} as part of https://issues.redhat.com/browse/OSD-19769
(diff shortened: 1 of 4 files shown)


joshbranham:
> Great catch, @joshbranham! Thanks for the PR!
> 
> /lgtm

Gotta throw some kudos to @tnierman too for helping to dig in and spotting the lacking `.*` for the SRE specific alerts 🥇

iamkirkbater:
/hold cancel

Ship it!

PR #2330, opened 2025-01-22
Added sts admin ack for 4.18

btoll:
### What type of PR is this?
bug

This fixes [OSD-27689](https://issues.redhat.com//browse/OSD-27689). The SOP also needs to be updated, as it reads now that an admin ack only needs to be created when there are permission differences between 4.17 and 4.18 (there weren't any).
https://github.com/openshift/ops-sop/blob/master/v4/howto/gap-analysis.md#identify-and-put-in-place-upgrade-gates-and-admin-acks

[OSD-27689](https://issues.redhat.com//browse/OSD-27689)

### DIFF
--- a/deploy/osd-cluster-acks/sts/4.18/config.yaml
+++ b/deploy/osd-cluster-acks/sts/4.18/config.yaml
@@ -1,0 +1,9 @@
+deploymentMode: SelectorSyncSet
+selectorSyncSet:
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: In
+    values: ["4.17"]
+  - key: api.openshift.com/sts
+    operator: In
+    values: ["true"]
--- a/deploy/osd-cluster-acks/sts/4.18/osd-sts-ack_CloudCredential.yaml
+++ b/deploy/osd-cluster-acks/sts/4.18/osd-sts-ack_CloudCredential.yaml
@@ -1,0 +1,6 @@
+apiVersion: operator.openshift.io/v1
+kind: CloudCredential
+name: cluster
+applyMode: AlwaysApply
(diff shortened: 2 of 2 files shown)


fahlmant:
Another part of the doc does say
```
There should always be an OCM Version Gate corresponding to the STS Admin Ack, even if there are no new permissions. Add additional version gates as needed for any existing OCP Admin Acks.
```
So it looks like we just need a little cleanup

btoll:
> uld always be an OCM Version Gate correspondin

Thanks, I'll also include that.

iamkirkbater:
Looks like the only thing that's failing is the autogenerate `make` step.

If you run `make` locally with your changes and then commit the generated files that should fix it.

You can also enable github actions on your fork which should add an `on push: make` step where every time you push changes to your fork the github action runs the `make` command for you and commits those changes.

iamkirkbater:
Also, as an FYI we most likely will encounter a similar scenario for GCP WIF clusters - I'd assume at least. We'll ~~burn~~ cross that bridge when we get to it but just want to give you a heads up if we see this for those clusters as well.

btoll:
I had noticed that the call to the `make` target had failed, because it appears to have ran successfully on this one that was closed b/c I had accidentally deleted my fork (derp): https://github.com/openshift/managed-cluster-config/pull/2329

Some questions:
1.  Is the PR that was closed a complete and correct example of the fix for this?
2.  Are admin gates completely automated when an ack is added?  I've read a lot of the documentation, but I'm still unclear about this.

Thanks

btoll:
I just manually ran `make` and add the three files that were changed in the `hack/` directory and force pushed.

btoll:
@iamkirkbater @fahlmant Could you guys review this, please?

PR #2338, opened 2025-01-24
Remove architects from TL alias

cblecker:
### What type of PR is this?
cleanup

Removes @cblecker, @jewzaam, and @jharrington22 from the TL alias in OWNERS_ALIASES. This ensures that review pings go to the right group. For folders where architect oversight is needed, there are other aliases that are actively used, so this is redundant.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -24,6 +24,3 @@ approvers:
 - typeid
 - xiaoyu74
 - zmird-r
-maintainers:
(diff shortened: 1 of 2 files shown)


cblecker:
/hold cancel
confirmed with @jewzaam and @jharrington22 offline

typeid (approved):
/lgtm

PR #2353, opened 2025-01-31
OSD-27987: Add CronJob to patch control-plane-operator on 4.17 HCP clusters

joshbranham:
### What type of PR is this?
_feature_

In order to workaround OCPBUGS-48519, SRE needs to pin the `control-plane-pki-operator` to a specific Z stream that includes the bugfix. The mechanism to do this is by setting the annotation `hypershift.openshift.io/control-plane-operator-image` on the `HostedCluster` resource.

We do this by creating a CronJob on the service clusters that runs every 5min and enumerates all `ManifestWork`s to patch the nested `HostedCluster` resource with the annotation.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


deploy/ocpbugs-48519/10-CronJob-4-17.yaml
```diff
@@ -0,0 +1,76 @@
...
+              imagePullPolicy: Always
+              securityContext:
+                allowPrivilegeEscalation: false
+                capabilities:
+                  drop:
+                    - ALL
+                runAsNonRoot: true
+              env:
+                - name: IMAGE
+                  # 4.17.15
```
wking, line 44: You could make this comment:

```golang
# oc adm release info --image-for hypershift quay.io/openshift-release-dev/ocp-release:4.17.15-multi
```

just so it's extra clear where the pullspec comes from, in case folks want to bump it in the future?

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
```diff
@@ -0,0 +1,76 @@
...
+                  for clusterID in ${managedclusters[@]};do 
+                    # Extract namespace and name
+                    namespace=$(oc get managedclusters $clusterID -o json | jq -r '.metadata.labels["api.openshift.com/management-cluster"]')
+                    kinds=$(oc get manifestwork -n $namespace $clusterID -o json | jq -r '.spec.workload.manifests[].kind')
+                    num=0 
+                    for kind in $kinds;do 
+                      if [[ $kind == "HostedCluster" ]]; then
+                          echo $clusterID
+                          #~1 escapes / in bash
+                          json_payload='[{"op":"replace","path":"/spec/workload/manifests/'"$num"'/metadata/annotations/hypershift.openshift.io~1control-plane-operator-image","value":"'"$IMAGE"'"}]'
```
wking, line 65: I don't do enough [JSON Patch][1] to have internalized this, so I checked the `~1` escaping, and [it's JSON Pointer for `/`][2], to set up the expected `hypershift.openshift.io/control-plane-operator-image` annotation 👍 

[1]: https://www.rfc-editor.org/rfc/rfc6902
[2]: https://www.rfc-editor.org/rfc/rfc6901#section-3

joshbranham: Yeah this is just copied from another CronJob we have doing this same behavior, I did not look into it too much besides confirming it works as expected.

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
```diff
@@ -0,0 +1,76 @@
...
+                          echo "oc patch manifestwork $clusterID -n $namespace --type='json' -p "$json_payload""
+                          oc patch manifestwork $clusterID -n $namespace --type='json' -p "$json_payload"
```
wking, line 67: `patch` is probably not a big deal for the API server when it's a no-op, but would still get logged in Kube API audit logs.  Since we already get the ManifestWork above, maybe we can set this up to only patch when we know there's a change?  Something like:

```bash
oc get manifestwork -n $namespace $clusterID -o json | jq -r '.spec.workload.manifests[] | .kind + " " + .metadata.annotations["hypershift.openshift.io/control-plane-operator-image"]' | while read KIND ANNOTATION; do
  if [[ $kind == "HostedCluster" ]]; then
    if test "${IMAGE}" = "${ANNOTATION}"; then
      echo "$clusterID already has hypershift.openshift.io/control-plane-operator-image set to ${ANNOTATION}; no change needed"
      continue
    fi
    ...
```

or something?

joshbranham: I am not too worried about optimizing this, at least with the way I have it now we ensure no one else can override this value for longer than the CronJob duration.

typeid (commented):
One question added, in general it looks okay to me but there is one thing I'm particularly worried about: patching manifestworks. 
I understand in this case it's likely more widely tested than when an SRE does it on the fly, but it's an operation that in my opinion is not worth the risk vs just annotating the HostedCluster objects directly using `oc annotate`.

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
```diff
@@ -0,0 +1,76 @@
...
+                capabilities:
+                  drop:
+                    - ALL
+                runAsNonRoot: true
+              env:
+                - name: IMAGE
+                  # oc adm release info --image-for hypershift quay.io/openshift-release-dev/ocp-release:4.17.15-multi
+                  value: "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:1ac8880940e256abb01cf8d8276beeb53db37572ab7369a752de5af244bc24e4"
+                - name: MAJOR_MINOR_VER
+                  value: "4.17"
```
typeid, line 47: We'll likely want to have this down to z versions, e.g. this should apply until 4.17.14, as 4.17.15 will have the fix.

joshbranham: Following the pattern we established when pinning CAPA, we opted to just pin all Y streams to the latest Z that has the fix.

To do what you're proposing would require us to find <=4.17.14 and then pin correct? I'll see if I can come up with an easy way to do that in bash/jq

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
```diff
@@ -0,0 +1,76 @@
...
+                  for clusterID in ${managedclusters[@]};do 
+                    # Extract namespace and name
+                    namespace=$(oc get managedclusters $clusterID -o json | jq -r '.metadata.labels["api.openshift.com/management-cluster"]')
+                    kinds=$(oc get manifestwork -n $namespace $clusterID -o json | jq -r '.spec.workload.manifests[].kind')
+                    num=0 
+                    for kind in $kinds;do 
+                      if [[ $kind == "HostedCluster" ]]; then
+                          echo $clusterID
+                          #~1 escapes / in bash
+                          json_payload='[{"op":"replace","path":"/spec/workload/manifests/'"$num"'/metadata/annotations/hypershift.openshift.io~1control-plane-operator-image","value":"'"$IMAGE"'"}]'
```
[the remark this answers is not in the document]
joshbranham: It only changes the image for the control-plane-operator and the control-plane-pki-operator. And yes, per Hypershift team, we can't pin just PKI operator.

joshbranham:
> One question added, in general it looks okay to me but there is one thing I'm particularly worried about: patching manifestworks. 
> 
> I understand in this case it's likely more widely tested than when an SRE does it on the fly, but it's an operation that in my opinion is not worth the risk vs just annotating the HostedCluster objects directly using `oc annotate`. 
> 
> 
> 
> 

Since we are automating this, testing in stage, have a known version of this running for ~3mo on the fleet, I'm not too worried about patching MW. In my mind, MW is the source of truth for the HostedCluster, so patching it here means we ensure our annotation is not overridden easily.

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
typeid: Thanks for looking into it! Without this, we'd be effectively downgrading the version for everything that comes after and might run into some future issues.

joshbranham: This is now handled by a wrapper python script, which includes some unit tests to handle a couple scenarios.

typeid (approved):
/approve

I'm still hesitant about patching the MFW. We have a PMR action item to explicitly dissallow this. However, this is tested and already running for another override annotation on our fleet so let's not block on that. 

I'll leave some time for others to look and then lgtm.

fahlmant:
`We have a PMR action item to explicitly dissallow this.` Where? Manually, yes. Automated, no, afaiu.

typeid:
@fahlmant I stand corrected:
```Remove access/permissions or using a webhook to block **manual** modifications
```

@joshbranham mentioned that he's working on adding the annotation removal logic for clusters that make it out of the z stream range that need the override, which makes perfect sense to include as part of this PR. Let's wait for that and we can merge it if everyone's happy with it :)

deploy/ocpbugs-48519/10-CronJob-4-17.yaml
joshbranham: Updated script to account for removing the annotation when it is no longer needed.

PR #2359, opened 2025-02-10
OSD-26790: Adding UpgradeNotificationFailedSRE alert

TheUndeadKing:
### What type of PR is this?
_(bug)_

TLDR: When the user gets banned (OR) PS has been rotated in a middle of the upgrade, the upgrade will stall. We're creating a metric if MUO unable to notify it's state to the OCM. This metric will help us to track via PD, so that we can unblock the upgrade.

In previous [PR-494](https://github.com/openshift/managed-upgrade-operator/pull/494), we'd implemented the logic to enable upgrade_notification_failed metric. In this PR, we're triggering UpgradeNotificationFailedSRE alert, if we cluster unable to notify it's state more than 30 mins

_Fixes #_ [OSD-26790](https://issues.redhat.com//browse/OSD-26790)

### DIFF
--- a/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml
@@ -62,3 +62,12 @@ spec:
       annotations:
         summary: "UpgradeConfig has not successfully synced in 4 hours."
         description: "This clusters UpgradeConfig has not been synced in 4 hours and may be out of date"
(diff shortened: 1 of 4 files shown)


devppratik (changes requested):

a7vicky:
Could we use templating so that we get to know the state it failed to send notification to?
https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#templating

typeid (approved):
/lgtm
/approve

PR #2368, opened 2025-02-27
OSD-27692: New crontab to clean backplane remediation RBACs in case the remediation was not properly deleted

Nikokolas3270:
### What type of PR is this?
_feature_

Remediations created through `ocm-backplane remediation create` need to be deleted once the investigation is finished.
Remediations are normally deleted running or programmatically running (for CAD):
```
ocm-backplane remediation delete
```
If not done, RBAC resources created during `ocm-backplane remediation create` will be leaked forever (except the service account which is already deleted after 720 minutes)

The purpose of this new crontab is to delete any resource:
- Labelled with `managed.openshift.io/remediation` (whatever the value)
- Which is more than 720 minutes == 12 hours old
- Which is of one of this type:
   - `clusterrole`
   - `clusterrolebinding`
   - `role`
   - `rolebinding` 

_Implemtens [OSD-27692](https://issues.redhat.com//browse/OSD-27692)

- [ ] Not a FedRAMP change

### DIFF
--- a/deploy/osd-delete-backplane-remediation-rbacs/00-delete-backplane-remediation-rbacs.namespace.yml
+++ b/deploy/osd-delete-backplane-remediation-rbacs/00-delete-backplane-remediation-rbacs.namespace.yml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-backplane
--- a/deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
+++ b/deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
@@ -1,0 +1,68 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: osd-delete-backplane-remediation-rbacs
+  namespace: openshift-backplane
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      ttlSecondsAfterFinished: 86400
+      backoffLimit: 12
+      template:
+        spec:
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - preference:
+                  matchExpressions:
+                  - key: node-role.kubernetes.io/infra
+                    operator: Exists
+                weight: 1
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/infra
+              operator: Exists
+          restartPolicy: Never
+          serviceAccountName: osd-delete-backplane-remediation-rbacs
+          containers:
+            - name: osd-delete-backplane-remediation-rbacs
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  now=$(date +%s);
+                  expiryMins=720;
+                  for line in $(oc get rolebinding -A -l managed.openshift.io/remediation -o custom-columns=:.metadata.namespace,:.metadata.name --no-headers);do
+                    ns=$(echo "$line" | awk '{print $1}')
+                    rolebinding=$(echo "$line" | awk '{print $2}')
+                    creation=$(date -d $(oc get rolebinding $rolebinding -n $ns -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                    if [[ $(expr $((now-creation)) \> $expiryMins \* 60) == 1 ]]; then
+                      oc delete rolebinding $rolebinding -n $ns
+                    fi
+                  done;
+                  for line in $(oc get role oc get role -A -l managed.openshift.io/remediation -o custom-columns=:.metadata.namespace,:.metadata.name --no-headers);do
+                    ns=$(echo "$line" | awk '{print $1}')
+                    role=$(echo "$line" | awk '{print $2}')
+                    creation=$(date -d $(oc get role $role -n $ns -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                    if [[ $(expr $((now-creation)) \> $expiryMins \* 60) == 1 ]]; then
+                      oc delete role $role -n $ns
+                    fi
+                  done;
+                  for clusterrolebinding in $(oc get clusterrolebinding -l managed.openshift.io/remediation --no-headers | awk '{print $1}');do
+                    creation=$(date -d $(oc get clusterrolebinding $clusterrolebinding -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                    if [[ $(expr $((now-creation)) \> $expiryMins \* 60) == 1 ]]; then
+                      oc delete clusterrolebinding $clusterrolebinding
+                    fi
+                  done;
+                  for clusterrole in $(oc get clusterrole -l managed.openshift.io/remediation --no-headers | awk '{print $1}');do
+                    creation=$(date -d $(oc get clusterrole $clusterrole -o jsonpath='{.metadata.creationTimestamp}') '+%s');
+                    if [[ $(expr $((now-creation)) \> $expiryMins \* 60) == 1 ]]; then
+                      oc delete clusterrole $clusterrole
+                    fi
+                  done;
--- a/deploy/osd-delete-backplane-remediation-rbacs/10-delete-backplane-remediation-rbacs.rbac.yaml
+++ b/deploy/osd-delete-backplane-remediation-rbacs/10-delete-backplane-remediation-rbacs.rbac.yaml
@@ -1,0 +1,35 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: osd-delete-backplane-remediation-rbacs
+  namespace: openshift-backplane
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: osd-delete-backplane-remediation-rbacs
+rules:
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - roles
+      - rolebindings
+      - clusterroles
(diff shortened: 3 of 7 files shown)


Nikokolas3270:
Lets put this change on hold for now as we may actually want to do the cleanup from backplane API when creating a new remediation.

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
```diff
@@ -0,0 +1,66 @@
...
+            - name: osd-delete-backplane-remediation-rbacs
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  now=$(date +%s);
+                  expiryMins=720;
+                  for ns in $(oc get ns --no-headers | grep -e ^openshift-.* | awk '{print $1}');do
```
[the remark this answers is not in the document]
Nikokolas3270: The remediations are NOT creating RBAC roles et rolebindings there...
... at least for now.
Actually remediations are currently only creating clusterroles and clusterrolebindings.
Do you think some new remediation may create role & rolebinding there?
Asking because I am no expert on that. If "yes", then I will change the filter as you suggest... the important thing being that there is a filter in order to be able to bound the time taken by the cleanup script (what if the customer creates tons of namespaces for instance)

Nikokolas3270:
> Looks good to me otherwise, did you to test these changes on a cluster? I think we want to do good QE on this PR we are deleting clusterroles after all ;)

Yes, I tested it with a syncset deployed on hive. I lowered the cleanup period from 12h to 5m :)

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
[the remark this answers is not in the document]
RaphaelBut: True, currently there are no investigations for those namespaces ( or even with roles haha )
but i expect we will have some soon and would propose to add those 5 ns to the filter 

The filter is definitely needed for the reason you mention :D 
Although, we might actually want to target other namespaces but its not something for the near future and can be tackled in the followup backplane-api implementation ^^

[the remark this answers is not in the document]
Nikokolas3270: Ok, adding more ns in the filter.
Thanks for the explanation :)

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: osd-delete-backplane-remediation-rbacs
+  namespace: openshift-backplane
+spec:
+  failedJobsHistoryLimit: 5
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "*/10 * * * *"
```
bmeng, line 10: Is it a bit too frequent? If the expire time is 12h, I think we can run this clean up every several hours

Nikokolas3270: Hello, this is consistent with the period used between cleans for the cleanup of service accounts:
https://github.com/openshift/managed-cluster-config/blob/master/deploy/osd-delete-backplane-serviceaccounts/20-delete-backplane-serviceaccounts.CronJob.yaml#L10

That said, I agree, the job period could be 1 hour and still be acceptable

deploy/osd-delete-backplane-remediation-rbacs/00-delete-backplane-remediation-rbacs.namespace.yml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-backplane
```
typeid, line 4: **Suggestion:** should we call this `openshift-backplane-remediations`? For managed scripts, it's called `openshift-backplane-managed-scripts`

deploy/osd-delete-backplane-remediation-rbacs/15-delete-backplane-remediation-rbacs.SubjectPermission.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: osd-delete-backplane-remediation-rbacs
+  namespace: openshift-rbac-permissions
+spec:
+  clusterPermissions:
+    - osd-get-namespace
+    - osd-delete-backplane-remediation-cluster-roles-and-bindings
+  permissions:
+    - clusterRoleName: osd-delete-backplane-remediation-roles-and-bindings
+      namespacesAllowedRegex: "(^openshift-.*)"
```
typeid, line 12: We're adding this to clusterPermissions as well as to permissions, wouldn't we just get the `clusterPermissions` then?

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
```diff
@@ -0,0 +1,66 @@
...
+            - name: osd-delete-backplane-remediation-rbacs
+              imagePullPolicy: Always
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              args:
+                - /bin/bash
+                - -c
+                - |
+                  now=$(date +%s);
+                  expiryMins=720;
+                  for ns in $(oc get ns --no-headers | awk '{print $1}' | grep -e '^\(\(openshift\)\|\(kube\)-.*\)\|\(default\)\|\(dedicated-admin\)$');do
```
typeid, line 41: Why does the logic differ so much from managed-scripts? Do we really spawn roles/permissions accross namespaces, instead of spawning everything in a single namespace? I'm looking at https://github.com/openshift/managed-cluster-config/blob/d9006932da780c1fbf458e54863bfbe42b068e28/deploy/osd-delete-backplane-script-resources/20-delete-backplane-script-resources.CronJob.yaml for reference. 

@RaphaelBut can you take a look as well?

typeid:
/hold

Just holding to clarify a few last things before we merge.

deploy/osd-delete-backplane-remediation-rbacs/00-delete-backplane-remediation-rbacs.namespace.yml
Nikokolas3270: We could do that, even if the long term plan is to no longer run crontabs on the customer clusters for those cleanups.
As such, namespaces hosting those crontabs should, in my mind, be kept to a minimal.
Hence I think it is better to have only one namespace to host both crontabs.

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
typeid: Solved: this approach is more secure as it doesn't get all roles/bindings on the cluster, only for select namespaces.

deploy/osd-delete-backplane-remediation-rbacs/15-delete-backplane-remediation-rbacs.SubjectPermission.yaml
Nikokolas3270: Cluster permissions are targeting `clusterrole`s & `clusterrolebinding`s.
Those permissions are targeting `role`s & `rolebinding`s.
Said differently cluster role `osd-delete-backplane-remediation-roles-and-bindings` and cluster role `osd-delete-backplane-remediation-cluster-roles-and-bindings` do not target the same resources.
Also:
- `osd-delete-backplane-remediation-roles-and-bindings` is bounded to namespaces starting with `openshift-`
- While `osd-delete-backplane-remediation-cluster-roles-and-bindings` is bounded to the whole cluster.

We could maybe have only one cluster role; this would simplifies things a bit. But this would also means that the cleanup crontab will now have access to customer workload namespaces which is unnecessary and which may be seen as a security issue.

Anyway, I have to fix the `namespacesAllowedRegex` regexp as, following yesterday update, it is no longer in phase with the namespaces looked at by the crontab.

deploy/osd-delete-backplane-remediation-rbacs/00-delete-backplane-remediation-rbacs.namespace.yml
typeid: Ah I see, so we already have that namespace, we just had it in this syncset again. I think that's okay.

deploy/osd-delete-backplane-remediation-rbacs/20-delete-backplane-remediation-rbacs.CronJob.yaml
Nikokolas3270: Remediations may choose to spawn roles/permissions in specific namespaces:
https://gitlab.cee.redhat.com/service/backplane-api/-/blob/master/pkg/remediation/remediation.go#L318

For now however, they only use cluster roles applying to all namespaces, for instance:
https://github.com/openshift/configuration-anomaly-detection/blob/d90ed0bf09a77276cc154469ae62a873790de23b/pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml#L4

But the possibility is there to have remediations only require permissions on some specific namespaces.
Those namespaces have to be in this list:
- Namespaces starting with `openshift-`
- Namespaces starting with `kube-`
- The `default` namespace
- The `dedicated-admin` namespace

Managed scripts may also require permissions on some specific namespaces only... but there is (for now) no restrictions on what those namespaces can be.
Hence the cleanup script for the roles & bindings in those namespaces just do a full cluster scan (`-A` option) when looking for roles and bindings to clean.

Note that in both cases, it is possible for a remediation or a managed script to get permissions to look in or interfere with a customer namespace. It is just that for remediations, this has to be done through a cluster role.

PR #2371, opened 2025-03-06
OSD-26790: Adding UpgradeStateNotificationFailureSRE alert

TheUndeadKing:
What type of PR is this?
(bug)

What this PR does / why we need it?
TLDR: When the user gets banned (OR) PS has been rotated in a middle of the upgrade, the upgrade will stall. We're creating a metric if MUO unable to notify it's state to the OCM. This metric will help us to track via PD, so that we can unblock the upgrade.

In previous https://github.com/openshift/managed-upgrade-operator/pull/494, we'd implemented the logic to enable upgrade_notification_failed metric. In this PR, we're triggering UpgradeNotificationFailedSRE alert, if we cluster unable to notify it's state more than 30 mins

Which Jira/Github issue(s) this PR fixes?
Fixes # [OSD-26790](https://issues.redhat.com//browse/OSD-26790)


TheUndeadKing:
Previous [PR](https://github.com/openshift/managed-cluster-config/pull/2359) has been reverted because missing SOP. I'd created SOP for this alert

PR #2373, opened 2025-03-08
Remove control plane operator image pin annotation from all clusters

jharrington22:
### What type of PR is this?
bug

This PR will remove the control plane operator image pin annotation from all ManifestWork resources on service clusters within the fleet (Originally applied in [SDE-4686](https://issues.redhat.com//browse/SDE-4686)). 

The pinning of the operator version causes a bug where control plans can't upgrade because the control-plane-operator-image lays down a ConfigMap "kas-config" with invalid kube-apiserver arguments causing the kube-apiserver pods to crashloop. 

_Fixes # OCPBUGS-52819

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/ocpbugs-48519/04-ConfigMap.yaml
+++ b/deploy/ocpbugs-48519/04-ConfigMap.yaml
@@ -63,31 +63,19 @@ data:
     
     # Loop through each manifestwork object
     for clusterID in ${managedclusters[@]};do
-      # Extract full version, namespace and name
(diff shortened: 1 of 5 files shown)


joshbranham (commented):
Logic is simplified to just remove the annotation, agreed.

/lgtm

joshbranham (commented):
/lgtm

PR #2377, opened 2025-03-11
swap to job and fix some logic

jimdaga:
### What type of PR is this?
feature

- Swap from `CronJob` to `Job` so logic executes right away 
- Configure Job to delete hung pods after 5 minutes
- Configure Job to self delete after 10 minutes 
- Update SSS to purge created resources when we follow up with a delete PR ([settings](https://github.com/openshift/hive/blob/master/docs/syncset.md#syncset-object-definition))
- Fix a mistake in a grep to ensure we only get the resource name
- Add workaround for `Failed to watch *unstructured.Unstructured: unknown"` when cleaning CSVs

[OSD-28723](https://issues.redhat.com//browse/OSD-28723)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-route-monitor-operator-reinstall/04-openshift-route-monitor-operator.CronJob.yaml
+++ b/deploy/osd-route-monitor-operator-reinstall/04-openshift-route-monitor-operator.CronJob.yaml
@@ -1,30 +1,26 @@
 ---
 apiVersion: batch/v1
-kind: CronJob
+kind: Job
 metadata:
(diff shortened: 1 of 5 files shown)


joshbranham (commented):
I think this looks good, were you able to test this on stage?

jimdaga:
Yes

rafael-azevedo (approved):

PR #2400, opened 2025-04-04
Maint(OSD-29242): exclude openshift-cluster-observability-operator from pdb alerting

typeid:
### What type of PR is this?
feature

We don't care about PDBs in `openshift-cluster-observability-operator`, this is a customer installed operator via the hub. 
As a little background, cluster-observability-operator (COO) used to be observability-operator (OBO). It changed name recently as it moved into the operatorhub. We already had OBO in the list of namespaces we don't check PDBs on. 

_Fixes #_https://issues.redhat.com/browse/OSD-29242

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


typeid:
Had accidentally branched out from upstream, opened https://github.com/openshift/managed-cluster-config/pull/2401 to correct that

PR #2407, opened 2025-04-09
fix(OSD-25143): containerize python PR check calls

typeid:
### What type of PR is this?
bug

The `app_sre_pr_check.sh` calls to validate promql & other things was failing silently. This seems to be broken since the runners have been migrated, and no longer contain python.

This PR containerizes the python calls, and prevents future incidents like https://issues.redhat.com/browse/SDE-4312.

_Fixes #https://issues.redhat.com/browse/OSD-25143_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/hack/app_sre_pr_check.sh
+++ b/hack/app_sre_pr_check.sh
@@ -5,14 +5,17 @@ set -exv
 trap "rm -f sorted-before*.yaml.tmpl sorted-after*.yaml.tmpl" EXIT
 
 PROMTOOL_IMAGE=quay.io/prometheus/prometheus
-CONTAINER_ENGINE=docker
+
+# Image that contains python, yaml & co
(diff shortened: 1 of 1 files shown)


hack/app_sre_pr_check.sh
```diff
@@ -5,22 +5,25 @@ set -exv
 trap "rm -f sorted-before*.yaml.tmpl sorted-after*.yaml.tmpl" EXIT
 
 PROMTOOL_IMAGE=quay.io/prometheus/prometheus
-CONTAINER_ENGINE=docker
+
+# Image that contains python, yaml & co
+PYTHON_IMAGE=quay.io/redhat-services-prod/openshift/boilerplate:latest
```
typeid, line 10: Open for alternative suggestions for a smaller image. I think we're in a good place with boilerplate:latest though.

rafael-azevedo (approved):

PR #2408, opened 2025-04-10
Rescope capa

rafael-azevedo:
### What type of PR is this?
This rescopes the capa managed policy in order to remove the requirement of a request tag on ENI that is conditioned in the run-instances call.


fahlmant (approved):
/lgtm
/approve
Begrudingly

PR #2409, opened 2025-04-11
removing managed policy in versioned sts folders

rafael-azevedo:
### What type of PR is this?
Managed policies are not versioned. To remove the ambiguity as to which is the current version being worked on and most up to date this PR remove them from the STS/{verison} folder and moves them into the parent {STS} folder.


joshbranham (commented):
This is a welcome change from my side

/lgtm

PR #2410, opened 2025-04-11
Fix OCMAgentResponseFailureServiceLogsSRE alert

cblecker:
### What type of PR is this?
bug

The current alert will not resolve unless it's been successful for an hour. We don't need to use the avg over time when we are also setting the "for" stanza

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


joshbranham (commented):
Nice find

/lgtm

cblecker:
As this is a small bug in a production alert, I'm manually approving..

PR #2415, opened 2025-04-16
OSD-29046: Fixing SlowCustomerWebhookSRE, Not to trigger if service name missing

TheUndeadKing:
### What type of PR is this?
_(bug)_

**Problem statement:**
ocm-agent tries to send the SL for slowcustomerwebhookSRE alert & it gets failed because of missing service_name label. Also ocm-agent keeps on re-sending SL & triggered OCMAgentResponseFailureServiceLogsSRE.

The Service name is the one contain actual webhook causing the delay issue. in order to unblock OCMAgentResponseFailureServiceLogsSRE alert, Changing alert not to triggger if service name is missing

For more details refer: https://redhat-internal.slack.com/archives/C01JJKHC0CF/p1744795698368939?thread_ts=1744121845.696969&cid=C01JJKHC0CF

_Fixes #_
https://issues.redhat.com/browse/OSD-29046


typeid (approved):
/lgtm
/approve

PR #2420, opened 2025-04-29
Remove RHOBS_URL template variable that is no longer in use

dustman9000:
### What type of PR is this?
_cleanup_

Remove RHOBS_URL that is no longer in use.

_Fixes APPSRE-11851_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bmeng:
@dustman9000 Seems it has been 90 days, do we still need this?

PR #2433, opened 2025-05-27
SREP-400: Refactor CAPI annotator job

joshbranham:
### What type of PR is this?
feature

This is a refactor of the CAPI annotator job that we run on the fleet to ensure HCP clusters work with and without the ability to tag ENIs on `RunInstances` calls.

The following changes have been made, with inspiration from the work done in https://github.com/openshift/managed-cluster-config/pull/2353

* Only pin the image for clusters that are in the affected range.
* Unpin clusters once they upgrade out of the affected range.
* Consolidate the logic + scripts into a single ConfigMap. Changes to the scripts need to be copied to the `04-ConfigMap.yaml` by running `generate_configmap.sh`

Given we have north of 500 clusters in the affected Z stream ranges, we will need to run this job for some time. By adding the ability to remove the pin when not needed, we ensure new CAPI changes are applied to the fleet.

Fixes https://issues.redhat.com/browse/SREP-400

I have tested this on a service cluster in staging to confirm the logic works, where it patches affected clusters, removes the patch for clusters that don't need it, and handles MCs and SCs being skipped. See below for some logs.

**Note** As with the previous PKI operator job, we get a `The request is invalid` when the job runs subsequently after removing the annotation. The OCP API rejects the patch since the annotation doesn't exist. We could add an additional guard here to check for it first, but I see no harm in this.

```
> oc logs -f capa-annotator-4-16-29139550-rjfqw -n openshift-capa-annotator
don't patch: version 4.16.41 is >= x.y.23
cluster 2j1d7tra4n5oj1t2t0deerarn1k9oc6e does not need the override because it's Z stream is >= 23
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2j1d7tra4n5oj1t2t0deerarn1k9oc6e
oc patch manifestwork 2j1d7tra4n5oj1t2t0deerarn1k9oc6e -n hs-mc-kbj34s6bg --type='json' -p [{"op":"remove","path":"/spec/workload/manifests/0/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]
The request is invalid: the server rejected our request due to an error in our request
-------------------------------------------------------------------------
don't patch: version 4.16.40 is >= x.y.23
cluster hs-mc-jm0nah7n0 does not need the override because it's Z stream is >= 23
removing hypershift.openshift.io/capi-provider-aws-image annotation
skipping cluster hs-mc-jm0nah7n0 because it is a management or local cluster
```

```
> oc logs -f capa-annotator-29139550-thh2h -n openshift-capa-annotator
don't patch: version 4.17.4 is >= x.y.4
cluster 2f93ch1ksbg6kd3v3bmtsedbp0jsn1jm does not need the override because it's Z stream is >= 4
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2f93ch1ksbg6kd3v3bmtsedbp0jsn1jm
oc patch manifestwork 2f93ch1ksbg6kd3v3bmtsedbp0jsn1jm -n hs-mc-kbj34s6bg --type='json' -p [{"op":"remove","path":"/spec/workload/manifests/6/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]
The request is invalid: the server rejected our request due to an error in our request
-------------------------------------------------------------------------
don't patch: version 4.17.4 is >= x.y.4
cluster 2f9a6kv1nkjm1985helf1188vb8e9ej1 does not need the override because it's Z stream is >= 4
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2f9a6kv1nkjm1985helf1188vb8e9ej1
oc patch manifestwork 2f9a6kv1nkjm1985helf1188vb8e9ej1 -n hs-mc-kbj34s6bg --type='json' -p [{"op":"remove","path":"/spec/workload/manifests/6/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]
The request is invalid: the server rejected our request due to an error in our request
-------------------------------------------------------------------------
don't patch: version 4.17.4 is >= x.y.4
cluster 2f9b9f88vrtl0vpiubck9ha515b8q2l7 does not need the override because it's Z stream is >= 4
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2f9b9f88vrtl0vpiubck9ha515b8q2l7
oc patch manifestwork 2f9b9f88vrtl0vpiubck9ha515b8q2l7 -n hs-mc-kbj34s6bg --type='json' -p [{"op":"remove","path":"/spec/workload/manifests/6/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]
The request is invalid: the server rejected our request due to an error in our request
-------------------------------------------------------------------------
don't patch: version 4.17.4 is >= x.y.4
cluster 2fac28tjclnvobgrt8l34624g33eg1m1 does not need the override because it's Z stream is >= 4
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2fac28tjclnvobgrt8l34624g33eg1m1
oc patch manifestwork 2fac28tjclnvobgrt8l34624g33eg1m1 -n hs-mc-kbj34s6bg --type='json' -p [{"op":"remove","path":"/spec/workload/manifests/6/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]
The request is invalid: the server rejected our request due to an error in our request
-------------------------------------------------------------------------
```

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-25821-capa-annotator/04-ConfigMap.yaml
+++ b/deploy/osd-25821-capa-annotator/04-ConfigMap.yaml
@@ -74,6 +74,14 @@ data:
       fi
     
       namespace=$(oc get managedclusters "$clusterID" -o json | jq -r '.metadata.labels["api.openshift.com/management-cluster"]')
+
+      # If the managedcluster does not have a api.openshift.io/management-cluster label, then it should be skipped. This is
+      # usually because the local cluster + service cluster are registered as "managedcluster"s.
+      if [ "$namespace" = "null" ]; then
+        echo "skipping cluster $clusterID because it is a management or local cluster"
+        break
+      fi
+
       kinds=$(oc get manifestwork -n "$namespace" "$clusterID" -o json | jq -r '.spec.workload.manifests[].kind')
       num=0
       for kind in $kinds;do
--- a/deploy/osd-25821-capa-annotator/04-ConfigMap.yaml
+++ b/deploy/osd-25821-capa-annotator/04-ConfigMap.yaml
@@ -1,0 +1,96 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: capa-annotator
+  namespace: openshift-capa-annotator
+data:
+  should_patch.py: |
+    import sys
+    # should_patch.py takes two arguments: version and compare_version.
+    # It will then compare the Z stream of version to see if it's >= compare_version,
+    # exiting non-zero if it's not. Exiting non-zero means we need to patch.
+    #
+    # Example:
+    #   # This exits non-zero as the Z stream (10) is not >= 11
+    #   python should_patch.py 4.17.10 11
+    #
+    #   # This exits zero as the Z stream (10) is >= 9
+    #   python should_patch.py 4.17.10 9
+    
+    usage = """usage: python should_patch.py $version $compare_version
+    example: python should_patch.py 4.17.15 14 # exits 0 
+      """
+    
+    def should_patch(version, compare_version):
+        try:
+            x, y, z = version.split('.')
+            if int(z) >= int(compare_version):
+                print(f"don't patch: version {version} is >= x.y.{compare_version}")
+                return False
+            else:
+                print(f"should patch: version {version} is not >= x.y.{compare_version}")
+                return True
+        except ValueError:
+            print(f"could not compare version {version} to {compare_version}, defaulting to should patch")
+            return True
+    
+    def main():
+        if len(sys.argv) != 3:
+            print(usage)
+            sys.exit(1)
+        
+        version = sys.argv[1].rstrip()
+        compare_version = sys.argv[2].rstrip()
+        
+        if should_patch(version, compare_version):
+            sys.exit(1)
+    
+    if __name__=="__main__":
+        main()
+  patch.sh: |
+    #!/bin/bash
+
+    # Required Environment Variables:
+    # - IMAGE
+    # - MAJOR_MINOR_VER
+    # - Z_STREAM_FIXED_VER
+    
+    # Get all manifestwork objects and extract their names
+    managedclusters=$(oc get managedclusters -l openshiftVersion-major-minor=${MAJOR_MINOR_VER} -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
+    
+    # Loop through each manifestwork object
+    for clusterID in ${managedclusters[@]};do
+      # Extract full version, namespace and name
+      version=$(oc get managedclusters $clusterID -o json | jq -r '.metadata.labels.openshiftVersion')
+    
+      # Only patch the image if the clusters Z stream version is < $Z_STREAM_FIXED_VER
+      # This also handles any case where the script exits non-zero for any other reason, opting to patch as a result.
+      cleanup="false"
+      if python /tmp/scripts/should_patch.py "${version}" "${Z_STREAM_FIXED_VER}"; then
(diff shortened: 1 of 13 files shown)


deploy/osd-25821-capa-annotator/patch.sh
```diff
@@ -0,0 +1,53 @@
...
+  version=$(oc get managedclusters $clusterID -o json | jq -r '.metadata.labels.openshiftVersion')
+
+  # Only patch the image if the clusters Z stream version is < $Z_STREAM_FIXED_VER
+  # This also handles any case where the script exits non-zero for any other reason, opting to patch as a result.
+  cleanup="false"
+  if python /tmp/scripts/should_patch.py "${version}" "${Z_STREAM_FIXED_VER}"; then
+    echo "cluster ${clusterID} does not need the override because it's Z stream is >= ${Z_STREAM_FIXED_VER}"
+    echo "removing hypershift.openshift.io/capi-provider-aws-image annotation"
+    cleanup="true"
+  fi
```
joshbranham, line 23: New behavior, copied from what we did for PKI operator patching process

deploy/osd-25821-capa-annotator/patch.sh
```diff
@@ -0,0 +1,53 @@
...
+  num=0
+  for kind in $kinds;do
+    if [[ $kind == "HostedCluster" ]]; then
+        echo "patching cluster: $clusterID"
+
+        if [[ $cleanup = "true" ]]; then
+          json_payload='[{"op":"remove","path":"/spec/workload/manifests/'"$num"'/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image"}]'
+        else
+          json_payload='[{"op":"replace","path":"/spec/workload/manifests/'"$num"'/metadata/annotations/hypershift.openshift.io~1capi-provider-aws-image","value":"'"$IMAGE"'"}]'
+        fi
```
joshbranham, line 44: New behavior to either remove the annotation or set a value, based on previous `cleanup` var.

deploy/osd-25821-capa-annotator/should_patch.py
```diff
@@ -0,0 +1,43 @@
+import sys
```
bergmannf, line 1: I assume we are duplicating the shell-script and python-script, so it's:
1. Easier to review.
2. There are tests for the Python part.
If so - and the actual code that matters - is just the one in the ConfigMap, I'd like to see a comment explaining this.

bergmannf:
/lgtm

I think this looks fine - I like the fact that his adding tests & also having the files separately makes them easier to review.
I wonder if we should maybe think about:
1. Having a way to 'slurp' in files during build time, so we don't have to duplicate the scripts in the ConfigMap that end up being used.
2. Look into having an image that has `boto3` and [k8s-python](https://github.com/kubernetes-client/python) available, so we don't have to split into shell-scripts & python-code just because we have to use `oc`.

But obviously those are enhancements and nothing to do with this PR.

deploy/osd-25821-capa-annotator/should_patch.py
joshbranham: I added a script for generating the ConfigMap, as well as added some more comments and clarified in the README.

deploy/osd-25821-capa-annotator/04-ConfigMap.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: v1
```
joshbranham, line 1: This file is generated with `./generate_configmap.sh` from the bundled scripts.

PR #2448, opened 2025-06-12
disables boot image management in 4.18+ clusters

iamkirkbater:
### What type of PR is this?
Feature

Explicitly disables MCO Boot Image management so that we don't override existing AMIs.

_Fixes [OSD-30391](https://issues.redhat.com/browse/OSD-30391)_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
+++ b/hack/00-osd-managed-cluster-config-integration.yaml.tmpl
@@ -34931,6 +34931,79 @@ objects:
       managed.openshift.io/gitHash: ${IMAGE_TAG}
       managed.openshift.io/gitRepoName: ${REPO_NAME}
       managed.openshift.io/osd: 'true'
+    name: osd-machine-configuration
+  spec:
+    clusterDeploymentSelector:
+      matchLabels:
+        api.openshift.com/managed: 'true'
+      matchExpressions:
+      - key: hive.openshift.io/version-major-minor
+        operator: NotIn
+        values:
+        - '4.0'
+        - '4.1'
+        - '4.2'
+        - '4.3'
+        - '4.4'
+        - '4.5'
+        - '4.6'
+        - '4.7'
+        - '4.8'
+        - '4.9'
+        - '4.10'
+        - '4.11'
+        - '4.12'
+        - '4.13'
+        - '4.14'
+        - '4.15'
+        - '4.16'
+        - '4.17'
+    resourceApplyMode: Sync
+    patches:
+    - apiVersion: operator.openshift.io/v1
+      applyMode: AlwaysApply
+      kind: MachineConfiguration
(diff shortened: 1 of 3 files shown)


iamkirkbater:
/hold

Additional work to disable 4.17 and local testing still to be done.

deploy/osd-machine-configuration/config.yaml
```diff
@@ -0,0 +1,7 @@
+deploymentMode: "SelectorSyncSet" 
+selectorSyncSet: 
+  resourceApplyMode: "Sync"
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: NotIn
+    values: ["4.0", "4.1", "4.2", "4.3", "4.4", "4.5", "4.6", "4.7", "4.8", "4.9", "4.10", "4.11", "4.12", "4.13", "4.14", "4.15", "4.16", "4.17"]
```
jharrington22, line 7: We should do this for only 4.17 and 4.18

deploy/osd-machine-configuration/00-disable-managed-boot-images.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: MachineConfiguration
+metadata:
+  name: cluster
+  namespace: openshift-machine-config-operator
+spec:
+  managedBootImages:
+    machineManagers: []
```
jharrington22, line 8: Do we want to patch this instead to no stomp any future configurations here?

jharrington22 (changes requested):

deploy/osd-machine-configuration/config.yaml
jharrington22: Ideally we don't want to manage this in two places if its going to be set at install time, will it be set by the installer?

iamkirkbater: It would be set at install time, but it would not be managed further past that point. So if it somehow was changed manually at any point it would never be reverted. So this would prevent it from becoming managed after install time.

Ideally we would also couple this with a webhook that would then prevent anything from overwriting it at that point.

deploy/osd-machine-configuration/config.yaml
```diff
@@ -0,0 +1,7 @@
+deploymentMode: "SelectorSyncSet" 
+selectorSyncSet: 
+  resourceApplyMode: "Sync"
+  matchExpressions:
+  - key: hive.openshift.io/version-major-minor
+    operator: NotIn
```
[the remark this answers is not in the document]
iamkirkbater: I'm going to resolve this specific comment since the outcome will be the same based on the other ongoing conversation.

PR #2449, opened 2025-06-12
Disable updated boot images for 4.17 and 4.18

jharrington22:
### What type of PR is this?
feature

Disable the update boot image feature so that MCO doesn't replace the worker node AMIs we install clusters with. 

https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/machine_configuration/mco-update-boot-images#mco-update-boot-images-disable_machine-configs-configure

For 4.19+ we will apply this manifest at install time with [OCM-16382 ](https://issues.redhat.com/browse/OCM-16382 )

_Fixes # [OSD-30391](https://issues.redhat.com//browse/OSD-30391)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-machine-configuration-ami/config.yaml
+++ b/deploy/osd-machine-configuration-ami/config.yaml
@@ -1,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: hive.openshift.io/version-major-minor
+      operator: In
+      values: ["4.17", "4.18"]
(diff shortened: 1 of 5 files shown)


jharrington22:
Replaces #2448

joshbranham (commented):
Based on the docs, and the desire to only do this for 4.17 and 4.18, this looks good to me

/lgtm

deploy/osd-machine-configuration-ami/config.yaml
```diff
@@ -0,0 +1,6 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: hive.openshift.io/version-major-minor
+      operator: In
+      values: ["4.17", "4.18"]
```
rogbas, line 6: Per slack discussion, if we move to a "notIn" [4.0 to 4.16] rule, we don't need to revisit this adding every new Y version in the future.

rogbas (commented):
Overall it LGTM. The only suggestion has been added as a comment.

PR #2474, opened 2025-07-16
Adds back the configmap for HCP clusters.

iamkirkbater:
Additional debugging as to why the json patch is not working for removing this for HCP clusters.

### What type of PR is this?

bug

For some reason the patch isn't applying in prod clusters. Maybe because some of them are updating at the moment?

Additional debugging is needed.

[SREP-901](https://issues.redhat.com//browse/SREP-901)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


PeterCSRE:
lgtm

PR #2477, opened 2025-07-21
[SREP-198] Use new PKO metrics.

bergmannf:
The original metrics were missing cluster_id & other SRE relevant labels.

The new metrics are now merged with the 'sre:telemetry:managed_labels' to add our labels to the metrics.

These are added in the PKO deployment, as they already add labels:

See: https://gitlab.cee.redhat.com/lp-sre/package-operator/-/merge_requests/90

### What type of PR is this?
feature

Allows checking installs of PKO packages across clusters.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


bergmannf:
I've removed my auto-approve to ensure we only merge this when we are sure we don't need to promote MCC before next Monday.

PR #2486, opened 2025-07-30
ACM-22748 | fix: update rosa ingress policy pruneObjectBehavior

cdoan1:
set rosa-ingress-certificate-policies pruneObjectBehavior to None

### What type of PR is this?
bug

When we upgrade to ACM 2.13; the immediate problem we see is ROSA HCP fail to successfully delete.
The managedcluster CR stalls deleting because the klusterlet addon pods associated to the HCP cluster cannot be deleted.
They cannot be deleted because there is a race condition where ACM is trying to delete resources under management while ACM agent itself is being uninstalled.
This was not a problem at 2.10, 2.11, but at 2.13, a new feature was added to ACM policy that expose this problem more.

Once ingress certificate secrets are created; the main other activity on them is to renew their content.
We do not need ACM to manage the deletion of the resource.

So changing the prune behavior from DeleteIfCreated to None is appropriate.

pruneObjectBehavior: None is the default behavior for policy.

Setting this should have no negative impact to existing HCP, but should make HCP cleanup more reliable, removing ACM policy from a dependency.

[ACM-22748](https://issues.redhat.com//browse/ACM-22748)

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-rosa-ingress-certificate-policies.Policy.yaml
@@ -81,7 +81,7 @@ spec:
                           name: '{{hub (printf "%s-primary-cert-bundle-secret" .ManagedClusterName) hub}}'
                           namespace: openshift-ingress
                     {{hub- end hub}}
-                pruneObjectBehavior: DeleteIfCreated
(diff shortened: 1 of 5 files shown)


rafael-azevedo (approved):
/lgtm

PR #2488, opened 2025-08-07
Revert "Add rosa-legacy-brand-migration cronjob with priority class and resource requests"

iamkirkbater:
Reverts openshift/managed-cluster-config#2484

I found that we reverted the previous PR and then Amarthya created this one with the appropriate controls to satisfy conformance testing, however we never promoted after the revert. So the original CronJob was still present across the fleet and has been running for the last two weeks successfully. With that - we no longer need this Cronjob to run as I've checked the fleet and ensured that as many clusters as possible are patched, with notifications going out to the remaining customers.

TL;DR - This cronjob is no longer needed.


iamkirkbater:
/override ci/prow/checklinks-pr

Checklinks is broken due to #2490. Shouldn't hold up existing PRs for that.

PR #2495, opened 2025-08-18
SREP-1000: Add supplemental node-exporter to collect ethtool AWS packet drop metrics

nephomaniac:
### What type of PR is this?
Feature. Deploy a supplemental node-exporter daemonset to management-clusters in order to monitor dropped packets per [AWS ENA allowance exceeded metrics](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-network-performance-ena.html#network-performance-metrics). 
This is intended to be a temporary measure to help address [SREP-1000](https://issues.redhat.com/browse/SREP-1000) until CMO supports the node-exporter ethtool collector per [RFE-7342](https://issues.redhat.com/browse/RFE-7342).

(See [SREP-1000](https://issues.redhat.com/browse/SREP-1000) for additional info). This helps SRE identify the root cause of specific packet drops events due to AWS bandwidth limits.  

https://issues.redhat.com/browse/SREP-1000

This has been validated with manual deployments. Per normal release cycle, this will need to be validated in int, stage before promoted to prod.  
When current MR is acceptable, this should be tested on an MC in staging prior to merging. 
This should including verifying :
- this installs/applies to cluster w/o error 
- works as expected. Targets are healthy, metrics are query-able in prom and only the exact metrics expected are exported and scraped
- does not interfere with CMO or any other processes
- pod/node resource consumption is reasonable
- important to check on uninstall/delete does not have a negative impact CMO.

I've been testing on a 4.16.42 cluster, and will definitely test on a staging cluster once we are ok with the current draft, and before merging.

- [ X] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
+++ b/deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
@@ -3,13 +3,13 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/description: "Provide supplemental metrics not currently available in Openshift CMO"
-    openshift.io/display-name: "openshift-sre-observe"
+    openshift.io/display-name: "openshift-sre-ethtool-exporter"
     security.openshift.io/MinimallySufficientPodSecurityStandard: privileged
   labels:
-    kubernetes.io/metadata.name: openshift-sre-observe
+    kubernetes.io/metadata.name: openshift-sre-ethtool-exporter
     openshift.io/cluster-monitoring: "true"
     pod-security.kubernetes.io/audit: privileged
     pod-security.kubernetes.io/audit-version: latest
     pod-security.kubernetes.io/warn: privileged
     pod-security.kubernetes.io/warn-version: latest
-  name: openshift-sre-observe
+  name: openshift-sre-ethtool-exporter
--- a/deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
+++ b/deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/description: "Provide supplemental metrics not currently available in Openshift CMO"
+    openshift.io/display-name: "openshift-sre-observe"
+    security.openshift.io/MinimallySufficientPodSecurityStandard: privileged
+  labels:
+    kubernetes.io/metadata.name: openshift-sre-observe
+    openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/audit-version: latest
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/warn-version: latest
+  name: openshift-sre-observe
--- a/deploy/srep-1000-ethtool-exporter/08-Secret.yaml
+++ b/deploy/srep-1000-ethtool-exporter/08-Secret.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+data:
+  config.yaml: ImF1dGhvcml6YXRpb24iOgogICJzdGF0aWMiOgogIC0gInBhdGgiOiAiL21ldHJpY3MiCiAgICAicmVzb3VyY2VSZXF1ZXN0IjogZmFsc2UKICAgICJ1c2VyIjoKICAgICAgIm5hbWUiOiAic3lzdGVtOnNlcnZpY2VhY2NvdW50Om9wZW5zaGlmdC1tb25pdG9yaW5nOnByb21ldGhldXMtazhzIgogICAgInZlcmIiOiAiZ2V0Ig==
+metadata:
+  labels:
+  name: ethtool-exporter-kube-rbac-proxy-config
+  namespace: openshift-sre-observe
+type: Opaque
--- a/deploy/srep-1000-ethtool-exporter/08-Secret.yaml
+++ b/deploy/srep-1000-ethtool-exporter/08-Secret.yaml
@@ -5,5 +5,5 @@ data:
 metadata:
   labels:
   name: ethtool-exporter-kube-rbac-proxy-config
-  namespace: openshift-sre-observe
+  namespace: openshift-sre-ethtool-exporter
 type: Opaque
--- a/deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
+++ b/deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
@@ -85,13 +85,13 @@ spec:
             - name: sysfs
               mountPath: /host/sys
               readOnly: true
-            - mountPath: /var/node_exporter/text/file
-              name: node-exporter-textfile
+            - mountPath: /var/ethtool_exporter/text/file
+              name: ethtool-exporter-textfile
               readOnly: true
             - name: procfs
               mountPath: /host/proc
               readOnly: true
-          workingDir: /var/node_exporter/textfile
+          workingDir: /var/ethtool_exporter/textfile
         - args:
             - --secure-listen-address=[$(IP)]:9500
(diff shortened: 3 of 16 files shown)


deploy/srep-1000-ethtool-exporter/08-Secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+data:
+  config.yaml: ImF1dGhvcml6YXRpb24iOgogICJzdGF0aWMiOgogIC0gInBhdGgiOiAiL21ldHJpY3MiCiAgICAicmVzb3VyY2VSZXF1ZXN0IjogZmFsc2UKICAgICJ1c2VyIjoKICAgICAgIm5hbWUiOiAic3lzdGVtOnNlcnZpY2VhY2NvdW50Om9wZW5zaGlmdC1tb25pdG9yaW5nOnByb21ldGhldXMtazhzIgogICAgInZlcmIiOiAiZ2V0Ig==
```
joshbranham, line 4: Does this have to be a secret?

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
```diff
@@ -0,0 +1,190 @@
...
+            - --tls-min-version=VersionTLS12
+            - --tls-private-key-file=/etc/tls/private/tls.key
+            - --client-ca-file=/etc/tls/client/ca-bundle.crt
+          env:
+            - name: IP
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: status.podIP
+          image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:971174692a1aae68ac07ffa132254eb3c018a52507eb32621b6dc286fd25ca0d
```
joshbranham, line 102: What version is this?

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
```diff
@@ -0,0 +1,190 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
joshbranham, line 2: I presume most of this is copied from the existing node-exporter DS in an OCP cluster? Might be good to (if the list is short) add a comment to the file with what you have modified/changed.

deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
```diff
@@ -0,0 +1,15 @@
...
+    openshift.io/display-name: "openshift-sre-observe"
+    security.openshift.io/MinimallySufficientPodSecurityStandard: privileged
+  labels:
+    kubernetes.io/metadata.name: openshift-sre-observe
+    openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/audit-version: latest
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/warn-version: latest
+  name: openshift-sre-observe
```
joshbranham, line 15: Maybe call this something specific like `openshift-sre-ethtool-exporter`?

deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
joshbranham, line 1: `nit`: bunch of files missing a new line

deploy/srep-1000-ethtool-exporter/08-Secret.yaml
nephomaniac: Likely no, but since this is for the RBAC perms it might be preferred? This is borrowed directly from CMO in order to provide access specifically to the openshift-monitoring/prometheus-k8s serviceAccount/token.

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
nephomaniac: Thanks for pointing this out Josh. The recent commit has been updated to add comments/info around each image + version. The images are taken from CMO on 4.16.42 MCs.  This would be better involving a trigger/stream, but not sure what makes sense for these images specifically?

deploy/srep-1000-ethtool-exporter/00-Namespace.yaml
nephomaniac: Good suggestion. The previous naming was just a placeholder. Now using 'openshift-sre-ethtool-exporter'.

deploy/srep-1000-ethtool-exporter/08-Secret.yaml
joshbranham: Yeah I decrypted the value and it's just yaml config with no secret in it. But probably fine if that is what is expected

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
```diff
@@ -0,0 +1,199 @@
...
+            - --tls-private-key-file=/etc/tls/private/tls.key
+            - --client-ca-file=/etc/tls/client/ca-bundle.crt
+          env:
+            - name: IP
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: status.podIP
+          # Taken from PROD 4.16.42 MCs currently using: sha256:82f5ad5a8c1827db890f6bbb9f85e3f571eceba5bc6903b721be6fdeaea651fa
+          image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:82f5ad5a8c1827db890f6bbb9f85e3f571eceba5bc6903b721be6fdeaea651fa
```
joshbranham, line 111: I am not sure how much this version matters, but I presume it will get out of date pretty quickly. If the reality is this is all temporary, then maybe that is okay.

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
```diff
@@ -0,0 +1,199 @@
...
+            - name: sysfs
+              mountPath: /host/sys
+              readOnly: true
+            - mountPath: /var/node_exporter/text/file
+              name: node-exporter-textfile
+              readOnly: true
+            - name: procfs
+              mountPath: /host/proc
+              readOnly: true
+          workingDir: /var/node_exporter/textfile
```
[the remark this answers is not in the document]
nephomaniac: Thanks, good catch. This should be specific to this exporter. Updated in recent commit. 

For test purposes, we need to test that this installs, works correctly, has the exact metrics expected, does not interfere with any other processes, and then important to check on uninstall/delete does not have a negative impact CMO. 
I've been testing on a 4.16.42 cluster, and will definitely test on a staging cluster once we are ok with the current draft, and before merging.

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
nephomaniac: Agree.  Depending on the timeline vs complexity, maybe can create some follow up tasks/ticket(s) to check and address this w/o automations for the near term?

nephomaniac:
I've updated SREP-1000 to note the staging management clusters this is currently installed on, as well as attached screenshots showing the new metrics.  New non-zero *_allowance_exceeded metrics can be seen on each MC. 

I've tested installing->uninstalling->installing and have not seen issues. CMO seems to function fine during and after the installation tests. No noticeable spikes in CPU/Mem on CMO prom pods.

deploy/srep-1000-ethtool-exporter/config.yaml
```diff
@@ -0,0 +1,14 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+    - key: ext-hypershift.openshift.io/cluster-type
+      operator: In
+      values: ["management-cluster"]
+    - key: api.openshift.com/fedramp
+      operator: NotIn
+      values: ["true"]
+    - key: api.openshift.com/environment
```
dustman9000, line 10: It's actually "stage" and "integration". Just use NotIn to keep it simple.
```suggestion
    - key: api.openshift.com/environment
    operator: NotIn
    values:
      - "production"
```

deploy/srep-1000-ethtool-exporter/09-DaemonSet.yaml
dustman9000: Can we use quay.io/prometheus/node-exporter:latest instead of the ocp image?

PR #2540, opened 2025-09-30
[SREPHOA-42] Add osd-master-node-labeller cronjob

bmeng:
This adds a CronJob that automatically labels master nodes with the node-role.kubernetes.io/control-plane label for backward compatibility.

The job runs daily at midnight and targets OSD/ROSA clusters running OpenShift 4.14-4.18. It exits early if all master nodes already have the control-plane label.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### What type of PR is this?

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
+++ b/deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
@@ -1,0 +1,78 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: osd-master-node-labeller
+  namespace: openshift-config
+spec:
+  failedJobsHistoryLimit: 3
+  successfulJobsHistoryLimit: 3
+  concurrencyPolicy: Replace
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+                - preference:
+                    matchExpressions:
+                      - key: node-role.kubernetes.io/infra
+                        operator: Exists
+                  weight: 1
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/infra
+              operator: Exists
+          serviceAccountName: osd-master-node-labeller
+          restartPolicy: Never
+          containers:
+            - name: osd-master-node-labeller
+              image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+              imagePullPolicy: Always
+              command:
+                - /bin/bash
+                - -c
+                - |
+                  #!/bin/bash
+                  set -euo pipefail
+
+                  echo "Starting master node labeller job..."
+
+                  # Get all nodes with node-role.kubernetes.io/master label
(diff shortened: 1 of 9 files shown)


deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
```diff
@@ -0,0 +1,81 @@
...
+                    echo "No nodes found with label node-role.kubernetes.io/master"
+                    exit 0
+                  fi
+
+                  echo "Found master nodes: $master_nodes"
+
+                  # Check if all master nodes already have control-plane label
+                  nodes_needing_label=0
+                  for node in $master_nodes; do
+                    if ! oc get node "$node" --show-labels | grep -q "node-role.kubernetes.io/control-plane"; then
```
xiaoyu74, line 55: why the script loop the node twice to check labels (here in `line 55` and `line 72`)

deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
```diff
@@ -0,0 +1,81 @@
...
+                  # Label each master node with control-plane label
+                  for node in $master_nodes; do
+                    echo "Processing node: $node"
+
+                    # Check if the node already has the control-plane label
+                    if oc get node "$node" --show-labels | grep -q "node-role.kubernetes.io/control-plane"; then
+                      echo "Node $node already has node-role.kubernetes.io/control-plane label, skipping"
+                    else
+                      echo "Adding node-role.kubernetes.io/control-plane label to node $node"
+                      oc label node "$node" node-role.kubernetes.io/control-plane="" --overwrite
```
xiaoyu74, line 76: should we add error handling for the labeling?  currently, if labeling fails, the script  continues silently.

deploy/osd-master-node-labeller/config.yaml
```diff
@@ -0,0 +1,10 @@
+deploymentMode: "SelectorSyncSet"
+selectorSyncSet:
+  matchExpressions:
+  - key: api.openshift.com/product
+    operator: In
+    values: ["osd", "rosa"]
+  - key: hive.openshift.io/version-major-minor-patch
+    operator: In
+    values: ["4.14", "4.15", "4.16", "4.17", "4.18"]
```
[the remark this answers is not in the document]
bmeng: That's a good point, I was thinking the upgrade would happen on 4.18 to 4.19, so the old versions are unnecessary. But considering the job would be removed shortly after the changes are made. It would be good to include the old affected versions as well.

deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
bmeng: I do not feel adding error handling here would be quite valuable. It will still continue silently.
We have only warning alert to catch the job failures. And the cronjob will retry periodically anyway.

deploy/osd-master-node-labeller/10-osd-master-node-labeller.CronJob.yaml
bmeng: yes, we can count the node with label directly without looping

PR #2547, opened 2025-10-03
Scope Elasticsearch low-disk alert to only page SRE if disk capacity < 1200Gi

abyrne55:
### What type of PR is this?
_feature_

This PR lowers the `ElasticsearchDiskSpaceRunningLowSRE` alert's severity to warning and creates a new critical alert, `ElasticsearchDiskNeedsResizingSRE` that fires only if both free space is low and the disk space is less than the threshold at which the alert becomes a no-op for SRE (400Gi per disk, or 1200Gi total, per the [SOP](https://github.com/openshift/ops-sop/blob/master/v4/howto/logging.md#can-i-grow-es-storage)) 

Finishes the job started by [OSD-18456](https://issues.redhat.com/browse/OSD-18456)

### DIFF
--- a/deploy/sre-prometheus/extended-logging/101-parsed_elasticsearch_openshift-logging_elasticsearch-prometheus-rules.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/extended-logging/101-parsed_elasticsearch_openshift-logging_elasticsearch-prometheus-rules.PrometheusRule.yaml
@@ -152,7 +152,19 @@ spec:
           for: 1h
           labels:
             namespace: openshift-logging
-            severity: critical
+            severity: warning
+        - alert: ElasticsearchDiskNeedsResizingSRE
+          annotations:
+            message: Cluster {{ $labels.cluster }} is predicted to be out of disk space within the next 6h and should be resized to 1200Gi.
+            summary: Cluster low on disk space
+          expr: |
+            sum(predict_linear(es_fs_path_available_bytes[6h], 6 * 3600)) < 0
+              and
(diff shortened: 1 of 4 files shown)


deploy/sre-prometheus/extended-logging/101-parsed_elasticsearch_openshift-logging_elasticsearch-prometheus-rules.PrometheusRule.yaml
```diff
@@ -152,7 +152,19 @@ spec:
           for: 1h
           labels:
             namespace: openshift-logging
-            severity: critical
+            severity: warning
+        - alert: ElasticsearchDiskNeedsResizingSRE
+          annotations:
+            message: Cluster {{ $labels.cluster }} is predicted to be out of disk space within the next 6h and each of its disks should be resized to 400Gi.
```
joshbranham, line 158: Is there a reason we need to include specific sizing guidance? Iirc, the SOP has some options for this?

abyrne55: [Indeed](https://github.com/openshift/managed-cluster-config/pull/2547). From the SOP:
```
if currentStorage() < "1200Gi":
    notifyCustomer(action="Grow", target="1200Gi")
    growElasticStorageTo("1200Gi")
else:
    notifyCustomer(action="Reduce Logging")
    snoozeAlert("7 days")
```
Since the resize is always going to be to 1200Gi (spread across 3 400Gi disks), I figured why not specify it.

PR #2550, opened 2025-10-08
SREP-1000: Deploy MC ethtool exporter to all non-fedramp sectors

nephomaniac:
### What type of PR is this?
feature

Deploys supplemental ethtool node-exporter to Management Clusters in all non-fedramp sectors.  
This is currently deployed on 'most' int/staging management clusters, as well eu-west-2 in prod. 
Removes previously applied filters allowing this to be deployed in all int, stage and prod sectors. 

https://issues.redhat.com/browse/SREP-1000
(gather metrics for packet drops related to AWS ENA thresholds)

- [ x] Tested latest changes against a cluster
- [ x] Included documentation changes with PR
- [ x] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with:

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

### DIFF
--- a/deploy/srep-1000-ethtool-exporter/config.yaml
+++ b/deploy/srep-1000-ethtool-exporter/config.yaml
@@ -7,16 +7,3 @@ selectorSyncSet:
     - key: api.openshift.com/fedramp
       operator: NotIn
       values: ["true"]
-    - key: ext-hypershift.openshift.io/cluster-sector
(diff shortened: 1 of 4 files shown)


joshbranham (commented):
/lgtm

PR #2670, opened 2026-03-16
SREP-4038, STOR-2766: Update hcp aws-ebs-csi-driver credentials

dfajmon:
### What type of PR is this?
_feature_

This PR updates the IAM permission policy for the aws-ebs-csi-driver to align with upstream changes required for the OpenShift rebase. These updates ensure full driver functionality and support for enhanced EBS features.

It might seem like a lot of changes but a lot of them are just moved in the file so we are mroe aligned with upstream version https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/release-1.61/docs/AmazonEBSCSIDriverPolicyV2.json and following the AWS example of policies https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicyV2.html

## Summary by CodeRabbit

* **New Features**
  * Expanded the AWS IAM policy for the EBS CSI driver with additional EC2/EBS permissions, including availability zone and volume status reads, faster snapshot restores, and snapshot locking/deletion.
  * Added tag-scoped support for creating, copying, and snapshotting managed EBS volumes, with attachment/detachment allowed to any instance while other operations remain constrained by managed resource tags.
  * Updated KMS permissions for encrypted volume/snapshot workflows, constrained by the expected key tags.
* **Chores**
  * Tightened and refined tagging permissions for managed resources, restricting tag keys and ensuring tag operations apply correctly to copy/create actions.

### DIFF
--- a/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
+++ b/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
@@ -4,11 +4,16 @@
     {
       "Effect": "Allow",
       "Action": [
+        "ec2:CopyVolumes",
+        "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
         "ec2:DescribeSnapshots",
         "ec2:DescribeTags",
         "ec2:DescribeVolumes",
-        "ec2:DescribeVolumesModifications"
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus",
+        "ec2:EnableFastSnapshotRestores",
+        "ec2:LockSnapshot"
       ],
       "Resource": "*"
     },
--- a/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
+++ b/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
@@ -169,6 +169,39 @@
           ]
         }
       }
+    },
+    {
+      "Sid": "KMSEncryptedVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "kms:Decrypt",
+        "kms:GenerateDataKeyWithoutPlaintext"
+      ],
+      "Resource": "arn:aws:kms:*:*:key/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat": "true"
+        },
+        "StringLike": {
+          "kms:ViaService": "ec2.*.amazonaws.com"
+        }
+      }
+    },
+    {
+      "Sid": "KMSCreateGrant",
+      "Effect": "Allow",
+      "Action": [
+        "kms:CreateGrant"
+      ],
+      "Resource": "arn:aws:kms:*:*:key/*",
+      "Condition": {
+        "Bool": {
+          "kms:GrantIsForAWSResource": true
+        },
+        "StringLike": {
+          "kms:ViaService": "ec2.*.amazonaws.com"
+        }
+      }
     }
   ]
 }
--- a/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
+++ b/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
@@ -2,26 +2,110 @@
   "Version": "2012-10-17",
   "Statement": [
     {
+      "Sid": "ReadOnlyDescribeOperations",
       "Effect": "Allow",
       "Action": [
+        "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
         "ec2:DescribeSnapshots",
-        "ec2:DescribeTags",
         "ec2:DescribeVolumes",
-        "ec2:DescribeVolumesModifications"
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus"
       ],
       "Resource": "*"
     },
     {
+      "Sid": "CreateAndCopyVolumesWithManagedTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume",
+        "ec2:CopyVolumes"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:RequestTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "CopyManagedVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CopyVolumes"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/vol-*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "CreateSnapshotsWithManagedTag",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
+      ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:RequestTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "CreateSnapshotsFromManagedVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateSnapshot"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "ManageManagedVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:ModifyVolume",
+        "ec2:AttachVolume",
+        "ec2:DetachVolume",
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "CreateVolumesFromAndEnableFSROnManagedSnapshots",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume",
+        "ec2:EnableFastSnapshotRestores"
+      ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        }
+      }
+    },
+    {
+      "Sid": "AttachDetachVolumesToAnyInstance",
       "Effect": "Allow",
       "Action": [
         "ec2:AttachVolume",
         "ec2:DetachVolume"
       ],
-      "Resource": [
-        "arn:aws:ec2:*:*:instance/*",
-        "arn:aws:ec2:*:*:volume/*"
-      ],
+      "Resource": "arn:aws:ec2:*:*:instance/*",
       "Condition": {
         "StringEquals": {
           "aws:ResourceTag/red-hat-managed": "true"
@@ -29,14 +113,13 @@
       }
     },
     {
+      "Sid": "DeleteAndLockManagedSnapshots",
       "Effect": "Allow",
       "Action": [
-        "ec2:DeleteVolume",
-        "ec2:ModifyVolume"
-       ],
-      "Resource": [
-        "arn:aws:ec2:*:*:volume/*"
-       ],
+        "ec2:DeleteSnapshot",
+        "ec2:LockSnapshot"
+      ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
       "Condition": {
         "StringEquals": {
           "aws:ResourceTag/red-hat-managed": "true"
@@ -44,74 +127,7 @@
       }
     },
     {
-      "Effect": "Allow",
-      "Action": [
-        "ec2:CreateVolume"
-      ],
-      "Resource": [
-        "arn:aws:ec2:*:*:volume/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "aws:RequestTag/red-hat-managed": "true"
-        }
-      }
-    },
-    {
-      "Sid": "CreateVolumeFromSnapshot",
-      "Effect": "Allow",
-      "Action": [
-        "ec2:CreateVolume"
-      ],
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
-      ]
-    },
-    {
-      "Sid": "CreateSnapshotResourceTag",
-      "Effect": "Allow",
-      "Action": [
-        "ec2:CreateSnapshot"
-      ],
-      "Resource": [
-        "arn:aws:ec2:*:*:volume/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "aws:ResourceTag/red-hat-managed": "true"
-        }
-      }
-    },
-    {
-      "Sid": "CreateSnapshotRequestTag",
-      "Effect": "Allow",
-      "Action": [
-        "ec2:CreateSnapshot"
-      ],
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "aws:RequestTag/red-hat-managed": "true"
-        }
-      }
-    },
-    {
-      "Effect": "Allow",
-      "Action": [
-        "ec2:DeleteSnapshot"
-      ],
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
-      ],
-      "Condition": {
-        "StringEquals": {
-          "aws:ResourceTag/red-hat-managed": "true"
-        }
-      }
-    },
-    {
+      "Sid": "TagResourcesOnCreation",
       "Effect": "Allow",
       "Action": [
         "ec2:CreateTags"
@@ -124,7 +140,32 @@
         "StringEquals": {
           "ec2:CreateAction": [
             "CreateVolume",
-            "CreateSnapshot"
+            "CreateSnapshot",
+            "CopyVolumes"
+          ]
+        }
+      }
+    },
+    {
+      "Sid": "ModifyTagsOnManagedVolumes",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateTags",
+        "ec2:DeleteTags"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        },
+        "Null": {
+          "aws:TagKeys": "false"
+        },
+        "ForAllValues:StringNotEquals": {
+          "aws:TagKeys": [
+            "red-hat-managed",
+            "ebs.csi.aws.com/cluster",
+            "kubernetes.io/created-for/pvc/name"
           ]
         }
       }


resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -4,11 +4,16 @@
...
+        "ec2:DescribeAvailabilityZones",
         "ec2:DescribeInstances",
         "ec2:DescribeSnapshots",
         "ec2:DescribeTags",
         "ec2:DescribeVolumes",
-        "ec2:DescribeVolumesModifications"
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus",
+        "ec2:EnableFastSnapshotRestores",
+        "ec2:LockSnapshot"
```
feichashao: We need to restrict the resources. For example, we need to restrict them with tag
```
 "aws:ResourceTag/red-hat-managed": "true"
```

dfajmon: Updated the permissions to be more aligned with upstream https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/release-1.61/docs/example-iam-policy.json

Do we need the restrictions on all resources or just some?

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
         "ec2:DescribeInstances",
+        "ec2:DescribeInstanceTypes",
         "ec2:DescribeSnapshots",
-        "ec2:DescribeTags",
         "ec2:DescribeVolumes",
-        "ec2:DescribeVolumesModifications"
+        "ec2:DescribeVolumesModifications",
+        "ec2:DescribeVolumeStatus"
       ],
       "Resource": "*"
```
feichashao, line 16: Note for myself.

I checked that those Describe* actions are List operations. The "Resource" must be "*".  I have no concern on this.

https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html

I will need to double check whether those actions are needed. Maybe check the upstream code.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
+        "ec2:CopyVolumes"
       ],
+      "Resource": "arn:aws:ec2:*:*:volume/*",
       "Condition": {
         "StringEquals": {
-          "aws:ResourceTag/red-hat-managed": "true"
+          "aws:RequestTag/red-hat-managed": "true"
         }
       }
     },
```
feichashao, line 31: Review notes to myself.

Looks good. They can both scope to a volume resource and supports RequestTag.
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html

The `ec2:CopyVolumes` is newly added, need to double check why we need it.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
-        "arn:aws:ec2:*:*:volume/*"
-       ],
+        "ec2:CopyVolumes"
+      ],
+      "Resource": "arn:aws:ec2:*:*:volume/vol-*",
       "Condition": {
         "StringEquals": {
           "aws:ResourceTag/red-hat-managed": "true"
         }
       }
```
feichashao, line 43: I see a `ec2:CopyVolumes` action in the `CreateAndCopyVolumesWithManagedTag` section already.  This one seems to give the condition `ResourceTag`, which is in addition to the `CreateAndCopyVolumesWithManagedTag`.

The resource has `vol-*` which is inconsistent with `CreateAndCopyVolumesWithManagedTag`.


@dfajmon could you help me understand more about this section? (I don't mean this policy is not suitable, but want to understand more about this pattern.)

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
       "Action": [
-        "ec2:CreateVolume"
+        "ec2:CopyVolumes"
       ],
-      "Resource": [
-        "arn:aws:ec2:*:*:volume/*"
+      "Resource": "arn:aws:ec2:*:*:volume/vol-*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name": "*"
```
feichashao: I assume this is to copy the PVC volume which doesn't have the red-hat-managed tag.  Are we suppose to do so?

@dfajmon do you have more context about this part?

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
-      ]
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        }
+      }
     },
```
feichashao, line 70: Review notes for myself.

This part looks good.  It limits a RequestTag to snapshot and a ResourceTag to volume.
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html

The original policy already have this. No concern.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
       ],
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
+      "Resource": "arn:aws:ec2:*:*:volume/*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name": "*"
+        }
+      }
+    },
```
feichashao: This is on the resources with tag `kubernetes.io/created-for/pvc/name`, which may not be `red-hat-managed`.

@dfajmon do you have more context about why we need to allow this?

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
+        "ec2:EnableFastSnapshotRestores"
       ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
       "Condition": {
         "StringEquals": {
-          "aws:RequestTag/red-hat-managed": "true"
+          "aws:ResourceTag/red-hat-managed": "true"
         }
       }
     },
```
feichashao, line 100: The permission looks good.

`EnableFastSnapshotRestores` is a new permission, need to understand why we need it.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
       "Effect": "Allow",
       "Action": [
-        "ec2:DeleteSnapshot"
+        "ec2:AttachVolume",
+        "ec2:DetachVolume"
       ],
-      "Resource": [
-        "arn:aws:ec2:*:*:snapshot/*"
+      "Resource": "arn:aws:ec2:*:*:instance/*"
+    },
```
feichashao, line 114: Looks good. Existing permission in the original policy, no concern.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -2,116 +2,156 @@
...
+        "ec2:DeleteSnapshot",
+        "ec2:LockSnapshot"
       ],
+      "Resource": "arn:aws:ec2:*:*:snapshot/*",
       "Condition": {
         "StringEquals": {
           "aws:ResourceTag/red-hat-managed": "true"
         }
       }
     },
```
feichashao, line 128: `DeleteSnapshot` was in the original policy, no concern.

`LockSnapshot` is a new one, need to understand why we need it. The condition looks good though.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -124,10 +164,35 @@
         "StringEquals": {
           "ec2:CreateAction": [
             "CreateVolume",
-            "CreateSnapshot"
+            "CreateSnapshot",
+            "CopyVolumes"
```
feichashao, line 144: dependent action, looks good.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -124,10 +164,35 @@
...
+        "Null": {
+          "aws:TagKeys": "false"
+        },
+        "ForAllValues:StringNotEquals": {
+          "aws:TagKeys": [
+            "red-hat-managed",
+            "ebs.csi.aws.com/cluster-name",
+            "kubernetes.io/created-for/pvc/name"
           ]
         }
```
feichashao, line 170: This makes sense, we have a similar policy like this https://github.com/openshift/managed-cluster-config/pull/2720/changes

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: I tried to copy them as close as possible to the upstream so we don't miss anything from
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/release-1.61/docs/AmazonEBSCSIDriverPolicyV2.json?rgh-link-date=2026-03-16T15%3A27%3A18Z
- https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicyV2.html

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: As mentioned above mostly copied from upstream policies, this would be useful if there are resources that don't have red-hat-managed labels

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: As mentioned above mostly copied from upstream policies.
> Locks an Amazon EBS snapshot in either governance or compliance mode to protect it against accidental or malicious deletions for a specific duration. A locked snapshot can't be deleted.

[From AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LockSnapshot.html)

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: As mentioned above mostly copied from upstream policies, this would be useful if there are resources that don't have red-hat-managed labels. Not sure if that could happen though.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: As mentioned above mostly copied from upstream policies.
> Enables fast snapshot restores for the specified snapshots in the specified Availability Zones.
> You get the full benefit of fast snapshot restores after they enter the enabled state.

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EnableFastSnapshotRestores.html

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: As mentioned above mostly copied from upstream policies.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: I have linked Amazon example policies and upstream (they are fairly similar) in the PR description. Feel free to check. We aim to have it as close to upstream as possible while maintaining compatibility. Might be some conditions are not necessarily needed for Rosa. Let me know which one of those is.

dfajmon:
Added permissions to create and manage KMS-encrypted volumes ([source](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md)). Might need red-hat-managed? 

~This is a request coming from https://redhat.atlassian.net/browse/ROSA-206~

feichashao:
> Added permissions to create and manage KMS-encrypted volumes ([source](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md)). Might need red-hat-managed?

I believe the KMS key needs to have a tag. eg, https://github.com/openshift/managed-cluster-config/blob/043824ceb45090e8cae076655ad5d013c6b2e44d/resources/sts/hypershift/openshift_hcp_karpenter_controller_credentials_policy.json#L44

The customer should add the `red-hat=ture` tag to the KMS which used by ROSA, https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/html/install_clusters/rosa-hcp-creating-cluster-with-aws-kms-key  

> This is a request coming from https://redhat.atlassian.net/browse/ROSA-206

Does it mean the original policy was not supposed to have KMS permissions, but now we want to add it?  I am wondering why we need this extra permission if the original policy worked fine previously.  I am yet to read through [ROSA-206](https://redhat.atlassian.net/browse/ROSA-206) but maybe you have a quick answer :D

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Thanks for checking!

I am a bit concerning the non-redhat tag as any k8s cluster may have it.  When AWS reviewing the managed policy, they usually require us to scope it to only red hat managed resources.  This tag potentially gives us permission to take action on non-redhat resources.

The SID `ManageCSIMigratedVolumes` hints about migrations, not sure if we need it.  If we do need it, we might need to provide a use case as justification.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Thanks for clarifying!  If things are copied from upstream and we are using the same code, it has a strong evidence to include those.

The things we may need to pay attention is the tags, which could be different from the AWS one.

dfajmon:
> Does it mean the original policy was not supposed to have KMS permissions, but now we want to add it? I am wondering why we need this extra permission if the original policy worked fine previously. I am yet to read through [ROSA-206](https://redhat.atlassian.net/browse/ROSA-206) but maybe you have a quick answer :D

Apologies, the [ROSA-206](https://redhat.atlassian.net/browse/ROSA-206) is not linked to this and is different issue. 

The KMS policy is about enabling users to use already existing KMS keys for encryption/decryption for EBS volumes. I suppose this is more of a feature enablement that is not included in the basic example from upstream, but I believe it will be used by users that we would like to include.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Thanks for clarifying!

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: The `ec2:CopyVolumes` is for VPC cloning.

Introduced by:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2716

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2793


Use case:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/clone

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: This section is for `ec2:CopyVolumes` to READ the volume, so it is `ResourceTag`.  The above section `CreateAndCopyVolumesWithManagedTag` is for the copy action to WRITE a new volume, so it is `RequestTag`. This part makes sense.

For the resource pattern, it is following the upstream pattern in AmazonEBSCSIDriverPolicyV2, and every volume has a `vol-` prefix, so I have no concern.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Hi @dfajmon 

I would suggest we remove this section `CopyCSIMigratedVolumes`.


From what I read, this permission mainly allows the CSI driver to clone an existing volume, which doesn't have a managed tag.


For AWS's case, the old EBS created by the in-tree driver doesn't have the AWS managed tag, then this allow user to migrate the EBS volume.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2918


For the ROSA case, every resource created by ROSA should have a red-hat-managed tag, whenever the resources supports tag.


As we can see from the current policy, when creating a volume, it must request a tag `red-hat-managed`.  https://github.com/openshift/managed-cluster-config/blob/4120d40754657e4e3674ae3ea0ce9b753d1c54de/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json#L56   so we don't expect there will be a volume without the tag.


The proposed `ec2:ResourceTag/kubernetes.io/created-for/pvc/name` will allow the role to copy any k8s volume, maybe from other non RH clusters. This could be a risk.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Hi @dfajmon ,

I would suggest we remove this section "ManageCSIMigratedVolumes".

For ROSA, every volume should already been created with the `red-hat-managed` tag.

https://github.com/openshift/managed-cluster-config/blob/4120d40754657e4e3674ae3ea0ce9b753d1c54de/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json#L56

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: This is a feature introduced by 
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1554

Since we are having a rebase, and the permission is restrict with a `red-hat-managed` tag, this change lgtm.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: `LockSnapshot` is introduced by https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2797 . 

As we are rebasing the CSI driver and this action is restricted to `red-hat-managed`, it looks good to me.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Looks good and I can validate it in the existing CSI. 

See https://redhat.atlassian.net/browse/ROSAENG-1052 for details.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -124,10 +164,45 @@
...
+        "StringEquals": {
+          "aws:ResourceTag/red-hat-managed": "true"
+        },
+        "Null": {
+          "aws:TagKeys": "false"
+        },
+        "ForAllValues:StringNotEquals": {
+          "aws:TagKeys": [
+            "red-hat-managed",
+            "ebs.csi.aws.com/cluster-name",
```
feichashao: @dfajmon   do you know if `ebs.csi.aws.com/cluster-name` is EKS specific?  In the current latest ROSA, the EBS volume created by EBS CSI doesn't have this tag.  It has a `ebs.csi.aws.com/cluster` though.

I am thinking if we should include this in `ForAllValues:StringNotEquals`. But I am ok to keep it as it doesn't harm and user should not touch such key.

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -124,10 +164,45 @@
...
+            "kubernetes.io/created-for/pvc/name"
+          ]
+        }
+      }
+    },
+    {
+      "Sid": "KMSEncryptedVolumes",
```
feichashao, line 174: Hi @dfajmon ,

Please restrict the KMS permission to a `red-hat` tag.

You could refer this as an example:

https://github.com/openshift/managed-cluster-config/blob/043824ceb45090e8cae076655ad5d013c6b2e44d/resources/sts/hypershift/openshift_hcp_karpenter_controller_credentials_policy.json#L44

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
feichashao: Having another look, this missed condition to restrict on `red-hat-managed`.

The SID is also a bit wide, we shouldn't `AttachDetachVolumesToAnyInstance`, but we should only attach/detach to instances with `red-hat-managed`.

The EC2 instances (worker nodes) in ROSA should all have a tag `red-hat-managed`.

The original policy has a restriction https://github.com/openshift/managed-cluster-config/blob/4120d40754657e4e3674ae3ea0ce9b753d1c54de/resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json#L27 

@dfajmon  could you check this and add the resource restriction here?

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
dfajmon: Does not look like EKS specific, seems that this appears only when `k8s-tag-cluster-id` driver argument is configured. I  updated it to `"ebs.csi.aws.com/cluster",`

resources/sts/hypershift/openshift_hcp_cluster_csi_driver_ebs_operator_cloud_credentials_policy.json
```diff
@@ -124,10 +140,50 @@
...
+        "kms:GenerateDataKeyWithoutPlaintext",
+        "kms:CreateGrant"
+      ],
+      "Resource": "arn:aws:kms:*:*:key/*",
+      "Condition": {
+        "StringEquals": {
+          "aws:ResourceTag/red-hat": "true"
+        }
+      }
     }
```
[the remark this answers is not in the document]
feichashao: @dfajmon   could you follow the instruction to further scope the `CreateGrant`?

You might refer this policy as an example:
https://github.com/openshift/managed-cluster-config/blob/331e9573a9641bb416b0172c578945fa6e7c8b8e/resources/sts/hypershift/openshift_hcp_karpenter_controller_credentials_policy.json#L52-L66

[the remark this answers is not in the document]
dfajmon: Updated :+1:

feichashao:
/hold cancel
/lgtm

We are ready to get this merged.

feichashao:
### Note to approvers

I have reviewed and validated the managed policy change of this PR. Please refer to the comments in [ROSAENG-1325](https://redhat.atlassian.net/browse/ROSAENG-1325) .

This PR is changing the AWS IAM policy for ROSA HCP:
- It is changing the policy [ROSAAmazonEBSCSIDriverOperatorPolicy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/ROSAAmazonEBSCSIDriverOperatorPolicy.html).
- It is for the upcoming EBS CSI driver rebase.
- It doesn't remove any existing permissions, so it is compatible with the existing HCP clusters.


#### What permissions this PR adds

While it seem like a big change, it is effectively adding the below permissions.

**READ permissions**
ec2:DescribeAvailabilityZones
ec2:DescribeInstanceTypes
ec2:DescribeVolumeStatus

**WRITE permissions**
ec2:CopyVolumes
ec2:EnableFastSnapshotRestores
ec2:LockSnapshot
ec2:DeleteTags
ec2:CreateTags
kms:Decrypt
kms:GenerateDataKeyWithoutPlaintext
kms:CreateGrant

Those write permissions are scoped to the `red-hat-managed` tag for volumes and snapshots, and the `red-hat` tag for customer provided KMS key.

Those read permissions can not be scoped as per AWS documents.

PR #2696, opened 2026-04-08
SREP-4339: Preserve customer pull secret entries during ECR credential refresh

reedcort:
### What type of PR is this?
bug

The `ecr-credential-refresh` CronJob on zero-egress HCP clusters overwrites the entire `additional-pull-secret` with only ECR registry credentials every 4 hours, wiping out any customer-added entries for private registries.

Updated the `update_secret()` function to read the existing secret, use `python3` for JSON-aware merging to preserve all entries except those matching IDMS-discovered registries, and patch with the merged result. Added error guards for unreadable or malformed secret data. Only exact registries found in ImageDigestMirrorSets are replaced, so any customer-owned ECR registries are also preserved.

Only affects zero-egress HCP clusters — regular HCP clusters are unaffected.

Fixes [SREP-4339](https://redhat.atlassian.net/browse/SREP-4339)

**Root Cause:** The `update_secret()` function built a new `.dockerconfigjson` containing only ECR auths and patched it over the existing secret without merging.

**Fix:** When the secret already exists, the script now:
1. Reads the current `.dockerconfigjson` with error guards (validates oc get, non-empty, valid base64)
2. Uses `python3 -c` with `json.loads()`/`json.dumps()` for JSON-aware merging — handles any valid dockerconfigjson regardless of field ordering or auth format (reordered fields, `identitytoken`, `username`/`password` entries)
3. Passes IDMS-discovered registries via env var, only removes exact matches — customer-owned ECR registries are preserved
4. Patches the secret with the merged result

**Why python3:** The container image (`ocp-v4.0-art-dev` / node-resolver) includes Python 3.9. grep/sed-based JSON parsing is brittle — it misses entries where `auth` is not the first field.

Verified merge logic locally with tests covering:
- Customer non-ECR entries preserved
- Customer-owned ECR registries preserved
- Reordered fields (`auth` not first key) preserved
- `identitytoken`-based auth entries preserved
- Managed ECR tokens refreshed
- Multiple managed registries
- Invalid JSON rejected with error

Confirmed the bug on test cluster `2820d2ba-4d41-4ef2-ab16-998865e45fa6`.

- [x] If this is a new object that is not intended for the FedRAMP environment (if unsure, please reach out to team FedRAMP), please exclude it with: N/A - existing object

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deploy/acm-policies/50-GENERATED-hcp-ze-ecr-creds.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-hcp-ze-ecr-creds.Policy.yaml
@@ -207,6 +207,8 @@ spec:
 
                             # Build a dockerconfigjson with auth entries for all ECR registries
                             # using a single shared token, and write it into the additional-pull-secret.
+                            # When the secret already exists, merges ECR entries with existing customer
+                            # entries to preserve any user-added pull secrets (SREP-4339).
                             update_secret() {
                                 log "Updating $SECRET_NAME in $NAMESPACE with ${#ECR_REGISTRIES[@]} registry(s)..."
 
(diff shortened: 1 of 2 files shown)


deploy/acm-policies/50-GENERATED-hcp-ze-ecr-creds.Policy.yaml
```diff
@@ -222,19 +224,51 @@ spec:
...
+
+                                    # Merge using python3: remove only IDMS-discovered registries (managed
+                                    # by this script), preserve all other entries including any customer-owned
+                                    # ECR registries. Handles any valid dockerconfigjson regardless of field
+                                    # ordering or auth format.
+                                    local managed_registries
+                                    managed_registries=$(IFS=,; echo "${ECR_REGISTRIES[*]}")
+
+                                    local merged_json
+                                    if ! merged_json=$(MANAGED="$managed_registries" EXISTING="$existing_json" ECR="$ecr_json" python3 -c 'import json,os,sys;managed=set(os.environ["MANAGED"].split(","));existing=json.loads(os.environ["EXISTING"]);ecr=json.loads(os.environ["ECR"]);existing["auths"]={k:v for k,v in existing.get("auths",{}).items() if k not in managed};existing["auths"].update(ecr.get("auths",{}));print(json.dumps(existing))' 2>&1); then
```
joshbranham, line 255: This is pretty rough. I wonder if we should rather migrate this script to `python` altogether as a first pass, then add the merging in a follow up PR so it's a bit easier to reason about?

reedcort: I agree there doesn't seem to be a way to implement the fix entirely in bash and there doesn't seem any other way to handle the json manipulation other than using Python. So I'll close this PR and open a new one to convert the the entire refresh.sh script to Python.

PR #2698, opened 2026-04-09
[SREP-1815] Add Recording rule for sre operators missing to help drill down on missing operators that current recording rules are msising.

Mhodesty:
## Summary

  Adds deployment-level alerts for critical SRE operators to detect when operator namespaces exist but their deployments are missing.

  ## Problem

  Currently, if an operator installation partially fails (namespace created but deployment missing) or a deployment gets accidentally deleted, there's no alert to catch this state. This can silently break critical cluster functionality.

  ## Solution

  New PrometheusRule that fires critical alerts when:
  - Operator namespace exists
  - Expected deployment is absent for 30+ minutes

  Covers 7 critical SRE operators:
  - Route Monitor Operator
  - Managed Upgrade Operator
  - OCM Agent Operator
  - Must Gather Operator
  - Managed Velero Operator
  - Configure Alertmanager Operator
  - Managed Node Metadata Operator

 ## Concern
   - I am concerned that once this gets to production we are going to get a ton of alerts at once because this is a gap in our alerting. I believe this is a gap we should close, but still will create a lot of noise at least initially. 

  ## Testing

  - [ ] Validated Prometheus rule syntax
  - [ ] Tested alert fires when deployment missing
  - [ ] Verified alert clears when deployment present


deploy/sre-prometheus/centralized-observability/100-sre-operators-recording-rules.yaml
```diff
@@ -62,3 +62,22 @@ spec:
             or on(_id, name)
             sre:operators:succeeded:raw{exported_namespace !~ "openshift-.*"}
           record: sre:operators:succeeded
+        # Detect missing critical operator deployments
```
dustman9000, line 65: These are not our critical operators. MGO and MVO should be removed. We should also create a separate SSS for AVO that targets MCs/SCs.

deploy/sre-prometheus/centralized-observability/100-sre-operators-recording-rules.yaml
```diff
@@ -62,3 +62,13 @@ spec:
...
+            or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-managed-upgrade-operator", deployment="managed-upgrade-operator"}) * 1, "namespace", "openshift-managed-upgrade-operator", "", "")
+            or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-ocm-agent-operator", deployment="ocm-agent-operator"}) * 1, "namespace", "openshift-ocm-agent-operator", "", "")
+            or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-route-monitor-operator", deployment="route-monitor-operator-controller-manager"}) * 1, "namespace", "openshift-route-monitor-operator", "", "")
+            or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-splunk-forwarder-operator", deployment="splunk-forwarder-operator"}) * 1, "namespace", "openshift-splunk-forwarder-operator", "", "")
+            or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-monitoring", deployment="configure-alertmanager-operator"}) * 1, "namespace", "openshift-monitoring", "", "")
+          record: sre:operators:deployment_missing
```
tnierman: I believe we'll want to combine this with our `sre:telemetry:managed_labels` rule, per the [README for centralized-observabililty](https://github.com/openshift/managed-cluster-config/tree/master/deploy/sre-prometheus/centralized-observability#how-does-it-work)

tnierman: Per that same README, it looks like we'll also need to add this new recording rule to the `remoteWrite` [regex](https://github.com/openshift/managed-cluster-config/blob/master/resources/cluster-monitoring-config/config.yaml#L22) in order for it to be successfully forwarded to RHOBS

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -50954,6 +50954,32 @@ objects:
...
+
+              or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-splunk-forwarder-operator",
+              deployment="splunk-forwarder-operator"}) * 1, "namespace", "openshift-splunk-forwarder-operator",
+              "", "")
+
+              or label_replace(absent(kube_deployment_status_replicas{namespace="openshift-monitoring",
+              deployment="configure-alertmanager-operator"}) * 1, "namespace", "openshift-monitoring",
+              "", "")
+
+              '
```
tnierman: nit:
```suggestion
```

hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -50954,6 +50954,32 @@ objects:
           - expr: sre:operators:succeeded:raw{exported_namespace =~ "openshift-.*"}
               or on(_id, name) sre:operators:succeeded:raw{exported_namespace !~ "openshift-.*"}
             record: sre:operators:succeeded
+          - expr: 'label_replace(absent(kube_deployment_status_replicas{namespace="openshift-aws-vpce-operator",
```
tnierman: nit:
```suggestion
          - expr: label_replace(absent(kube_deployment_status_replicas{namespace="openshift-aws-vpce-operator",
```

Just thinking it'd be wise to keep this unquoted in order to match the other expressions in this file 🙂

tnierman: nevermind, this is a hack/ file

deploy/sre-prometheus/centralized-observability/100-sre-operators-recording-rules.yaml
```diff
@@ -62,3 +62,17 @@ spec:
             or on(_id, name)
             sre:operators:succeeded:raw{exported_namespace !~ "openshift-.*"}
           record: sre:operators:succeeded
+        # Detect missing critical operator deployments
+        # Returns 1 for each operator where deployment is missing (namespace may or may not exist)
+        - expr: |
+            label_replace(sre:telemetry:managed_labels, "cluster_version", "$0", "version", ".*")
+            * on () group_right (_id, cluster_version, provider)
+            (
+              label_replace(absent(kube_deployment_status_replicas{namespace="openshift-aws-vpce-operator", deployment="aws-vpce-operator"}) * 1, "namespace", "openshift-aws-vpce-operator", "", "")
```
tnierman: I think we'll want to exclude AVO from this selectorsyncset, as most clusters won't have it installed by design. There's a [management-cluster/ directory](https://github.com/openshift/managed-cluster-config/tree/master/deploy/sre-prometheus/management-cluster) this rule could be added to, which will add it to MCs only

clcollins:
/lgtm
/approve

🤖 Claude <claude@anthropic.com> commenting on behalf of @clcollins

PR #2702, opened 2026-04-15
[SREP-4470] Reduce write permissions in SREP RBAC in backplane

feichashao:
### What type of PR is this?
cleanup

For security and compliance reasons, we want to limit the permissions SRE have. Currently, SRE can perform many write actions (create/patch/delete) on many openshift platform namespaces.

This PR reduces most write permissions for the SREP RBAC.

If SRE need write operations to solve issues, they can use `ocm backplane elevate`, which requires a justification and go through the normal elevation process.

The PR keeps some write permissions which are not actually changing the cluster resources.
- The `create pods/portforward` is kept for monitoring, which is needed by `ocm backplane console`.
- The create `podsecuritypolicyreviews `, `podsecuritypolicyselfsubjectreviews`, `podsecuritypolicysubjectreviews` and `tokenreviews` are kept as they are actually read-only operations which is for checking auth related stuff.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Release Notes

* **Chores**
  * Restricted backplane SREP administrative access permissions by removing unnecessary elevated privileges for node operations, project administration, namespace management, and cluster backup capabilities.
  * Tightened access controls by replacing broad wildcard permissions with explicit read-only operations across administrative roles.
  * Enhanced security posture by limiting administrative capabilities to essential operations only.

### DIFF
--- a/deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
+++ b/deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
@@ -39,12 +39,6 @@ spec:
                             - apiGroups:
                                 - ""
                               resources:
-                                - nodes
-                              verbs:
-                                - patch
-                            - apiGroups:
-                                - ""
-                              resources:
                                 - nodes/proxy
                               verbs:
                                 - get
@@ -53,11 +47,8 @@ spec:
                               resources:
                                 - projects
                               verbs:
-                                - create
                                 - get
                                 - list
-                                - patch
-                                - update
                                 - watch
                             - apiGroups:
                                 - project.openshift.io
@@ -65,36 +56,22 @@ spec:
                                 - projects
                                 - projectrequests
                               verbs:
-                                - create
                                 - get
                                 - list
-                                - patch
-                                - update
                                 - watch
-                                - delete
                             - apiGroups:
                                 - ""
                               resources:
                                 - namespaces
                                 - namespaces/finalize
                               verbs:
-                                - create
                                 - get
                                 - list
-                                - patch
-                                - update
                                 - watch
-                                - delete
-                            - apiGroups:
-                                - certificates.k8s.io
-                              resources:
-                                - certificatesigningrequests/approval
-                              verbs:
-                                - update
                             - nonResourceURLs:
                                 - '*'
                               verbs:
-                                - '*'
+                                - get
                             - apiGroups:
                                 - authentication.k8s.io
                               resources:
@@ -146,12 +123,6 @@ spec:
                                 - list
                                 - watch
                             - apiGroups:
-                                - machine.openshift.io
-                              resources:
-                                - machines
-                              verbs:
-                                - delete
-                            - apiGroups:
                                 - apiserver.openshift.io
                               resources:
                                 - apirequestcounts
@@ -167,7 +138,6 @@ spec:
                                 - get
                                 - list
                                 - watch
-                                - delete
                             - apiGroups:
                                 - avo.openshift.io
                               resources:
@@ -176,7 +146,6 @@ spec:
                                 - get
                                 - list
                                 - watch
-                                - delete
                     - complianceType: mustonlyhave
                       metadataComplianceType: musthave
                       objectDefinition:
@@ -188,37 +157,9 @@ spec:
                             - apiGroups:
                                 - ""
                               resources:
-                                - pods/eviction
-                              verbs:
-                                - create
-                            - apiGroups:
-                                - ""
-                              resources:
                                 - pods/portforward
                               verbs:
                                 - create
-                            - apiGroups:
-                                - batch
-                              resources:
-                                - jobs
-                              verbs:
-                                - delete
-                                - deletecollection
-                                - create
-                            - apiGroups:
-                                - build.openshift.io
-                              resources:
-                                - builds
-                              verbs:
-                                - delete
-                                - deletecollection
-                            - apiGroups:
-                                - ""
-                              resources:
-                                - pods
-                              verbs:
-                                - delete
-                                - deletecollection
                             - apiGroups:
                                 - security.openshift.io
                               resources:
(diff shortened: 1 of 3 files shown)


feichashao:
/hold

Now it is ready for review, but don't merge it until user communication is sent.

deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
```diff
@@ -3,71 +3,46 @@ kind: ClusterRole
...
-  resources:
-  - certificatesigningrequests/approval
-  verbs:
-  - update
-# SRE can get and post to all endpoints
+# SRE can get all endpoints
 - nonResourceURLs:
   - '*'
   verbs:
-  - '*'
```
bergmannf, line 70: Would we potentially want all the read verbs here? So adding 'list' and 'watch' as well?

bergmannf:
Apart from the really small nitpick these changes look reasonable to require elevation in the future.

deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
```diff
@@ -145,12 +122,6 @@ spec:
                                 - get
                                 - list
                                 - watch
-                            - apiGroups:
-                                - machine.openshift.io
-                              resources:
-                                - machines
-                              verbs:
-                                - delete
```
Makdaam, line 153: In normal circumstances a machine delete shouldn't impact any workload on the cluster (the machine enters terminating state and the machine api operator is supposed to create a replacement then drain the original).
Is there a specific reason why we want to put Machine replacement behind elevation or is it a blanket move of all changes to cluster state to elevated access?

Same question about the CertificateSigningRequests below, and other auto-restored resources.

bergmannf: I was thinking the same thing - we (sometimes) had to delete machines because the workers might have been broken - but I think in that cases having a record of this isn't too bad.
Either the PLEG was broken or there was some other issue with the node that shouldn't be a 'normal' occurrence.
For the CSRs I think those usually need removal or approval when hibernation went bad - which is still just a tech preview feature?

Makdaam: Just to clarify - this is not a blocking comment. I was just unable to find the decision in the related tickets. Moving this behind elevated access doesn't impact SRE work significantly.

bmeng: The goal is to make the ReadOnly as default, regardless the operation would impact things or not.

For all the existing operations, they should be all reviewed and accepted as safe for SRE.

hack/00-osd-managed-cluster-config-production.yaml.tmpl
```diff
@@ -5426,48 +5420,31 @@ objects:
...
-                    - certificates.k8s.io
-                    resources:
-                    - certificatesigningrequests/approval
-                    verbs:
-                    - update
                   - nonResourceURLs:
                     - '*'
                     verbs:
-                    - '*'
+                    - get
```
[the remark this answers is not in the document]
feichashao: Thank you Rabbit, it is a good idea. But I will keep it as-is for now. This task is focusing on removing the permissions, I have already remove the write access which is much safer.

For allowlist, it would be better to do it as a follow-up.

deploy/acm-policies/50-GENERATED-backplane-srep.Policy.yaml
feichashao: @Makdaam @bergmannf Thank you for your review!

This PR is mainly moving all the write access behind elevation. It doesn't mean an action is not needed by SRE, nor the action is risky, but we don't want to just hand them over directly. There are some requirements from compliance and management to move backplane to a more restricted and audited place.

SRE can still perform the delete machine and CSR actions. SRE just need to elevate and provide a justification.

deploy/backplane/srep/10-srep-admins-cluster.ClusterRole.yml
feichashao: Yes, we want to allow all the read operations here.

Just double checked.  For non-resource URLs, k8s only support http verbs, [ref](https://github.com/kubernetes/kubernetes/blob/b2f73c0d6b427e2ab5ba225375aaefc0b9bc45b2/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L49-L51)
```
	// Verb is the kube verb associated with the request for API requests, not the http verb.  This includes things like list and watch.
	// for non-resource requests, this is the lowercase http verb
	Verb string
```

It will also treat `HEAD` as `GET`, [ref](https://github.com/kubernetes/kubernetes/blob/b2f73c0d6b427e2ab5ba225375aaefc0b9bc45b2/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L180),
```
		case MethodGet, MethodHead:
			requestInfo.Verb = "get"
```

So allowing `get` here should be sufficient.

feichashao:
Tested applying these 2 updated clusterrole locally.

I had a random test and it can remove the write permissions, and we can use elevation for it.

Before:
```
oc delete pod -n openshift-monitoring alertmanager-main-0
pod "alertmanager-main-0" deleted
```

After:
```
oc delete pod -n openshift-monitoring alertmanager-main-0
Error from server (Forbidden): pods "alertmanager-main-0" is forbidden: User "system:serviceaccount:openshift-backplane-srep:6752b7bff460e5d07be60d10f1835b91" cannot delete resource "pods" in API group "" in the namespace "openshift-monitoring"

ocm backplane elevate "test new rbac in my test cluster" -- delete pod -n openshift-monitoring alertmanager-main-0
pod "alertmanager-main-0" deleted
```

PR #2709, opened 2026-04-18
SREP-4577: Add keep_firing_for to MHC short circuit alert to reduce flapping

dustman9000:
## Summary

- Adds keep_firing_for: 1h to the MachineHealthCheckUnterminatedShortCircuitSRE alert rule
- Keeps the original for: 30m detection threshold unchanged
- Reduces PagerDuty noise from flapping MHC alerts by keeping the alert firing through brief recovery windows

## Problem

Observed 40+ PagerDuty incidents in 29 hours on dapa-c1b.zl8c.p1 due to machines cycling between unhealthy/healthy states every ~45 minutes. Each cycle briefly cleared the alert (5 min recovery) before it re-fired, creating a new PD incident each time.

## Solution

keep_firing_for: 1h means once the alert fires (after the 30m for: threshold), it stays firing through recovery windows up to 1 hour. This collapses the flap into a single PD incident while preserving detection sensitivity for new occurrences.

## Note on hack/ changes

The hack/ template diffs include re-ordering of osd-cluster-acks-* SelectorSyncSet resources. This is a pre-existing make idempotency issue from the recent 4.22 credential policies merge (#2707), which did not regenerate hack/ templates. Running make on a clean upstream/master checkout produces the same re-ordering. The only actual change in this PR is the keep_firing_for: 1h addition.

Jira: https://redhat.atlassian.net/browse/SREP-4577

## Test plan

- [ ] Verify keep_firing_for is supported on OCP Prometheus (available since Prometheus 2.42 / OCP 4.13+)
- [ ] Deploy to stage and verify alert behavior on a cluster with MHC flapping
- [ ] Confirm no regression in alert detection for persistent MHC short-circuit events

## Summary by CodeRabbit

* **Chores**
  * Extended a Prometheus alert’s post-evaluation firing retention by 1 hour for improved monitoring persistence.
  * Retargeted multiple cluster selector sync sets to new OpenShift minor/patch versions for compatibility.
  * Updated cloud credential patch annotations to align with the revised selector version mappings.

### DIFF
--- a/deploy/sre-prometheus/100-machine-health-check-unterminated-short-circuit.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-machine-health-check-unterminated-short-circuit.PrometheusRule.yaml
@@ -14,6 +14,7 @@ spec:
       expr: |
         mapi_machinehealthcheck_short_circuit == 1
       for: 30m
+      keep_firing_for: 1h
       labels:
         severity: critical
         namespace: '{{ $labels.namespace }}'


hack/00-osd-managed-cluster-config-integration.yaml.tmpl
```diff
@@ -49895,6 +49895,7 @@ objects:
 
               '
             for: 30m
+            keep_firing_for: 1h
```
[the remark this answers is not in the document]
dustman9000: Not a concern for this PR as OSD/ROSA does not support 4.12 and older clusters.

deploy/sre-prometheus/100-machine-health-check-unterminated-short-circuit.PrometheusRule.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       expr: |
         mapi_machinehealthcheck_short_circuit == 1
       for: 30m
+      keep_firing_for: 1h
```
joshbranham, line 17: Iirc this is only valid on certain versions of Prometheus, worth double checking.

joshbranham:
Fixing the template issue in https://github.com/openshift/managed-cluster-config/pull/2711 standalone since I also have to open a PR and can't merge without it

deploy/sre-prometheus/100-machine-health-check-unterminated-short-circuit.PrometheusRule.yaml
dustman9000: Right and that was flagged by coderabbit. We do not have any versions that are supported where this would fail.

PR #2710, opened 2026-04-20
SREP-4471 Remove Write permissions in LPSRE RBAC

feichashao:
### What type of PR is this?
cleanup

For security and compliance reasons, we want to limit the permissions SRE have. Currently, SRE can perform many write actions (create/patch/delete) on many openshift platform namespaces.

This PR reduces most write permissions for the LPSRE RBAC.

If SRE need write operations to solve issues, they can use ocm backplane elevate, which requires a justification and go through the normal elevation process.

The PR keeps some write permissions which are not actually changing the cluster resources.

The `create pods/portforward` is kept for monitoring, which is needed by ocm backplane console.
The `create podsecuritypolicyreviews , podsecuritypolicyselfsubjectreviews, podsecuritypolicysubjectreviews and tokenreviews` are kept as they are actually read-only operations which is for checking auth related stuff.

https://redhat.atlassian.net/browse/SREP-4471

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Chores**
  * Reduced administrative permissions across multiple cluster roles by removing destructive and write actions (deletes, creates, patches) for jobs, builds, pods, PVCs, operators, backups, and scaling controls.
  * Converted many mutating permissions to read-only (get/list/watch) for monitoring, config, Tekton, and project/namespace resources.
  * Preserved essential observation and eviction/portforward capabilities to maintain troubleshooting access.


feichashao:
/hold

Communicating with LPSRE regarding the change.

deploy/backplane/lpsre/rhmi/01-rhmi-lpsre-admins-project.ClusterRole.yaml
```diff
@@ -67,19 +44,9 @@ rules:
...
   - persistentvolumeclaims
   verbs:
-  - create
   - get
   - list
-  - patch
-  - update
   - watch
-# LP SRE can interact with installplans
-- apiGroups:
```
Nanyte25, line 77: @feichashao Aligned with the direction of this PR overall, but flagging one carve-out request from the RHOAM side:We need to retain the ability to manually patch RHOAM Install-plans without elevation. All RHOAM upgrades are service-affecting, and manual Install-plan approval is the single most common upgrade-time intervention we perform. Blast radius is bounded — patch on an IP only lets us flip spec.approved=true on a plan OLM has already generated, we can't synthesize new state with it. Requiring backplane elevate per-approval during an active upgrade window is a real MTTR cost in exactly the moments we're already time-pressured.

Proposing we keep this rule:

```
# LP SRE can interact with installplans
- apiGroups:
  - "operators.coreos.com"
  resources:
  - installplans
  verbs:
  - "patch"
```
I would ask the same for  the equivalent block in `rhmi/01-rhmi-lpsre-admins-project.ClusterRole.yaml` for consistency.

feichashao: Hi @Nanyte25 , as discussed over slack, we will keep it removed to be consistent with other write permissions, and will monitoring if this brings any friction for handling upgrades.

PR #2734, opened 2026-05-01
SREP-4556: Fix SRE recording rule evaluation failures caused by series cardinality

dustman9000:
## Problem

SRE recording rules (`sre-telemetry-managed-labels.rules` and `sre-operators.rules`) produce `group_left` cardinality errors during OLM operator initialization on fresh clusters, causing the `shouldn't have failing rules evaluation` conformance test to fail.

Root cause: `subscription_sync_total` can have multiple series per subscription name (different `installedCSV`, `source` label combinations), and the telemetry join on synthetic `sre="true"` label is vulnerable to stale/duplicate series.

## Fix

- **`sre:operators:succeeded:olm`**: Wrap `subscription_sync_total` with `max by (name, channel, package)` before the `^ on(name) group_left(channel, package)` join
- **`sre:telemetry:managed_labels`**: Wrap each metric with `max by(...)` on join-relevant labels before the `* on(sre) group_left()` joins

The PKO rule already had `max by(...)` wrapping. The other 3 rules in `sre-operators.rules` use `or` unions or `absent()` and are not affected.

## Testing

Validated on two stage clusters:

1. **Stable cluster** (hs-sc-a8p1rblbg, 4.21.11): Applied fix, waited 2 min. Zero new evaluation failures (rate 0.0). Recording rule output unchanged.

2. **Fresh cluster during OLM initialization** (osde2e-bf2ze, 4.21.11): Applied fix immediately after Prometheus came up with only 4/23 operators installed. Zero evaluation failures across all 57 SRE rule groups during the initialization window. `sre:operators:succeeded:olm` correctly tracked 4 operators, `sre:telemetry:managed_labels` produced 1 series.

The `max by(...)` guards prevent `group_left` many-to-many errors when `subscription_sync_total` has multiple series per operator during OLM installation, without changing the recording rule semantics.

Jira: https://redhat.atlassian.net/browse/SREP-4556

## Summary by CodeRabbit

* **Chores**
  * Refined Prometheus recording rules to de-duplicate and normalize metric inputs, ensuring a single representative value per logical group. Improves accuracy and consistency of operator lifecycle and managed-label telemetry across environments, reducing spurious counts and making related alerts and dashboards more reliable.

### DIFF
--- a/deploy/sre-prometheus/centralized-observability/100-sre-operators-recording-rules.yaml
+++ b/deploy/sre-prometheus/centralized-observability/100-sre-operators-recording-rules.yaml
@@ -20,7 +20,7 @@ spec:
               sum by (exported_namespace, name, version, channel, package) (
                   label_replace(csv_succeeded, "name", "$1", "name", "([^.]+).*")
                 ^ on (name) group_left (channel, package)
-                  subscription_sync_total{name=~"addon-operator|aws-vpce-operator|custom-domains-operator|managed-node-metadata-operator|managed-upgrade-operator|must-gather-operator|ocm-agent-operator|osd-metrics-exporter|rbac-permissions-operator|openshift-splunk-forwarder-operator|aws-account-operator|certman-operator|cloud-ingress-operator|configure-alertmanager-operator|deadmanssnitch-operator|deployment-validation-operator|dynatrace-operator|gcp-project-operator|managed-velero-operator|observability-operator|opentelemetry-operator|pagerduty-operator|route-monitor-operator"}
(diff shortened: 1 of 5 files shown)


clcollins:
Held - saw Trevor's comments in Slack with followup question.  Feel free to un-hold.

dustman9000:
For posterity:
```
on (name) group_left (channel, package)
max by (name, channel, package) (subscription_sync_total...)
```

> this works if you only have one (channel, package) tuple in there for a given name.  I don't know the subscription_sync_total metric well enough to say if that's the case.  But a more defensive approach is:

```
...
on (name) group_left (channel, package)
topk by (name) (1, subscription_sync_total...)
```

> which says "if you find multiple (channel, package) tuples for a given name, just pick the biggest, thanks".  You could also do group by (name, channel, package) (...) instead of max by ... if you are confident in single-tuple, and want to save the PromQL engine the work of finding the tuple with the largest value. although group doesn't work if you actually need to get the metric's value, which I guess means you can't use it for these sum by rules.  You could use topk though.

dustman9000:
Reverted back to max by (name, channel, package). Verified on both a stable stage cluster (hs-sc-0jnsfblbg, 4.18.17, running 2+ months) and a fresh cluster (hs-sc-a8p1rblbg, 4.21.11) that subscription_sync_total has a single (channel, package) tuple per operator name. topk would be a valid defensive option but is not needed here.

PR #2738, opened 2026-05-05
Add make promrule-check to validate PrometheusRule conventions

bmeng:
Port the alerting rule checks from openshift/origin (test/extended/prometheus/prometheus.go) to catch violations before merge instead of at cluster deploy time.

Checks: valid severity label, description + summary annotations, runbook_url for critical alerts, and runbook_url format. Pre-existing violations are tracked in exception lists.

Move all the prometheusrule definitions into sre-prometheus dir. 
Integrate the new python script in the existing pr_check.sh 

### What type of PR is this?

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Improvements**
  * StaticPodMissing alerts are now scoped to component namespaces (kube-apiserver, kube-controller-manager, kube-scheduler) for clearer, targeted alerting.
  * Prometheus rules for network live migration moved under a dedicated selector for clearer management.

* **Chores**
  * Added a validation tool and CI check to enforce alerting conventions (severity, annotations, runbooks, linting) across PrometheusRule configs.


Makefile
```diff
@@ -95,6 +98,14 @@ enforce-backplane-rules:
 		${ENFORCE_BACKPLANE_RULES};\
 	fi
 
+.PHONY: promrule-check
+promrule-check:
```
joshbranham: You should invoke this from the pr-check script so it fails in prow.

bmeng: I am not sure if I understand your point, but there is a follow up change to enable it in prow for pre-merge checking https://github.com/openshift/release/pull/78799/

Is this what you want?

joshbranham: I mean that the repo already has a "suite" of PR check tests covered by this job and you can just extend it vs adding an individual new prow test https://github.com/openshift/release/blob/eb3403199fb8ff91fdda595a2741b3068f2d429e/ci-operator/config/openshift/managed-cluster-config/openshift-managed-cluster-config-master.yaml#L10-L13

But I suppose it doesn't matter which path you take, I would prefer us to remain consistent.

bmeng: Ah, thanks, I did not realize there are some checks there already for prometheusruels. Can see how to get the additional check into the existing one

joshbranham: Right, that job runs this script which does a few things in CI https://github.com/openshift/managed-cluster-config/blob/master/hack/pr_check.sh

joshbranham (commented):
/lgtm

PR #2761, opened 2026-05-27
Exlude alert PrometheusRemoteStorageFailures as this one appears to c…

rolandmkunkel:
…ommonly be in warning state in staging clusters, but doesn't have meaning for SRE

### What type of PR is this?
feature

It exlucdes alert "PrometheusRemoteStorageFailures" to be a trigger for must-gathers. This alert seems be constantly firing on at least staging clusters, without any significance for SRE

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined alert filtering behavior to exclude additional alert types from triggering automatic diagnostics collection, reducing unnecessary diagnostic data gathering and improving system efficiency.

### DIFF
--- a/deploy/sre-prometheus-must-gather/100-create-must-gather.PrometheusRule.yaml
+++ b/deploy/sre-prometheus-must-gather/100-create-must-gather.PrometheusRule.yaml
@@ -12,7 +12,7 @@ spec:
       rules:
         - alert: CreateMustGather
           expr: |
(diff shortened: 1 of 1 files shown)


deploy/sre-prometheus-must-gather/100-create-must-gather.PrometheusRule.yaml
```diff
@@ -12,7 +12,7 @@ spec:
...
+            max(ALERTS{alertstate="firing",alertname!="CreateMustGather",namespace=~"openshift-.*|kube-.*|default|redhat-.*",severity!~"info|alert|none",namespace!~"openshift-customer-monitoring|openshift-operators|openshift-storage|openshift-compliance|openshift-operators-redhat|openshift-user-workload-monitoring",alertname!~"Watchdog|CannotRetrieveUpdates|KubeQuotaExceeded|KubeQuotaFullyUsed|CPUThrottlingHigh|NodeFilesystemSpaceFillingUp|NodeFilesystemFilesFillingUp|NodeFilesystemAlmostOutOfSpace|NodeFilesystemAlmostOutOfFiles|NodeFileDescriptorLimit|CustomResourceDetected|ImagePruningDisabled|PodDisruptionBudgetLimit|KubeJobFailed|PrometheusRuleFailures|HAProxyDown|HAProxyReloadFail|CertificateIsAboutToExpire|KubeAPIErrorBudgetBurn|PrometheusRemoteStorageFailures"}) > 0
```
joshbranham, line 15: Is this not a constantly moving target with OCP versions, that is, excluding alerts from triggering is something that won't be easy to keep up with? We should instead be allow-listing alerts we _do_ want to trigger a must gather on.

rolandmkunkel: That's true. For https://redhat.atlassian.net/browse/SREP-4025 we want to create must-gathers for everything that pages primary for a limited time. This expression roughly approximates what is configured in CAMO to be sent to PD. CAMO effectively uses a block list as well, so this matches the approach.

If we had data about the most common alerts, we could also use an approximate allow list.

PR #2762, opened 2026-05-27
exclude fedramp from OADP cleanup. 

annelson-rh:
### What type of PR is this?
bug

excludes fedramp from oadp cleanup job. 
it currently causes the clustersyncs to the MC clusters to show failed, since we do not have oadp deployed yet. 

## Summary by CodeRabbit

* **Chores**
  * Updated deployment configuration to exclude clusters with the FedRAMP label from HyperShift OADP cleanup operations, ensuring compliance-labeled clusters are not processed by the cleanup service.


typeid (approved):
/lgtm
/approve

PR #2766, opened 2026-05-29
ROSAENG-8223: Fix ManifestWork race condition in capa-annotator

joshbranham:
### What type of PR is this?
bug

Replace index-based JSON patch with atomic read-modify-write to prevent racing with other controllers. The old approach read the manifests array to find a numeric index, then patched by that index — if the array was modified between read and patch, the wrong manifest could be patched.

Now uses oc get | jq (match by kind) | oc replace with a retry loop for resourceVersion conflicts. Also fixes break→continue bug that stopped processing all remaining clusters when a management cluster was found.

Fixes [ROSAENG-8223](https://redhat.atlassian.net/browse/ROSAENG-8223)

I ran this in staging by installing a 4.16 cluster, increasing the fixed version for 4.16 to be > my cluster, asserting it gets patched, and then flipping it back to test the patch gets removed.

**Adding**
```
❯ oc logs capa-annotator-4-16-29672170-ldv2j -n openshift-capa-annotator
should patch: version 4.16.34 is not >= x.y.35
patching cluster: 2qllgnu4h4qc8jpa8irflo5j7bbpi2sq
successfully patched cluster 2qllgnu4h4qc8jpa8irflo5j7bbpi2sq
-------------------------------------------------------------------------
```

**Removal**
```
❯ oc logs capa-annotator-4-16-29672175-54vlc -n openshift-capa-annotator
don't patch: version 4.16.34 is >= x.y.23
cluster 2qllgnu4h4qc8jpa8irflo5j7bbpi2sq does not need the override because it's Z stream is >= 23
removing hypershift.openshift.io/capi-provider-aws-image annotation
patching cluster: 2qllgnu4h4qc8jpa8irflo5j7bbpi2sq
successfully patched cluster 2qllgnu4h4qc8jpa8irflo5j7bbpi2sq
-------------------------------------------------------------------------
```

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Improvements**
  * Bounded retries and improved conflict handling for more reliable cluster annotation updates.
  * Atomic read-modify-write update flow to prevent partial or erroneous manifest changes.
  * Expanded RBAC to allow manifestwork updates (adds update verb) for smoother operations.
* **Bug Fixes**
  * Removed stale creationTimestamp metadata from configuration.

### DIFF
--- a/deploy/osd-25821-capa-annotator/patch.sh
+++ b/deploy/osd-25821-capa-annotator/patch.sh
@@ -13,6 +13,8 @@ if ! command -v python &>/dev/null; then
   exit 1
 fi
 
+MAX_RETRIES=3
+
 # Get all manifestwork objects and extract their names
 managedclusters=$(oc get managedclusters -l openshiftVersion-major-minor=${MAJOR_MINOR_VER} -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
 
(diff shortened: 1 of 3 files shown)


deploy/osd-25821-capa-annotator/patch.sh
```diff
@@ -36,26 +38,53 @@ for clusterID in ${managedclusters[@]};do
...
+    replace_output=$(echo "$modified" | oc replace -f - 2>&1)
+    replace_rc=$?
+
+    if [[ $replace_rc -eq 0 ]]; then
+      echo "successfully patched cluster $clusterID"
+      break
+    fi
+
+    if echo "$replace_output" | grep -qi "conflict\|the object has been modified"; then
+      echo "conflict on attempt $attempt/$MAX_RETRIES for cluster $clusterID, retrying..."
```
rafael-azevedo, line 77: should we put a small back off or sleep here. I am wondering if we launch quick retries its possible an active CS reconcile could cause an issue.

rafael-azevedo (approved):

PR #2786, opened 2026-06-10
Adds etcdMembersDownSRE with CPMS scaling awareness

iamkirkbater:
Adds a new alert, etcdMembersDownSRE that will only fire if the etcd members are down and CPMS is not in the process of scaling up, making this alert not fire when it's unactionable.

### What type of PR is this?

Fixes [ROSAENG-59423](https://redhat.atlassian.net/browse/ROSAENG-59423)

Testing Methodology:

* Spin up new cluster, wait for it to stabilize.
* Gather etcd pods and pick one to be the martyr and one to be the one doing the work. Doesn't matter which ones. In my case, I used `10.0.7.102` as the sacrificial member and `10.0.48.217` as the one doing the work.
* Run the following loop(s) to keep one etcd member down:
    ```bash
    while true; do oc exec -it etcd-ip-10-0-48-217.ec2.internal -n openshift-etcd --as backplane-cluster-admin -- /bin/bash -c "while true; do etcdctl member list | grep '10.0.7.102' | awk '{print $1}' | cut -d, -f1 | xargs etcdctl member remove; sleep 5; done"; done
    ```
    * This runs a loop (because sometimes the connection to your etcd host dies out) running an exec command that runs another loop to grep for the member id of the sacrificial etcd member. 
* Let this run for a long time. It's a 20 minute window for this to fire the etcdMembersDown original alert.
* Validate that the alert fires. As soon as the original etcdMembersDown alert fires the new one should also fire.

* update CPMS to scale the nodes to the next larger size.
* once the new CPMS machine creates the node and it joins the cluster, the etcdMembersDownSRE alert should resolve.

### DIFF
--- a/deploy/sre-prometheus/100-etcd-members-down.PrometheusRule.yaml
+++ b/deploy/sre-prometheus/100-etcd-members-down.PrometheusRule.yaml
@@ -21,7 +21,7 @@ spec:
       expr: |
         count(ALERTS{alertname="etcdMembersDown", alertstate="firing"}) >= 1
           and on()
-        count(kube_node_role{role=~"master|control-plane"}) <= 3
+        (count(kube_node_role{role="control-plane"}) or vector(0)) <= 3
       labels:
         severity: critical
         namespace: openshift-etcd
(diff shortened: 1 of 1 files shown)


typeid (approved):
/lgtm
/approve

typeid (approved):

PR #2787, opened 2026-06-11
exclude some noisy alerts from must-gather trigger

rolandmkunkel:
### What type of PR is this?
feature

This adds three more alerts to the blocklist for must-gather creation.

The alert uses a blocklist approach (exclude specific alertnames and namespaces) rather than an allowlist because the actual PagerDuty routing in configure-alertmanager-operator also uses a blocklist. Both warning and critical severity alerts page SRE, so we can't simply filter on severity. An allowlist of 74+  SRE-custom alerts plus an unknown number of upstream OpenShift alerts would be brittle and hard to maintain.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
 This alert-based approach is temporary. The long-term plan is to move the must-gather triggering logic entirely into CAD, which will receive a webhook callback for each alert, eliminating the need to replicate routing logic in PromQL. 

Fixes https://redhat.atlassian.net/browse/ROSAENG-5563

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **Bug Fixes**
  * Enhanced alert filtering for automated must-gather data collection by expanding the exclusion list for non-critical alert types. This prevents unnecessary investigations from being triggered, reducing system overhead and improving operational efficiency while ensuring critical system issues remain properly monitored.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


rolandmkunkel:
This extends https://github.com/openshift/managed-cluster-config/pull/2761
I have created this follow up ticket to implement this more cleanly: https://redhat.atlassian.net/browse/ROSAENG-59444
The current design was dictated by the original DDR.

PR #2788, opened 2026-06-11
ROSAENG-2104: Re-enable managed boot images on existing AWS OSD 4.19+ clusters

tnierman:
### What type of PR is this?
_feature_

Existing AWS OSD clusters that upgrade to 4.19 retain a stale managedBootImages patch from the old SelectorSyncSet, because Hive does not revert patches when a SelectorSyncSet stops matching. Add a new SelectorSyncSet that targets AWS OSD 4.19+ and patches managedBootImages to null, removing the stale disable and allowing the MCO to manage boot images.

_Fixes #_ https://redhat.atlassian.net/browse/ROSAENG-2104

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Chores**
  * Enabled managed boot images for cluster operations via an updated machine configuration (always applying the setting).
  * Added a selector-based machine configuration sync that targets integration AWS environments while excluding older OpenShift versions (4.0–4.18) and the Rosa product, ensuring updates apply only where intended.

### DIFF
--- a/deploy/osd-machine-configuration/osd-aws-419/config.yaml
+++ b/deploy/osd-machine-configuration/osd-aws-419/config.yaml
@@ -2,6 +2,9 @@ deploymentMode: "SelectorSyncSet"
 selectorSyncSet:
   resourceApplyMode: "Sync"
   matchExpressions:
+  - key: api.openshift.com/environment
+    operator: In
+    values: ["integration"]
   - key: hive.openshift.io/version-major-minor
     operator: NotIn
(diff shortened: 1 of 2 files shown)


deploy/osd-machine-configuration/osd-aws-419/config.yaml
dustman9000, line 1: Let's please only target the integration environment first. We can then verify and roll out to stage and production once verified.

aliceh: I opened a [PR to Trevor's branch](https://github.com/tnierman/managed-cluster-config/pull/1 ) that will target integration

aliceh: Looks like Trevor merged my change. Thanks @tnierman !

PR #2810, opened 2026-07-10
fix: use correct resendWait field in proxy-ca-validation ManagedFleetNotifications

AlexSmithGH:
### What type of PR is this?
_bug_
The proxy-ca-validation.yaml was using `resendInterval` instead of the correct CRD field name `resendWait`, causing the SelectorSyncSet to fail with "spec.fleetNotification.resendWait: Required value". This blocked all four proxy-ca MFNs from being created on management clusters.

Fixes: https://redhat.atlassian.net/browse/ROSAENG-61675

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated fleet notification retry behavior to preserve the configured waiting periods before resending notifications.
  * Existing notification content, severity, summaries, and reference links remain unchanged.
  * Resend timing continues to use the established 24- and 48-hour intervals.


joshbranham (commented):
/lgtm
/approve

PR #2828, opened 2026-07-27
ROSAENG-62424 have a dedicated provisioning_failure_notify for spot and send dedicated SL

feichashao:
### What type of PR is this?
feature

This PR is to exclude the spot nodepool in the `provisioning_failure_notify` recording rule and make it for on-demand nodepool.

Originally, the recording rule assumes the nodepool are all on-demand, so nodes should all be able to provision, otherwise there could be a problem.

But for spot instances, it is likely not to be able to provision due to pricing or AWS capacity, which is expected. We don't want to trigger alerts if the nodepool is running spot.

For the spot instance nodepool, we add `provisioning_failure_notify_spot` delicately, and add a SL template for spot nodepool. 

https://redhat.atlassian.net/browse/ROSAENG-62424

Validated, steps:
- Create a HCP cluster in staging.
- Create a nodepool manually in the HC namespace in the corresponding MC.
- Go to Grafana and check the metrics.
- Because I can provision the spot nodepool smoothly, I changed the PromQL from `<` to `==` to see how the recording rule behave.
```
(hypershift_nodepools_available_replicas / hypershift_nodepools_size == 1)
```
(keep the other conditions same as this PR)
- Without the change in this PR, I can see all nodepools including spot.
- With the change in this PR, the spot nodepool is excluded, the other nodepool / other clusters remain the same.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## New Features

- Added dedicated monitoring and warning notifications for Spot node pool provisioning failures.
- Added Spot-specific notification content, resend intervals, and limited-support status.

## Bug Fixes

- Improved provisioning failure monitoring by separating Spot and on-demand node pools.
- Clarified on-demand alerts and reduced misleading notifications by ensuring each node pool type receives relevant alerts.
- Updated notification details to provide clearer, more actionable provisioning-failure information.

### DIFF
--- a/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
+++ b/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
@@ -6,9 +6,9 @@ metadata:
 spec:
   fleetNotification:
     name: nodepool-provisioning-failure-spot
-    summary: Spot instance nodepool has not reached its desired size
+    summary: Spot instance machinepool has not reached its desired size
     notificationMessage: |-
-      One or more Spot instance nodepools in your ROSA cluster have not reached their desired size. This can prevent your workloads from running properly.
+      One or more Spot instance machinepools in your ROSA cluster have not reached their desired size. This can prevent your workloads from running properly.
 
       Common causes include:
       - Your maximum Spot price is below the current Spot price for the selected instance type
--- a/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
+++ b/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
@@ -1,0 +1,22 @@
+apiVersion: ocmagent.managed.openshift.io/v1alpha1
+kind: ManagedFleetNotification
+metadata:
+  name: nodepool-provisioning-failure-spot
+  namespace: openshift-ocm-agent-operator
+spec:
+  fleetNotification:
+    name: nodepool-provisioning-failure-spot
+    summary: Spot instance nodepool has not reached its desired size
+    notificationMessage: |-
+      One or more Spot instance nodepools in your ROSA cluster has not reached its desired size. This can prevent your workloads from running properly.
+
+      Common causes include:
+      - Your maximum Spot price is below the current Spot price for the selected instance type
+      - There is not enough Spot capacity available for the requested instance type in the selected availability zone
+
+      For other common issues, please refer to: https://access.redhat.com/articles/7132524
+
+      If you need assistance, please open a support case.
+    resendWait: 24
+    severity: Warning
+    limitedSupport: false
--- a/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
+++ b/deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
@@ -8,7 +8,7 @@ spec:
     name: nodepool-provisioning-failure-spot
     summary: Spot instance nodepool has not reached its desired size
     notificationMessage: |-
-      One or more Spot instance nodepools in your ROSA cluster has not reached its desired size. This can prevent your workloads from running properly.
+      One or more Spot instance nodepools in your ROSA cluster have not reached their desired size. This can prevent your workloads from running properly.
 
       Common causes include:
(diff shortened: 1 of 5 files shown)


xiaoyu74:
LGTM - defer to other MCC folks to take a look as well.

joshbranham:
This alert is used by ocm-agent-operator to send a service log to customers telling them their nodes aren't provisioning. Wouldn't they want to know that is the case, regardless of spot or not?

SRE wouldn't care if they had no nodes because spot requirements weren't meant, but this alert appears to be customer focused unless I am misunderstanding it.

xiaoyu74:
> This alert is used by ocm-agent-operator to send a service log to customers telling them their nodes aren't provisioning. Wouldn't they want to know that is the case, regardless of spot or not?
> 
> SRE wouldn't care if they had no nodes because spot requirements weren't meant, but this alert appears to be customer focused unless I am misunderstanding it.

@joshbranham Good point - You're right that this change would also suppress the customer-facing `NodePoolFailureNotification`, not just the SRE paging path.

From the Jira, this looks intentional, with Spot capacity/availability being treated as an expected condition and the service log considered potentially noisy. The key question to me is **whether suppressing the customer notification for all Spot nodepools was already discussed and agreed as the intended behavior.**

@feichashao - please correct me if I’m misunderstanding, or confirm if this was already aligned.

feichashao:
Hi @xiaoyu74 @joshbranham   I pushed a new commit for add a dedicated alert and SL template for spot instances. I edited the PR description accordingly.

feichashao:
Push new commits to change according to discussion over slack

deploy/sre-prometheus/ocm-agent/obo-monitoring/100-nodes-not-joining-nodepool.PrometheusRule.yaml
```diff
@@ -36,8 +66,8 @@ spec:
             )
         - alert: NodepoolFailureNotification
           annotations:
-            description: "One or more nodepool for cluster {{ $labels._id }} has not reached its target size for more than 1 hour"
-            summary: "NodePool provisioning Failure"
+            description: "One or more on-demand nodepools for cluster {{ $labels._id }} have not reached its target size for more than 1 hour"
```
joshbranham:
```suggestion
            description: "One or more on-demand nodepools for cluster {{ $labels._id }} have not reached their target size for more than 1 hour"
```

deploy/sre-prometheus/ocm-agent/obo-monitoring/100-nodes-not-joining-nodepool.PrometheusRule.yaml
```diff
@@ -46,3 +76,15 @@ spec:
             managed_notification_template: nodepool-provisioning-failure
             cluster_id: "{{ $labels._id }}"
             namespace: "{{ $labels.namespace }}"
+        - alert: SpotNodepoolFailureNotification
+          annotations:
+            description: "One or more spot instance nodepools for cluster {{ $labels._id }} have not reached its target size for more than 1 hour"
```
joshbranham:
```suggestion
            description: "One or more spot instance nodepools for cluster {{ $labels._id }} have not reached their target size for more than 1 hour"
```

deploy/ocm-agent-operator-managedfleetnotifications/nodepool-provisioning-failure-spot.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: ocmagent.managed.openshift.io/v1alpha1
+kind: ManagedFleetNotification
```
joshbranham, line 2: Have you tested this manifest to ensure it meets the CRD requirements? I have seen these fail due to invalid enums etc so worth double checking pre-merge

feichashao: Thank you for mentioning this!

I just did an e2e test to double confirm this.

1. Create a staging HCP cluster.
2. Create a spot nodepool in MC manually. (the ocm api is not available yet, we can create the nodepool manually for testing).
```
  platform:
    aws:
      instanceType: m5.xlarge
      placement:
        marketType: Spot
        spot:
          maxPrice: "0.05"  //<--- set a low price so it can likely trigger the alert.
```
3. Apply the `ManagedFleetNotification` and `PrometheusRule.monitoring.rhobs` CR in this PR. (It will get reverted by hive after some time.)
4. Observe and can see a SL sent to my cluster.

PR #2829, opened 2026-07-29
[FedRAMP] Make CgaoInactiveHeartbeat less sensitive in the case of gaps in series

jonahbrawley:
### What type of PR is this?
_(bug)_

Adds `keep_firing_for: 30m` to FedRAMP alert `CgaoInactiveHeartbeat`

In FedRAMP, we've had instances where a pod restart creates a minor gap in the metric series, AM then clears the alert and it re-fires, creating noise.

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

- **Bug Fixes**
  - Alerts for inactive heartbeats now remain active for an additional 30 minutes after the triggering condition clears, improving alert continuity and visibility.


cjnovak98:
Seems like a simple enough fix - even if it doesn't ultimately fix the issue, it's good practice to help prevent flapping. Nice find.

PR #2861, opened 2026-08-25
[ROSAENG-64788] removing noise from backup logs

TylerFilko:
### What type of PR is this?
[Ticket](https://redhat.atlassian.net/browse/ROSAENG-64788)
Aim here is to remove the noise within MC backups around resources that can not be backed up or are managed via ACM, thus are not recreated (or shouldn't be) via a restore

More on the issue can be found [here in the ticket](https://redhat.atlassian.net/browse/ROSAENG-64788?focusedCommentId=18137913)

####Summary is as follows:
On management cluster `hs-mc-kbj34s6bg`, the hive-managed `daily-full-backup` schedule completes as `PartiallyFailed` with ~16k item errors. Investigation shows the resulting object in the backup bucket contains **only `metadata/version`** — no resources at all. Every error is the *same* error, emitted by the `hypershift-oadp-plugin` BackupItemAction (BIA): it runs against resources that have **no associated HostedControlPlane**, performs an all-namespace HCP lookup, fails with `no HostedControlPlane found`, and returns an rpc error. Because Velero aborts an item's archival when its BIA errors, essentially every item is excluded from the tarball.

####Fix as per suggestion from @andrebeltrami 
[See suggestion here](https://redhat.atlassian.net/browse/ROSAENG-64788?focusedCommentId=18139157)

The storage layer is healthy and has been ruled out.

_Fixes #_

    ```yaml
    matchExpressions:
      operator: NotIn
      values: ["true"]
    ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated daily, hourly, weekly, and environment-specific backup schedules to exclude Velero’s own management resources.
  * Prevented backup metadata, restore records, repository resources, volume backup and restore objects, and related resources from being redundantly backed up.
  * Added exclusions for Open Cluster Management policy resources and selected OpenShift Tuned resources where applicable.

### DIFF
--- a/deploy/velero-configuration/110-velero.Schedules.yaml
+++ b/deploy/velero-configuration/110-velero.Schedules.yaml
@@ -25,6 +25,21 @@ spec:
     - serviceplans.servicecatalog.k8s.io
     - events.events.k8s.io
     - events
+    - backups.velero.io
+    - backupstoragelocations.velero.io
+    - schedules.velero.io
+    - deletebackuprequests.velero.io
+    - downloadrequests.velero.io
+    - restores.velero.io
+    - serverstatusrequests.velero.io
+    - podvolumebackups.velero.io
+    - podvolumerestores.velero.io
+    - resticrepositories.velero.io
+    - backuprepositories.velero.io
+    - configurationpolicies.policy.open-cluster-management.io
+    - policies.policy.open-cluster-management.io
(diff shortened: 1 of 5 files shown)


Ajpantuso:
@TylerFilko This is the MVO configuration. Will point you to the right direction in slack.

